PR #22, opened 2024-04-06
ci: Add supabase service to docker compose

topher-lo:
Setup all Tracecat services with a single bash script.

Note: AWS Fargate stack now broken. Will fix in next PR.


daryllimyt (approved):
LGTM

PR #38, opened 2024-04-14
feat: Integrate secrets manager with integration functions

daryllimyt:
# Changes
- Allow secrets to be declared in the `@registry.register()` decorator
- The key-value pairs inside each secret (that's added from the frontend) are loaded dynamically into each integration function call.
- See secrets/README.md for more details


topher-lo (approved):
LGTM

PR #58, opened 2024-04-17
feat(ui): Add more checks before initializing posthog

daryllimyt:
## Reference Issues/PRs
Inspired by the way Infisical does their posthog init.

## What does this implement/fix? Explain your changes.
- Performs additional checks before initializing posthog. Was facing an error when using `pnpm start` with Docker and using NODE_ENV=production -- I suspect that doing so has some less-documented effect on posthog/supabase. 
- Add explicit ENABLE_TELEMETRY env var.

## Any other comments?
- Currently just completely circumventing this issue by using `pnpm dev`


daryllimyt:
Closed as not planned

PR #66, opened 2024-04-19
feat(ui+engine): Improve case management [2/N]

### LINKED ISSUE #70: Update autocomplete for tags only
# Motivation
Don't need to update context and acitons, just add AI tags (v0 - autocompleted from our curated list of tracecat-owned tags)

daryllimyt:
# Changes
- Remove support for autocompleting `action` and `context`. `context` should be populated during the workflow, and action should be intentional (though these are subject to change)
- Update autocomplete to complete for tags only. These include tracecat default mitre tags (see db init) and can be extended include user defined tags (TODO)
- Refactor cases to use `list[Model]` types - allows for duplicate tags with different values (e.g. for mitre labelling)
- Remove null states for case types - easier to handle in zod and leads to cleaner pydantic serde
- Remove metrics banner
- Minor case table styles update
- Some refactoring/simplification of code

Closes #70


topher-lo (approved):
LGTM

PR #83, opened 2024-04-21
feat(integration): Add analyze URL via URLScan Action

### LINKED ISSUE #82: URLScan
**User story:** I want to submit a URL to URLScan for analysis.

API reference: https://urlscan.io/docs/api/

**Implementation details**
- Two stage API: POST call, then wait, then GET report
- Tracecat is not bounded by the one endpoint per Action constraint
- Action = function. Prioritize usefulness over purity
- After the initial POST call, wait with exponential backoff, 
- There are three `visibility` options: `public`, `private`, `unlisted` (only visible to verified security researchers)

**Key requirements**
- [ ] Can scan URL and return report
- [ ] `private` visibility by default : https://portswigger.net/daily-swig/urlscan-io-api-unwittingly-leaks-sensitive-urls-data

topher-lo:
Closes #82


tests/integrations/test_urlscan.py
```diff
@@ -0,0 +1,35 @@
...
+    serialized_secret = mock_secret.model_dump()
+    return serialized_secret
+
+
+def test_analyze_url_live(urlscan_secret, respx_mock):
+    test_url = "http://google.com"
+
+    # Mock secrets manager
+    route = respx_mock.get(f'{os.environ["TRACECAT__API_URL"]}/secrets/urlscan').mock(
+        return_value=Response(status_code=200, json=urlscan_secret)
```
topher-lo: @daryllimyt I'm just blocked on one final issue for the tests:
```
TypeError: Object of type bytes is not JSON serializable
```

From:
```
.mock(return_value=Response(status_code=200, json=urlscan_secret))
```
Where `urlscan_secret` is a JSON-serialized `Secret`. The issue I think is `encrypted_keys` is bytes.

We don't get this error in the `GET` secrets FastAPI endpoint because FastAPI / pydantic coerces bytes into strings before passing into the Response content (I think).

Also why `.content`: https://github.com/TracecatHQ/tracecat/blob/main/tracecat/secrets/helpers.py#L18C43-L18C63
Not `.json`?

daryllimyt: > Why `.content` and not `.json`

Want to leverage Pydantic to perform the model construction directly from bytes, which is likely more efficient (likely a micro-optimization at this stage, but why not) than using httpx's `.json`.

daryllimyt: > We don't get this error in the GET secrets FastAPI endpoint because FastAPI / pydantic coerces bytes into strings before passing into the Response content (I think).

Not exactly, we use the `Secret.keys` property to apply encryption/decryption. The definition is [here](https://github.com/TracecatHQ/tracecat/blob/4b2783028d47aca1b8ec159b1bc38bafebf19ae0/tracecat/db.py#L117C1-L130C54) and usage is [here](https://github.com/TracecatHQ/tracecat/blob/4b2783028d47aca1b8ec159b1bc38bafebf19ae0/tracecat/api/app.py#L1441)

tests/conftest.py
```diff
@@ -47,3 +51,19 @@ def setup_shared_env():
...
+def create_mock_secret():
+    def _get_secret(secret_name: str, secrets: dict[str, str]) -> list[Secret]:
+        keys = [SecretKeyValue(key=k, value=v) for k, v in secrets.items()]
+        new_secret = Secret(
+            owner_id=uuid4().hex,  # Assuming owner_id should be unique per secret
+            id=uuid4().hex,  # Generate a unique ID for each secret
+            name=secret_name,
+            type="custom",  # Assuming a fixed type; adjust as necessary
+        )
+        new_secret.keys = keys
```
daryllimyt, line 66: So this is the encryption step (because you're calling the [setter](https://github.com/TracecatHQ/tracecat/blob/4b2783028d47aca1b8ec159b1bc38bafebf19ae0/tracecat/db.py#L125-L130)), which why you're seeing a `bytes` object as your `encrypted_keys`

tests/integrations/test_urlscan.py
```diff
@@ -0,0 +1,35 @@
...
+
+from tracecat.integrations.urlscan import analyze_url
+
+
+@pytest.fixture
+def urlscan_secret(create_mock_secret) -> dict[str, str | bytes]:
+    mock_secret = create_mock_secret(
+        "urlscan", {"URLSCAN_API_KEY": os.environ["URLSCAN_API_KEY"]}
+    )
+    serialized_secret = mock_secret.model_dump()
```
daryllimyt: so based on the above comment, you need to perform the complementing deserialization (using the getter) to get the actual decoded values

daryllimyt (approved):
LGTM

PR #146, opened 2024-05-29
feat: Temporal rewrite

daryllimyt:
# Changes
- All actions and integrations are now UDFs (user defined function)
- UDF args and response schemas get parsed into pydantic models for validation
- Workflow engine rewritten in Temporal
- We can now operate the SOAR headless, using a yaml DSL interface
- Added CLI for convenience

# WIP
- Frontend nocode -> code still WIP
- UDF pydantic models will be used to generate UI form elements

# Testing
- Rewrote tests for major components
- Unit tests all passing


topher-lo (commented):
LGTM as discussed on Slack

PR #174, opened 2024-06-16
feat+fix(engine): Implement resource identifiers module + fix Temporal clashing workflow ID

daryllimyt:
# Features
- Implement (revamped utils) identifiers module, which contains a suite of methods, types, and validators for Tracecat resource identifiers. 
    - We leverage pydantic's `StringConstraint` with regex + `typing.Annotated` to achieve runtime pattern checking
    - Also contains method to create action keys/refs and workflow run ids etc.
- Add `ResourcePrefix` to centralize all resource class identifiers 

# Changes
- Align our world view with Temporal - we distinguish between workflow, workflow execution (a running workflow instance), and workflow runs (a try/attempt of a running instance).
    - See updated `contexts.RunContext`

# Testing
- Updated tests to use a fixed test id, workflow tests passing


topher-lo (changes requested):
Would prefer if gen_id was generate_id

topher-lo (approved):

PR #176, opened 2024-06-16
feat(engine): Dynamically register UDFs in DSLActivities

daryllimyt:
# Features
- Instead of displaying `run_udf` for every function, dynamically wrap each UDF in a Temporal activity and register it with workers.
    - In the UDF key we replace `.` with `__`. e.g. `ns.subns.method_name` -> `ns__subns__method_name`

# Remarks
- Improve workflow observability
- Before we start a worker, generate temporal activity wrappers for all registered UDFs so that they show up as activity types in the Temporal UI

# Screens
image

# Testing
- All workflow tests passing


topher-lo (approved):
LGTM

PR #180, opened 2024-06-18
ci: Add draft GH workflow to test all playbooks

topher-lo:
## NOTE:
This is an incomplete PR. Need to deal with accessing docker / localhost from within github actions issues in a follow-up PR.

## What changed
- [x] Add workflow to run playbooks via pytest (very similar setup to #177)
- [x] Made webhooks secret
- [x] Gave each playbook trigger ref a meaningful name  

## Comments
- Assumes no data is sent to webhook
- Need to extend this test suite to support playbooks with incoming data into the trigger
- Need to extend this test suite with scheduled workflows

### DIFF
--- a/tests/integration/test_playbooks.py
+++ b/tests/integration/test_playbooks.py
@@ -1,0 +1,71 @@
+import json
+import os
+import subprocess
+
+import pytest
+
+
+# Fixture to create Tracecat secrets
+@pytest.fixture(scope="session", autouse=True)
+def create_secrets():
+    secrets = {
+        "abusech": {"ABUSECH_API_KEY": os.getenv("ABUSECH_API_KEY")},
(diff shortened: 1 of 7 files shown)


tests/integration/test_playbooks.py
```diff
@@ -0,0 +1,76 @@
...
+    output = subprocess.run(
+        ["tracecat", "workflow", "commit", "--file", path_to_playbook, workflow_id],
+        capture_output=True,
+        text=True,
+    )
+    assert "Successfully committed to workflow" in output.stdout
+    # 3. Activate the workflow
+    subprocess.run(["tracecat", "workflow", "up", "--webhook", workflow_id])
+    # 4. Run the workflow
+    subprocess.run(["tracecat", "workflow", "run", workflow_id])
```
topher-lo: @daryllimyt how we can run this without ngrok?

PR #182, opened 2024-06-20
build(engine): Add basic Caddy reverse proxy for API requests

daryllimyt:
# Motivations
- Remove ngrok from default install path

# Changes
- We'll still suggest ngrok as an optional tool (if users can and wish to hit APIs external data sources)
- Include Caddy as a simple rev proxy
- Hit `localhost` to access the API

Screenshot 2024-06-20 at 14 00 58


topher-lo (commented):
Stale. Will redo in newer PR.

PR #191, opened 2024-06-22
feat(integrations): Migrate crowdstrike integrations to falconpy

topher-lo:
## Rationale
[falconpy](https://github.com/CrowdStrike/falconpy) is the official CS client from CS. Fully unit tested, robust support, and actively maintained.


daryllimyt (approved):
LGTM

PR #194, opened 2024-06-23
feat(engine): Implement runtime action tests

daryllimyt:
# API
```yaml
# tracecat workflow run <wf_id> --test
tests:
  # If an action has a test, do not run the action and mock/patch the output
  - ref: pull_crowdstrike_alerts
    enable: true # default: true
    validate_args: true # default: true
    # Path (use fsspec to support both local files and http)
    # to sample JSON or YAML content (to be JSONfied)
    success: # This is the mocked/patched output for the action.
    failure: # TODO

config:
    # Statically configure this here, or omit this and pass it in at execution time
    enable_runtime_tests: true
```
# Features
- Adds new top-level workflow directive `tests`
- Define a list of `tests` that mock the output value of actions at runtime
- By default, all tests defined are enabled. You can control whether they're enabled or disabled at runtime.
- You can override any action with any data structure
- You can load mock data using `fsspec` compliant uris/paths
- Add custom headers module
- Pass `--test` (CLI) or X-Tracecat-Enable-Runtime-Tests: true` (http) to enable runtime action tests
- At execution time, for incoming webhooks we check for `X-Tracecat-Enable-Runtime-Tests` header and set `DSLConfig` in the workflow input. Otherwise, this config can be statically configured in the `config` directive.

# Testing
- Added new tests
- All unit tests passing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,7 @@ dependencies = [
     "cryptography==42.0.7",
     "diskcache==5.6.3",
     "fastapi==0.111.0",
+    "fsspec==2024.6.0",
     "httpx==0.27.0",
     "jsonpath_ng==1.6.1",
     "lancedb==0.6.3",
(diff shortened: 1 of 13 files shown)


topher-lo (commented):
One failing test

topher-lo (approved):

PR #209, opened 2024-06-24
ci(perf): Add workflow concurrency stress tests

topher-lo:
## What changed
- Added GH Action workflow to run pytest benchmarking on `test_stress_workflows`
- Parameterized `test_stress_workflows` by num_workflows
- Parameterize GH Action via matrix with `num_workers` to scale number of workers for `worker` service
- Added health check endpoint in Tracecat API to check if Temporal cluster can be connected to
- Move dsl fixture into conftest


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,57 @@
+name: Benchmark
+
+on:
+  push:
+    branches: ["main"]
+    paths: ["tracecat/**"]
+  pull_request:
+    branches: ["main"]
+    paths: [".github/workflows/stress-test.yml"]
```
topher-lo:
```suggestion
    paths: [".github/workflows/benchmark.yml"]
```

PR #210, opened 2024-06-24
fix: Update playbooks test

daryllimyt:
# Changes
- Copy `workflow_id` matching from webhook test


topher-lo (commented):
Closed in favor of #213

PR #213, opened 2024-06-26
test: Fix flaky `workflow_id` parse in playbooks test

topher-lo:
## What changed
- Ignore secrets creation fixture errors (to deal with local development case when secrets have already been created and clash)
- Replaced create workflows CLI commands in test with direct calls to the REST API via the private create and activate workflow functions
- That function returns a json with workflow ID under the id field
- Replaced inputs with triggers in AWS guardduty to slack playbook

### DIFF
--- a/tracecat/cli/workflow.py
+++ b/tracecat/cli/workflow.py
@@ -123,20 +123,6 @@ async def _get_cases(workflow_id: str):
     return res.json()
 
 
-async def _create_activate_workflow(
-    title: str | None = None,
-    description: str | None = None,
-) -> str:
-    """Create and activate a workflow with a webhook.
-
-    Mainly used for testing.
-    """
-    workflow = await _create_workflow(title, description)
-    workflow_id = workflow["id"]
-    await _activate_workflow(workflow_id, with_webhook=True)
-    return workflow_id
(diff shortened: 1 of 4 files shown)


tracecat/cli/workflow.py
```diff
@@ -123,6 +123,20 @@ async def _get_cases(workflow_id: str):
...
+    """Create and activate a workflow with a webhook.
+
+    Mainly used for testing.
+    """
+    workflow = await _create_workflow(title, description)
+    workflow_id = workflow["id"]
+    await _activate_workflow(workflow_id, with_webhook=True)
+    return workflow_id
+
+
```
daryllimyt: nit: if this is only used for testing, let's put it in the test file

daryllimyt (changes requested):
Please add a short description of what changed, otherwise LGTM

tracecat/cli/workflow.py
topher-lo: Might be used else where. It's useful enough not to be just a testing thing imo. Just added that comment to remind me what's it for (for now)

daryllimyt (approved):

PR #225, opened 2024-07-03
feat(engine): Dockerise Tracecat CLI

topher-lo:
## What changed
- Added Dockerfile and image (pushed to ghcr) so the CLI can be used without a python installation
- Updated docs to show both options to run Tracecat CLI
- Updated `__versions__` to 0.1.4-dev

## QA
- Usage example: `docker run --rm ghcr.io/tracecathq/tracecat-cli --version --network host`

Note: need `--network host` to access Tracecat API via `http://localhost:8000`

Screenshot 2024-07-03 at 5 03 32 PM


daryllimyt (approved):
LGTM land it

PR #245, opened 2024-07-16
feat(integration): Slack list past messages

topher-lo:
## What changed
- Added new function list_slack_conversations to fetch past messages from a Slack channel, including parameters for channel ID, limit, latest, and oldest timestamps.
- Updated post_slack_message to include a thread_ts parameter for creating threads within Slack channels.
- Changed user_agent from "bolt-python" to "slack-python" for consistency across all Slack API calls.


daryllimyt (approved):
LGTM

PR #247, opened 2024-07-17
build(engine): Replace `polars` with `polars-lts-cpu`

### LINKED ISSUE #246: Inclusion of 'polars' Python library instead of 'polars-lts-cpu' limits who can use this
**Describe the bug**
The documentation does not call out in an obvious way that the use of the 'polars' library prevents anyone without AVX2 instructions from using this.  That means only AMD Zen 4 and newer or Intel Ice Lake and newer are able to run this using the pre-built Docker images from the registry.

**To Reproduce**
Run this on a machine with a CPU without AVX2 instructions.  The tracecat-worker and tracecat-api container will restart endlessly with entries in the log that the CPU does not support required instructions for the polars library.

**Expected behavior**
Give us an alternate image compiled with the polars-lts-cpu library instead of polars.

**Screenshots**
N/A

- OS: Debian 12
- CPU: Intel i7-4790K
- Browser version: N/A
- Docker version: 20.10.24
- Docker compose version: 2.18.1
- Are you using Docker desktop? No

**Additional context**
I did use the Dockerfile from the repo to make this change myself, and while the api container runs normally now, the worker container continues to restart repeatedly.

topher-lo:
Fixes #246. This sacrifices performance for SIMD heavy queries. But we have yet to fully leverage polars ETL in Tracecat (yet). Will push this temporary fix first.

## Future direction
- Build two images: one with `polars` and another with `polars-lts-cpu`
- Create docker compose extension for users without AVX

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ dependencies = [
     "loguru==0.7.2",
     "openai==1.30.3",
     "orjson==3.10.3",
-    "polars==0.20.21",
+    "polars==1.2.0",
     "psycopg[binary]==3.1.19",
     "psycopg2-binary==2.9.9",
     "pyarrow==16.1.0",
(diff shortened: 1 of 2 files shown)


daryllimyt (approved):
LGTM

PR #248, opened 2024-07-17
feat(engine)! Rename `forward` action to `reshape`

topher-lo:
## What changed?
- Rename `core.transform.forward` action to `core.transform.reshape`

## Why
- Reshape aligns with industry standard terminology for this particular use-case (also we've been naturally naming our action `refs` as `reshape_xyz` not `forward_xyz`.


daryllimyt (approved):
LGTM

Issue #255, opened 2024-07-21
Little issue in namespace configuration of ./playbooks/threat_intel/virustotal-to-email.yml


r0psteev:
**Description**

Hello, there is a little issue in this workflow definition about the namespace from which the
function `analyze_url` should be imported, and it causes a commit error.

```sh
(tracecat) developer@pc:~/Documents/tracecat$ tracecat workflow create --commit playbooks/threat_intel/virustotal-to-email.yml --title 'Virustotal email #1' --webhook --activate
Created workflow
{
    'id': 'wf-c399a3920a724b06ad550292406fb69f',
    'title': 'Virustotal email #1',
    'description': 'New workflow created Jul 21, 2024, 02:29:26',
    'status': 'offline',
    'icon_url': None,
    'created_at': '2024-07-21T02:29:26.128782Z',
    'updated_at': '2024-07-21T02:29:26.128782Z',
    'version': None
}
Failed to commit to workflow 'wf-c399a3920a724b06ad550292406fb69f'!
{
  "workflow_id": "wf-c399a3920a724b06ad550292406fb69f",
  "status": "failure",
  "message": "2 validation error(s)",
  "errors": [
    {
      "ok": false,
      "message": "Could not find UDF 'integrations.enrich.virustotal.analyze_url' in registry. Is this UDF 
registered?",
      "detail": {}
    },
  ....
  ],
  "metadata": {
    "filename": "virustotal-to-email.yml"
  }
}
(tracecat) developer@pc:~/Documents/tracecat$ 
```

- In file `./tracecat/actions/integrations/enrichment/virustotal.py` namespace is `integrations.virustotal` for the function `analyze_url`

```python
33 @registry.register(
 32     default_title="Analyze URL",
 31     description="Analyze a URL using VirusTotal.",
 30     display_group="VirusTotal",
 29     namespace="integrations.virustotal",
 28     secrets=[virustotal_secret],
 27 )
 26 async def analyze_url(
 25     url: Annotated[str, Field(..., description="The URL to analyze")],
 24 ) -> dict[str, Any]:
 23     url_id = base64.urlsafe_b64encode(url.encode()).decode().strip("=")
 22     async with create_virustotal_client() as client:
 21         response = await client.get(f"/v3/urls/{url_id}")
 20         response.raise_for_status()
 19         return response.json()
 18 
```

**Quick fix**

[...]

PR #256, opened 2024-07-21
fix(playbooks): Update virustotal action  namespace

r0psteev:
## Description

Issue #255 
As user base grows, similar sneaky bugs will eventually be factored out.


daryllimyt (approved):
LGTM!

topher-lo:
@r0psteev one last comment: we plan to defend against problems like this by adding validation checks across all public playbooks in CI.

We can't just rely on good folks like you to defend against regression bugs 🫡

PR #261, opened 2024-07-23
feat(ui): Add action control flow, workflow static inputs and workflow return value expression

daryllimyt:
# Features
- Add control flow editor
    - Run if
    - Loop 
- Add workflow return value expression editor. We need this for child workflows to return values to the parents
- Add workflow static inputs editor
- Add tooltips and usage examples for each
- Create custom editor component with upgraded editor styles
- Update api to handle the able use cases

# Changes
- Changed action node `copy jsonpath` to `copy action reference` which will copy `ACTIONS.<node slug>`

# Testing
- Manual QA with child workflows. Also tested running child workflows with schedules.
- Found some edge cases when parsing the workflow return value fails, which triggers infinite retry. Urgently needs to be fixed

# Screens
Screenshot 2024-07-23 at 15 25 53
Screenshot 2024-07-23 at 15 25 41
Screenshot 2024-07-23 at 15 25 31


frontend/src/components/workspace/canvas/udf-node.tsx
```diff
@@ -113,7 +114,7 @@ export default React.memo(function UDFNode({
               <DropdownMenuContent align="end">
                 <DropdownMenuItem onClick={handleCopyToClipboard}>
                   <Copy className="mr-2 size-4" />
-                  <span className="text-xs">Copy JSONPath</span>
+                  <span className="text-xs">Copy Action Reference</span>
```
topher-lo:
```suggestion
                  <span className="text-xs">Copy action reference</span>
```

topher-lo (approved):
LGTM

PR #262, opened 2024-07-23
build: Add dev Dockerfile and docker-compose yml

daryllimyt:
# Changes
- Create standalone `Dockerfile.dev` and `docker-compose.dev.yml`
- Invoke using `docker compose -f docker-compose.dev.yml up`
- Avoids image/build clashes with prod docker setup, use `docker restart tracecat_worker` to reflect changes in the worker (doesn't have hot reload)

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,55 @@
+FROM python:3.12-slim-bookworm
+
+# Define the environment variables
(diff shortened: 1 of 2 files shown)


topher-lo (approved):
LGTM but will rewrite Dockerfile to make it build faster for dev. No need for performance / sec optimisations

PR #263, opened 2024-07-23
Add new SentinelOne actions: lookup by username, hostname, isolate, unisolate

mattdurant:
## Description

This adds additional actions to the existing Sentinel One integration:

- Agent lookup by hostname
- Agent lookup by username
- Isolate endpoint
- Unisolate endpoint

These are common response actions for an EDR product; as an example, based on an alert, you could lookup a hostname or username to find which machines they are active on, then isolate those which would block all connections except the communication channel that the EDR tool uses to communicate with it's control plane.

No open issues for this

## Screenshots/Recordings

N/A

## Steps to QA

I tested this by creating a workflow that looked up an agent, performed an isolation, and then performed an unisolation and confirmed in Sentinel One that all actions took place correctly.


mattdurant:
This can't be merged yet, I didn't add the new actions to the __init__.py file in the edr directory, will do in a minute

mattdurant:
Ready for review

topher-lo (approved):
Clean API 👍

PR #269, opened 2024-07-25
Okta enhancement/fix

mattdurant:
## Description

Okta API calls do not always accept supplying the user's ID using the profile.login value.  Added a new action to lookup user(s) by login/email and modified other actions to take Okta user id as input.

## Steps to QA

Tested against live Okta environment

### DIFF
--- a/tracecat/actions/integrations/iam/okta.py
+++ b/tracecat/actions/integrations/iam/okta.py
@@ -27,6 +27,7 @@ okta_secret = RegistrySecret(
     - `OKTA_API_TOKEN`
 """
 
+
 @registry.register(
     default_title="Find Okta User",
     description="Find an Okta user by login username or email",
@@ -49,12 +50,13 @@ async def find_okta_user(
     }
 
     async with httpx.AsyncClient() as client:
-        response = await client.post(
+        response = await client.get(
             f"{base_url}/api/v1/users?search=profile.login%20eq%20%22{username_or_email}%22%20or%20profile.email%20eq%20%22{username_or_email}%22",
             headers=headers,
         )
         response.raise_for_status()
         return response.json()
+
 
 @registry.register(
     default_title="Suspend Okta user",
--- a/tracecat/actions/integrations/iam/okta.py
+++ b/tracecat/actions/integrations/iam/okta.py
@@ -27,6 +27,34 @@ okta_secret = RegistrySecret(
     - `OKTA_API_TOKEN`
 """
 
+@registry.register(
+    default_title="Find Okta User",
+    description="Find an Okta user by login username or email",
+    display_group="Okta",
+    namespace="integrations.okta",
+    secrets=[okta_secret],
+)
+async def find_okta_user(
+    username_or_email: Annotated[
+        str,
+        Field(..., description="Login username or e-mail to find"),
+    ],
+) -> list:
+    api_token = os.getenv("OKTA_API_TOKEN")
+    base_url = os.getenv("OKTA_BASE_URL")
+    headers = {
+        "Authorization": f"SSWS {api_token}",
+        "Accept": "application/json",
+        "Content-Type": "application/json",
+    }
+
+    async with httpx.AsyncClient() as client:
+        response = await client.post(
+            f"{base_url}/api/v1/users?search=profile.login%20eq%20%22{username_or_email}%22%20or%20profile.email%20eq%20%22{username_or_email}%22",
+            headers=headers,
+        )
+        response.raise_for_status()
+        return response.json()
 
 @registry.register(
     default_title="Suspend Okta user",
@@ -36,9 +64,9 @@ okta_secret = RegistrySecret(
     secrets=[okta_secret],
 )
 async def suspend_okta_user(
-    username: Annotated[
+    okta_user_id: Annotated[
         str,
-        Field(..., description="Username to suspend"),
+        Field(..., description="Okta user id to suspend"),
     ],
 ) -> bool:
     api_token = os.getenv("OKTA_API_TOKEN")
@@ -51,7 +79,7 @@ async def suspend_okta_user(
 
     async with httpx.AsyncClient() as client:
         response = await client.post(
-            f"{base_url}/api/v1/users/{username}/lifecycle/suspend",
+            f"{base_url}/api/v1/users/{okta_user_id}/lifecycle/suspend",
             headers=headers,
         )
         response.raise_for_status()
@@ -66,9 +94,9 @@ async def suspend_okta_user(
     secrets=[okta_secret],
 )
 async def unsuspend_okta_user(
-    username: Annotated[
+    okta_user_id: Annotated[
         str,
-        Field(..., description="Username to unsuspend"),
+        Field(..., description="Okta user id to unsuspend"),
     ],
 ) -> bool:
     api_token = os.getenv("OKTA_API_TOKEN")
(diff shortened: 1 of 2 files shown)


tracecat/actions/integrations/iam/okta.py
```diff
@@ -28,6 +28,36 @@
...
+) -> list:
+    api_token = os.getenv("OKTA_API_TOKEN")
+    base_url = os.getenv("OKTA_BASE_URL")
+    headers = {
+        "Authorization": f"SSWS {api_token}",
+        "Accept": "application/json",
+        "Content-Type": "application/json",
+    }
+
+    async with httpx.AsyncClient() as client:
```
daryllimyt: Could we extract this into a function outside the UDF that returns the preconfigured client? I see you have a similar setup between the other okta integrations too.

There are some examples of this under `integrations/enrichment`

tracecat/actions/integrations/iam/okta.py
```diff
@@ -28,6 +28,36 @@
...
+        response = await client.get(
+            f"{base_url}/api/v1/users?search=profile.login%20eq%20%22{username_or_email}%22%20or%20profile.email%20eq%20%22{username_or_email}%22",
+            headers=headers,
+        )
```
daryllimyt:
```suggestion
        response = await client.get(
            f"{base_url}/api/v1/users",
            params={ ... }.
            headers=headers,
        )
```

daryllimyt: Could we use the `params` kwarg here instead? (example above)

daryllimyt (changes requested):
Just some minor things, but LGTM otherwise

tracecat/actions/integrations/iam/okta.py
mattdurant: I can, the reason I left it in the uri is that there is only one parameter; it's encoded that way due to how Okta wants the search query passed.  Basically, this one would be pulled out and the ugly search query would be the only item in the param list:

`params={search=f" profile.login%20eq%20%22{username_or_email}%22%20or%20profile.email%20eq%20%22{username_or_email}%22"}`

Not a problem to change though.

mattdurant:
@daryllimyt Requested changes committed

tracecat/actions/integrations/iam/okta.py
daryllimyt: @mattdurant  Gotcha. (I know this is an outdated thread but just gonna put this here) I think it's easier to read if we format it without the encoding, and if Okta requires it url-encoded, `httpx` does perform the urlsafe percent encoding in the method (unless I'm misunderstanding what you mean specifically by encoding?)

Something like this should work:
```python
params = {
    "search": (
        f'profile.login eq "{username_or_email}" or profile.email eq "{username_or_email}"'
    )
}
response = await client.get(f"{base_url}/api/v1/users", headers=headers, params=params)
return response
```

Am I missing something here?

tracecat/actions/integrations/iam/okta.py
```diff
@@ -28,6 +28,49 @@
 """
 
 
+def create_okta_client() -> httpx.AsyncClient:
+    OKTA_API_TOKEN = os.getenv("OKTA_API_TOKEN")
+    if OKTA_API_TOKEN is None:
+        raise ValueError("OKTA_API_TOKEN is not set")
+    client = httpx.AsyncClient(
+        base_url=os.getenv("OKTA_BASE_URL"),
```
daryllimyt: nit: you can make the base_url include `/api/v1` so you don't have to include that in downstream usage. So later on you can just pass `/users` as the url in the client

tracecat/actions/integrations/iam/okta.py
```diff
@@ -28,6 +28,49 @@
...
+    username_or_email: Annotated[
+        str,
+        Field(..., description="Login username or e-mail to find"),
+    ],
+) -> list:
+    async with create_okta_client() as client:
+        params = [
+            "search",
+            f"profile.login%20eq%20%22{username_or_email}%22%20or%20profile.email%20eq%20%22{username_or_email}%22",
+        ]
```
daryllimyt: This should be a dict like
```python
params = {
    "search": ...
}
```

tracecat/actions/integrations/iam/okta.py
```diff
@@ -28,6 +28,49 @@
...
+        response = await client.get(
+            f"{client.base_url}/api/v1/users?search=profile.login%20eq%20%22{username_or_email}%22%20or%20profile.email%20eq%20%22{username_or_email}%22",
+            headers=client.headers,
+            params=params,
+        )
```
daryllimyt: when you create a preconfigured client, `headers` is already populated so you don't have to pass this in again. Similarly, when passing in `base_url` you only need to pass the specific endpoint like such:

```python
import asyncio

import httpx


def custom_client() -> httpx.AsyncClient:
    return httpx.AsyncClient(
        base_url="http://localhost:8005",
        headers={"X-Custom": "value"},
    )
async def main():
    async with custom_client() as client:
        response = await client.get("/ping") # This will hit http://localhost:8005/ping with the custom header X-Custom: value
        print(response)


asyncio.run(main())

```

tracecat/actions/integrations/iam/okta.py
```diff
@@ -36,23 +79,15 @@
...
-        "Content-Type": "application/json",
-    }
-
-    async with httpx.AsyncClient() as client:
+    async with create_okta_client() as client:
         response = await client.post(
-            f"{base_url}/api/v1/users/{username}/lifecycle/suspend",
-            headers=headers,
+            f"{client.base_url}/api/v1/users/{okta_user_id}/lifecycle/suspend",
+            headers=client.headers,
```
daryllimyt: same as above

tracecat/actions/integrations/iam/okta.py
```diff
@@ -66,23 +101,15 @@ async def suspend_okta_user(
...
-        "Content-Type": "application/json",
-    }
-
-    async with httpx.AsyncClient() as client:
+    async with create_okta_client() as client:
         response = await client.post(
-            f"{base_url}/api/v1/users/{username}/lifecycle/unsuspend",
-            headers=headers,
+            f"{client.base_url}/api/v1/users/{okta_user_id}/lifecycle/unsuspend",
+            headers=client.headers,
```
daryllimyt: same as above

tracecat/actions/integrations/iam/okta.py
```diff
@@ -96,23 +123,15 @@ async def unsuspend_okta_user(
...
-        "Content-Type": "application/json",
-    }
-
-    async with httpx.AsyncClient() as client:
+    async with create_okta_client() as client:
         response = await client.delete(
-            f"{base_url}/api/v1/users/{username}/sessions",
-            headers=headers,
+            f"{client.base_url}/api/v1/users/{okta_user_id}/sessions",
+            headers=client.headers,
```
daryllimyt: same as above

tracecat/actions/integrations/iam/okta.py
mattdurant: Ah, I was not aware of httpx performing the encoding for you, yes that will work.  I thought in your original comment you wanted it broken out due to being multiple parameters in the string, my mistake.

daryllimyt (changes requested):
Just some small things!

tracecat/actions/integrations/iam/okta.py
daryllimyt: No worries, all good!

tracecat/actions/integrations/iam/okta.py
mattdurant: Noticed that when I was looking at this to remove the encoding, fixing.

mattdurant:
@daryllimyt Ok, take a look

daryllimyt (approved):
LGTM! just waiting for CI

PR #279, opened 2024-07-28
feat(engine): Turn workflow into API

mattdurant:
Webhook endpoint for synchronous execution and return final workflow output.

## Description

This is so that you can call a webhook in Tracecat to perform some actions and receive as a response the final_context object from the workflow.

My current use case is an outgoing webhook in Teams instead of a full bot implementation, so I can send a message in a channel to @ChatOps with a command, parse the command in the workflow and then do some work and the final_context is returned.  In this case, a chat message indicating success or failure.

N/A

## Screenshots/Recordings

N/A

## Steps to QA

Test by calling a webhook, but add /wait to the URL, the response will return once the workflow is finished and contain the final_context of the workflow.


tracecat/api/app.py
```diff
@@ -373,6 +373,43 @@ async def incoming_webhook(
     return response
 
 
+@app.post("/webhooks/{path}/{secret}/wait", tags=["public"])
+async def incoming_webhook_Wait(
```
daryllimyt: style: snake case

tracecat/api/app.py
```diff
@@ -373,6 +373,43 @@ async def incoming_webhook(
...
+    path: str,
+    payload: dict[str, Any] | None = None,
+    x_tracecat_enable_runtime_tests: Annotated[str | None, Header()] = None,
+) -> dict[str, Any]:
+    """
+    Webhook endpoint to trigger a workflow.
+
+    This is an external facing endpoint is used to trigger a workflow by sending a webhook request.
+    The workflow is identified by the `path` parameter, which is equivalent to the workflow id.
+    """
```
daryllimyt: could we update the docstring here to reflect that this endpoint behaves differently from the regular one?

tracecat/api/app.py
topher-lo:
```suggestion
async def incoming_webhook_wait(
```

daryllimyt (changes requested):

tracecat/api/app.py
mattdurant: Out of curiosity, why would wait be capitalized here? I've not seen snake case have any words capitalized like this before.

mattdurant: Oh man, ignore me, I completely misread Chris' review and thought he suggested capitalizing Wait.  I didn't realize I committed it that way...

daryllimyt: no worries :)

daryllimyt (approved):
LGTM

topher-lo:
Will delay this PR for next week's release while we rework the API routers and add tests this week.

topher-lo (commented):
Wait for API work to complete this week

mattdurant:
@daryllimyt @topher-lo based on the conflict here, I think you guys did a big refactoring during the same time, I think I should close this PR and start clean make the 2-3 changes to enable this again and re-submit, agreed?

daryllimyt:
> @daryllimyt @topher-lo based on the conflict here, I think you guys did a big refactoring during the same time, I think I should close this PR and start clean make the 2-3 changes to enable this again and re-submit, agreed?

Yes - I'm also aiming to close #285  so I think it makes sense to adjust it accordingly

PR #280, opened 2024-07-28
feat(engine): Add organization and workspaces

### LINKED ISSUE #268: [FEATURE IDEA] Workspaces to separate credentials / variables / workflows
If you have multiple of a specific system that you manage, you can't distinguish them and provide different secrets.  For example, if I had 2 instances of Okta to manage, I only have 1 secret defined that is tied to the class, not to a specific instance of the class.

I'm thinking of other systems where you configure a connection, give it creds and config, and then when you add those actions to a workflow you specific which connection it's using.

There are likely MDRs or MSPs that would be interested in using this that wouldn't be able to manage multiple clients from one instance of Tracecat and would need to run their own multi-tenant setup to separate them.

**Describe the solution you'd like**
If we could instantiate the integration classes and save configuration/secrets against those, it would solve some upcoming issues as the project grows:
- More integrations means more configuration items, and right now those are managed through environment variables, that file is going to get longer and will likely contain many placeholders for solutions that every user does not need
- Juggling multiple instances of the same integration, i.e. if I had 2 Okta tenants or 2 M365 tenants to manage, I need to be able to differentiate between them both for secrets as well as configuration.  Even customers with a single instance of each product now could encounter this during M&A activities where you typically can't merge tenants on Day 1

**Describe alternatives you've considered**
There isn't really a good alternate solution to the above problem.

**Additional context**
N/A

topher-lo:
Closes #268

### DIFF
--- a/tracecat/db/schemas.py
+++ b/tracecat/db/schemas.py
@@ -27,7 +27,7 @@ class Resource(SQLModel):
     """Base class for all resources in the system."""
 
     surrogate_id: int | None = Field(default=None, primary_key=True, exclude=True)
-    owner_id: str
+    owner_id: str  # NOTE: Equivalent to team_id
     created_at: datetime = Field(
         sa_type=TIMESTAMP(timezone=True),  # UTC Timestamp
         sa_column_kwargs={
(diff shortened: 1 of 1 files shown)


topher-lo:
Replaced by #302

PR #283, opened 2024-07-29
Add function to extract text elements from HTML

mattdurant:
## Description

Use case for this is parsing input that may contain HTML tags, for example, incoming Teams messages are posted with HTML markup in the "text" field.

N/A

## Screenshots/Recordings

N/A

## Steps to QA

Unit tests and live testing


topher-lo (approved):
LGTM. I'm okay with the simple test since `html` is a base python lib.

PR #286, opened 2024-07-29
fix(ui): Use next-runtime-env to pass environment variables to client side

daryllimyt:
# Changes
- Remove hacky runtime-config.js, which only worked on the first load (for some unknown reasons...)
- Added next-runtime-env which we use to load `NEXT_PUBLIC_API_URL`


topher-lo (approved):
LGTM

PR #288, opened 2024-08-01
ci(infra): Add Caddy as reverse proxy

topher-lo:
## What changed
- Add Caddyfile to point http://localhost to UI service and http://localhost/api/ (with prefix stripping via handle path directive) to API service
- Add `BASE_DOMAIN` and `ADDRESS` env vars to docker-compose and .env.example
- Update public URLs in .env.example to the Caddy exposed URLs
- Specified internal routes for temporal and API in caddy
- Removed docker network configs in docker compose
- Style: just rearranged docker fields to be consistent between services (e.g. `restart` after `container_name`)
- !!! Removed overly verbose `tracecat_` prefix from container names

## QA
- Test on both `docker compose up` and `docker compose -f docker-compose.dev.yml up`
- Go to http://localhost (UI)
- Go to http://localhost/api/ (API)
- QA loop:
   1. Retrieve workflows
   2. Create new workflow
   3. Add case action to workflow
   4. Enable / commit workflow
   5. Trigger workflow from form in UI
   6. Check workflow runs success
   7. Check case is created
   8. Create schedule
   9. Wait for schedule to run
   10. Check workflow runs success
   11. Check case is created

## Screenshots

Screenshot 2024-08-01 at 1 18 12 PM

Screenshot 2024-08-01 at 1 35 28 PM

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,7 +23,6 @@ TRACECAT__API_URL=http://api:8000
 # to start ngrok and update this with the forwarding URL
 TRACECAT__PUBLIC_RUNNER_URL=http://localhost:8000
 # CORS (comman separated string of allowed origins)
-# NOTE: Defaults to `http://localhost:3000` for local development
 TRACECAT__ALLOW_ORIGINS=http://localhost:3000
 
 # --- CLI env vars ---
--- a/.env.example
+++ b/.env.example
@@ -2,6 +2,10 @@
 LOG_LEVEL=INFO
 COMPOSE_PROJECT_NAME=tracecat
 
+# -- Caddy env vars ---
+TRACECAT__API_DOMAIN=localhost:8000
+TRACECAT__UI_DOMAIN=localhost:3000
+
 # --- App and DB env vars ---
 # One of `development`, `staging`, or `production`
 TRACECAT__APP_ENV=development
--- a/.env.example
+++ b/.env.example
@@ -3,8 +3,9 @@ LOG_LEVEL=INFO
 COMPOSE_PROJECT_NAME=tracecat
 
 # -- Caddy env vars ---
-TRACECAT__API_DOMAIN=localhost:8000
-TRACECAT__UI_DOMAIN=localhost:3000
+BASE_DOMAIN=:80
+# Note: replace with your server's IP address
+ADDRESS=0.0.0.0
 
 # --- App and DB env vars ---
 # One of `development`, `staging`, or `production`
@@ -27,10 +28,10 @@ TRACECAT__API_URL=http://api:8000
 # to start ngrok and update this with the forwarding URL
 TRACECAT__PUBLIC_RUNNER_URL=http://localhost:8000
 # CORS (comman separated string of allowed origins)
-TRACECAT__ALLOW_ORIGINS=http://localhost:3000
+TRACECAT__ALLOW_ORIGINS=http://localhost:3000,http://localhost
 
 # --- CLI env vars ---
-TRACECAT__PUBLIC_API_URL=http://localhost:8000
+TRACECAT__PUBLIC_API_URL=http://localhost/api/
 
 # --- Postgres ---
 TRACECAT__POSTGRES_USER=postgres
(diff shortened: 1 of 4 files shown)


.env.example
```diff
@@ -42,9 +46,9 @@ TRACECAT__DB_URI=postgresql+psycopg://${TRACECAT__POSTGRES_USER}:${TRACECAT__POS
 NODE_ENV=development
 NEXT_PUBLIC_APP_ENV=development
 # The frontend app URL
-NEXT_PUBLIC_APP_URL=http://localhost:3000
+NEXT_PUBLIC_APP_URL=http://localhost
 # Allows the browser to communicate with the backend
-NEXT_PUBLIC_API_URL=http://localhost:8000
+NEXT_PUBLIC_API_URL=http://localhost/api/
 # Allows the frontend server (inside docker) to communicate with the backend server (inside docker)
 NEXT_SERVER_API_URL=http://api:8000
```
daryllimyt, line 53: Does this need to be changed?

docker-compose.dev.yml
```diff
@@ -26,17 +36,16 @@ services:
...
+      - core-app:/var/lib/tracecat
+
   worker:
     build:
       context: .
       dockerfile: Dockerfile.dev
-    volumes:
-      - ./tracecat:/app/tracecat
-      - core-app:/var/lib/tracecat
+    container_name: worker
```
daryllimyt: Any particular reason why the `container_name` was added back here? I believe we removed this when we were testing with replicating workers and you couldn't specify a name because it would autoincrement the suffix i.e. `workfer-N`

docker-compose.yml
```diff
@@ -20,16 +33,13 @@ services:
...
     volumes:
       - core-app:/var/lib/tracecat
-    networks:
-      - core
-      - temporal
 
   worker:
     image: ghcr.io/tracecathq/tracecat:${TRACECAT__IMAGE_TAG:-0.5.2}
-    # container_name: tracecat_worker
+    container_name: worker
```
daryllimyt: Same as above

docker-compose.yml
```diff
@@ -74,73 +81,58 @@ services:
...
-  core:
-  temporal:
+  temporal_ui:
+    image: temporalio/ui:${TEMPORAL__UI_VERSION}
+    container_name: temporal_ui
+    ports:
+      - 8080:8080
+    environment:
+      - TEMPORAL_ADDRESS=temporal:7233
+      - TEMPORAL_CORS_ORIGINS=http://localhost:3000
```
daryllimyt, line 132: same as above

docker-compose.dev.yml
```diff
@@ -86,74 +91,62 @@ services:
...
-  core:
-  temporal:
+  temporal_ui:
+    image: temporalio/ui:${TEMPORAL__UI_VERSION}
+    container_name: temporal_ui
+    ports:
+      - 8080:8080
+    environment:
+      - TEMPORAL_ADDRESS=temporal:7233
+      - TEMPORAL_CORS_ORIGINS=http://localhost:3000
```
daryllimyt, line 146: does networking change here? does referencing temporal here use caddy now?

.env.example
```diff
@@ -23,11 +28,10 @@ TRACECAT__API_URL=http://api:8000
 # to start ngrok and update this with the forwarding URL
 TRACECAT__PUBLIC_RUNNER_URL=http://localhost:8000
```
daryllimyt, line 29: does this need to change? this is used to construct the public facing webhook url (it should also have a more representive name)

daryllimyt (approved):
LGTM, some clarifications

.env.example
topher-lo: Yes because all localhost/api/ paths get redirected to the APi service by caddy.

docker-compose.dev.yml
topher-lo: Is mistake will remove

docker-compose.dev.yml
topher-lo: Yes caddy

.env.example
topher-lo: Can be changed to localhost/api/

.env.example
daryllimyt: @topher-lo referring to line 49 
`NEXT_SERVER_API_URL=http://api:8000`
or am i missing something here?

PR #289, opened 2024-08-02
fix(infra): Add root_path to FastAPI app

topher-lo:
- Fixes missing root_path in FastAPI app behind Caddy reverse proxy, which sets up prefixed /api* paths pointing to the api service.
- Also removed redundant internal routes in caddyfile (docker desktop has default bridge network implicitly defined)

## Screens
Screenshot 2024-08-01 at 11 57 35 PM


daryllimyt (approved):
LGTM

PR #292, opened 2024-08-06
feat: Replace Clerk auth with FastAPI users

daryllimyt:
# Features
- Add fastapi users cookie+session authentication
- Add modular `basic` (email+pw) and `google_oauth` login methods
- Add SQLModel adapter to bridge fastapi-users + sqlalchemy
- Update `authenticate_user` and `authenticate_user_or_service` to use the new auth mechanism
- Added auth cli
- Added auth fixture for tests
- Add google oauth (disabled github for now)

# Changes
- Removed Clerk-based JWT authentication
- Remove need for axios interceptors (bearer auth)
- Replace all fastapi auth dependencies with fastapi-users backed dependency
- Fix issue regarding future attached to different event loop in tests where either (in order of highest suspicion) `fastapi.TestClient`, `pytest-asyncio` or `fastapi-users` (not sure which exactly) causes the asyncio event loop to be changed during testing
- Remove unnecessary async usage in tests
- Add caddy to pytest


.env.example
```diff
@@ -1,6 +1,9 @@
 # --- Shared env vars ---
 LOG_LEVEL=INFO
 COMPOSE_PROJECT_NAME=tracecat
+PUBLIC_APP_URL=http://localhost
```
topher-lo: Repeated env vars. Please delete. See separate "Shared URL env vars" section

topher-lo (approved):
LGTM

PR #293, opened 2024-08-06
feat(integration): Add LDAP/AD find and enable / disable

mattdurant:
## Description

Adding ldap integration (uses ldap3 library, added to appropriate project files).
Actions implemented:

- Find user by email/username
- Disable AD user
- Enable AD user

N/A

## Screenshots/Recordings

N/A

## Steps to QA

Tested against a live Active Directory environment, and a test LDAP environment.


tracecat/actions/integrations/iam/ldap.py
```diff
@@ -0,0 +1,197 @@
...
+
+- name: `ldap`
+- keys:
+    - `LDAP_BIND_DN`
+    - `LDAP_BIND_PASS`
+"""
+
+
+class LdapClient:
+    _ldap_server = None
```
topher-lo: @mattdurant is there any reason why you've set these variables as class as opposed to instance attributes? So if it's declared here, these variables are scoped to the class and (may) introduce side effects across multiple instances.

For safety, it's better to define all attributes scoped at the instance level (i.e. under `def __init__`).

tracecat/actions/integrations/iam/ldap.py
```diff
@@ -0,0 +1,197 @@
...
+                        "attributes": json.loads(entry.entry_to_json())[
+                            "attributes"
+                        ],  # entry is CaseInsensitiveDict containing other types that cannot be serialized
+                    }
+                ]
+            return entries
+        else:
+            return []
+
+    def _searchone(self, base_dn: str):
```
topher-lo:
```suggestion
    def _search_one(self, base_dn: str):
```

topher-lo (changes requested):
Question on attributes scoping.

tracecat/actions/integrations/iam/ldap.py
mattdurant: @topher-lo I missed this; I initially built this integration as a separate script we used internally and rather than re-implement most of it I encapsulated the original script in a class and then refactored to match how other integrations are built. Corrected in latest commit.

tracecat/actions/integrations/iam/ldap.py
```diff
@@ -0,0 +1,192 @@
...
+- name: `ldap`
+- keys:
+    - `LDAP_BIND_DN`
+    - `LDAP_BIND_PASS`
+"""
+
+
+class LdapClient:
+    def __init__(
+        self, host: str, port: int, ssl: bool = False, active_directory: bool = False
```
topher-lo:
```suggestion
        self, host: str, port: int, use_ssl: bool = False, is_active_directory: bool = False
```

tracecat/actions/integrations/iam/ldap.py
```diff
@@ -0,0 +1,192 @@
...
+    ):
+        self._ldap_server = ldap3.Server(host, int(port), ssl, get_info=ldap3.ALL)
```
topher-lo:
```suggestion
        self._ldap_server = ldap3.Server(host, port=int(port), use_ssl=use_ssl, get_info=ldap3.ALL)
```

topher-lo (changes requested):
Style suggestions:
- Add `use_` and `is_` prefix to booleans
- Remove `self._ldap` in `self._xyz` to reduce verbosity

Then LGTM 👍

topher-lo (approved):

PR #294, opened 2024-08-07
Synchronous calling of webhooks

mattdurant:
## Description

Append /wait to a standard webhook to wait for the response of the workflow to return.  The final_context of the workflow will be returned as the response to the initial POST.

N/A

## Screenshots/Recordings

N/A

## Steps to QA

Call a webhook by appending /wait and the response should be the final_context of the workflow you called.

### DIFF
--- a/tracecat/api/routers/public/webhooks.py
+++ b/tracecat/api/routers/public/webhooks.py
@@ -46,7 +46,7 @@ async def incoming_webhook(
 
 
 @router.post("/{path}/{secret}/wait", tags=["public"])
-async def incoming_webhook_sync(
+async def incoming_webhook_wait(
     defn: Annotated[WorkflowDefinition, Depends(validate_incoming_webhook)],
     path: str,
     payload: dict[str, Any] | None = None,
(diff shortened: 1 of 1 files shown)


tracecat/api/routers/public/webhooks.py
```diff
@@ -43,3 +43,36 @@ async def incoming_webhook(
         enable_runtime_tests=enable_runtime_tests,
     )
     return response
+
+
+@router.post("/{path}/{secret}/wait", tags=["public"])
+async def incoming_webhook_sync(
```
daryllimyt: could we rename this to `incoming_webhook_wait`? `*_sync` might be slightly misleading implying that this function is a non-async function

daryllimyt (changes requested):
Just a small naming change request but otherwise LGTM

tracecat/api/routers/public/webhooks.py
mattdurant: > could we rename this to `incoming_webhook_wait`? `*_sync` might be slightly misleading implying that this function is a non-async function

done

daryllimyt (approved):
LGTM

PR #295, opened 2024-08-07
Enhance SMTP integration for STARTTLS

mattdurant:
## Description

Added separate environment variables for STARTTLS vs SSL for SMTP connections.

N/A

## Screenshots/Recordings

N/A

## Steps to QA

Send mail using a server that requires STARTTLS (O365, etc)


tracecat/actions/core/email.py
```diff
@@ -109,13 +110,16 @@ async def _send(
...
+            msg["Bcc"] = bcc
+        if cc is not None:
+            msg["Cc"] = cc
+        if reply_to is not None:
+            msg["Reply-To"] = reply_to
 
         try:
             config = self.smtp_config
-            if config["tls"] == "1":
+            if config["ssl"] == "1":
```
daryllimyt: if this is boolean the check here needs to change accordingly

tracecat/actions/core/email.py
```diff
@@ -125,6 +129,14 @@ async def _send(
             else:
                 server = smtplib.SMTP(config["host"], config["port"])
 
+            server.ehlo()
+
+            if config["tls"] == "1":
```
daryllimyt: this too

tracecat/actions/core/email.py
```diff
@@ -125,6 +129,14 @@ async def _send(
...
+                context = None
+                if config["ignore_cert_errors"] == "1":
```
daryllimyt: this too, and auth on L140

tracecat/actions/core/email.py
```diff
@@ -86,7 +86,8 @@ def smtp_config(self):
             "auth": os.environ.get("SMTP_AUTH", "0"),
             "username": os.environ.get("SMTP_USER", ""),
             "password": os.environ.get("SMTP_PASS", ""),
-            "tls": os.environ.get("SMTP_TLS", "0"),
+            "ssl": os.environ.get("SMTP_SSL", "0"),
+            "tls": os.environ.get("SMTP_STARTTLS", "0"),
```
daryllimyt: nit: can we make `auth`, `ssl`, `tls`, `ignore_cert_errors` booleans? e.g.
```python
"ssl": os.environ.get("SMTP_SSL", "0").lower() in ("1", "true")
```

daryllimyt (changes requested):
small changes!

tracecat/actions/core/email.py
daryllimyt: also, might be cleaner if we also pull these configs out into `tracecat.config` so that we could reuse this for auth email verification too

mattdurant:
@daryllimyt This pattern came from the existing Resend integration, I'm fine making these changes, but typically "boolean" environment variables are represented by 0 or 1, do you want to change that here, or just change to boolean when I read the vars to store the config during client creation?

daryllimyt:
@mattdurant  I'd prefer having environment variables loaded into the `tracecat.config` layer and transformed into whatever appropriate datatype. IMO its easier to reason about and share across other parts of the app -- unless you foresee a usecase that these variables may change over the app's lifetime and must be read dynamically from the environment then we can just leave it as is!

mattdurant:
@daryllimyt I moved the env vars to tracecat.config and use those instead

tracecat/config.py
```diff
@@ -106,3 +106,13 @@
...
 RETRY_MAX_WAIT_TIME = 60
 RETRY_STOP_AFTER_ATTEMPT = 5
+
+# SMTP Settings
+SMTP_HOST = os.environ.get("SMTP_HOST", "")
+SMTP_PORT = os.environ.get("SMTP_PORT", 25)
+SMTP_STARTTLS = os.environ.get("SMTP_STARTTLS", False)
+SMTP_SSL = os.environ.get("SMTP_SSL", False)
+SMTP_IGNORE_CERT_ERRORS = os.environ.get("SMTP_IGNORE_CERT_ERRORS", False)
+SMTP_AUTH = os.environ.get("SMTP_AUTH", False)
```
daryllimyt: Ah I might have not been too clear about this -- for these boolean env vars you're right they're typically stored as either "1" or "0" (some string value), which is why we'd do something like this to transform it into a boolean:
```python
SMTP_SSL = os.getenv("SMTP_SSL", "0").lower() in ("1", "true") # Defaults to `false`
# Same for SMTP_STARTTLS, SMTP_IGNORE_CERT_ERRORS, SMTP_AUTH
```
so that you can assign any of "true" (case insensitive) or "1" to the undelying env var to make it `true` and anything else will make it `false`. The way you've got it right now, checking the truthiness of the config var all strings will resolve to `true`.

Also, can we also make the names more descriptive so its more obvious that it's a boolean value and not some other kind of config - i.e.
`SMTP_AUTH` -> `SMTP_AUTH_ENABLED`
`SMTP_STARTTLS` -> `SMTP_STARTTLS_ENABLED`
`SMTP_SSL` -> `SMTP_SSL_ENABLED`

daryllimyt (changes requested):
@mattdurant  Looks good, just some final clarification on the changes

tracecat/config.py
mattdurant: @daryllimyt Do you want the names updated in the .env file as well?

daryllimyt: yes please, thanks!

mattdurant: > yes please, thanks!

done

daryllimyt (approved):
LGTM

PR #296, opened 2024-08-08
fix(integration): Require content-type is not None before check

mattdurant:
## Description

HTTP responses from the HTTP request action that do not contain a Content-Type header were causing exceptions.  Microsoft Teams' incoming webhooks do not return a Content-Type header and the workflow would not finish properly.

N/A

## Steps to QA

Make an HTTP request to a URL that does not return a Content-Type header.


topher-lo:
Good catch 👍

daryllimyt (approved):

PR #298, opened 2024-08-08
feat(integration): Exclude building block rules from Elastic list alerts

acumen-kevinr:
Typically you wouldn't want to raise tickets/cases for building block alerts in Elastic SIEM as there will be hundreds of these every minute/hour.


topher-lo (approved):
LGTM

topher-lo:
@acumen-kevinr linting fails as `pre-commit` is not yet installed in your local dev environment. But this seems to be a blocking issue for your elastic playbook, so we're going to make the fix first and merge it in 🤝

PR #303, opened 2024-08-10
fix(infra): Add "_enabled" to docker-compose.yml smtp envs

LuedTek:
## Description
When spinning up a new instance, I noticed warnings where environment variables are missing. Found that .env.example declares *_enabled for smtp booleans while docker-compose.yml has a typo for starttls and does not have *_enabled declared for the environment variable value.


topher-lo (approved):

topher-lo:
@mattdurant just checking in if this makes sense. Wondering how you got SMTP working during QA?

topher-lo (changes requested):
Waiting on response from SMTP contributor

mattdurant:
> @mattdurant just checking in if this makes sense. Wondering how you got SMTP working during QA? 

No this makes sense to me, I haven't retested since the renaming that was requested, the changes make sense to have everything align.

topher-lo (approved):

PR #304, opened 2024-08-11
feat(engine+infra): Implement DB migrations with alembic

topher-lo:
## What changed?
- Created empty initial migration (see note)
- Added github workflow to smoke test migration (i.e. load latest tagged version then recreate API service with latest code)
- Added entrypoint that automatically applies migrations on API service startup
- Modified default alembic `env.py` file to get `TRACECAT__DB_URI` from environment variables directly (as opposed to from the `alembic.ini` file

**Note:**
Our initial alembic migration does not contain any instructions as we autogenerate the migration against a database with 0.7.0 data models. This is because we use SQLModel `create_all` for initial schema creation. This is different to using Alembic to create an initial migration against an empty database.


alembic/versions/ebf7613506e0_initial_migration.py
```diff
@@ -0,0 +1,30 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision: str = 'ebf7613506e0'
+down_revision: Union[str, None] = None
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
```
daryllimyt, line 18: does alembic need support for python <3.10? if this is autogenerated, can ignore

daryllimyt (approved):
LGTM just small clarification

alembic/versions/ebf7613506e0_initial_migration.py
topher-lo: It's auto generated

PR #305, opened 2024-08-11
ci(infra): Docker network hardening & volume cleanup

LuedTek:
## Description
- Removes unnecessary local port mappings on most containers. Maintains them on dev.
- Isolates db networks from frontend network and sets internal: true on DB networks.
- removes core-app volume from prod and dev compose files.


topher-lo (changes requested):
No longer over-riding dev 🙏

topher-lo (approved):

PR #308, opened 2024-08-14
Update crowdstrike.py

acumen-kevinr:
Fixed wrong fields and updated capitalization. Untested


topher-lo:
Resolved by hot fix in 0.6.6 https://github.com/TracecatHQ/tracecat/releases/tag/0.6.6

PR #309, opened 2024-08-14
fix(engine): Allow object/attribute access for jsonpath

daryllimyt:
# Changes
- Extend Lark grammar to use recurvie `jsonpath_expression`
- Reason aboue jsonpaths as path segments of `ATTRIBUTE_ACCESS` and `BRACKET_ACCESS`
    - attribute access: `.cname`, `.'some_key'`, `.['bracket']`
    - bracket access: `['key']`, `[*]` (array wildcard), `[\d+]` (array index)
 - These changes also allow for repeated bracket access application: `.segment['key1']['key2']` (equivalent to .segment.key1.key2 if object, if array then it will index)
   

# Tests
- Added relevant expression unit tests
- Add test to check that we can use leading underscores for attribute access


topher-lo (approved):
LGTM

PR #328, opened 2024-08-21
feat(ui): Update wording and drop actions side panel

topher-lo:
## What changed
- Edited tool tips for "output schema" and "static inputs" sections in workflows settings
- Moved delete workflow button into nav bar under "export to json"
- Dropped actions side panel
- Configured Parameter column in schema inputs to be `min-w-max`
- Made "Input Schema" the default open form in actions settings 
- Fixed workflow breadcrumbs overflow

## QA
- Created, then deleted workflow with new button

## Screens
Screenshot 2024-08-20 at 9 51 42 PM

Screenshot 2024-08-21 at 12 17 33 PM


frontend/src/components/workbench/panel/workflow/form.tsx
```diff
@@ -333,17 +331,15 @@ function WorkflowReturnValueTooltip() {
...
         <span className="font-mono text-sm font-semibold">
-          Return Value Expression
+          Output Schema
         </span>
         <span className="text-xs text-muted-foreground/80">(optional)</span>
       </div>
       <span className="w-full text-muted-foreground">
-        Define an expression that will be evaluated and returned as the
-        workflow&apos;s output.
+        Key-value pairs that define the data returned by the workflow run.
```
daryllimyt: > Key-value pairs

This would make it seem like it can only return key-value pairs, but there may be certain use cases that you'd like to return a list or just a simple value so perhaps this should be reworded

frontend/src/app/workspaces/[workspaceId]/workflows/[workflowId]/page.tsx
```diff
@@ -10,26 +10,10 @@ export const metadata: Metadata = {
 export default function WorkbenchPage() {
   const layout = cookies().get("react-resizable-panels:layout")
   const defaultLayout = layout ? JSON.parse(layout.value) : undefined
-  const collapsed = cookies().get("react-resizable-panels:collapsed")
```
Comment (daryllimyt) [inline frontend/src/app/workspaces/[workspaceId]/workflows/[workflowId]/page.tsx:13]: Are these cookies not used by the other 2 thirds of the panel?

frontend/src/components/nav/workbench-nav.tsx
```diff
@@ -416,36 +429,78 @@ function WorkbenchNavOptions({
...
 }) {
+
+  const router = useRouter()
+  const { deleteWorkflow } = useWorkflowManager()
+
+  const handleDeleteWorkflow = async () => {
+    console.log("Delete workflow")
+    await deleteWorkflow(workflowId)
+    router.push(`/workspaces/${workspaceId}`)
+    router.refresh()
```
daryllimyt: I don't think you have to refresh here, the `deleteWorkflow` should clear the react-query query key inside and show updated data

daryllimyt (commented):
Added some thoughts

frontend/src/components/workbench/panel/workflow/form.tsx
topher-lo: Yes. But this doesn't have to be an expression right?

frontend/src/app/workspaces/[workspaceId]/workflows/[workflowId]/page.tsx
Comment (topher-lo) [reply frontend/src/app/workspaces/[workspaceId]/workflows/[workflowId]/page.tsx:13]: I don't think the other 2 thirds collapse. The collapse attribute is only relevant for the side panel?

frontend/src/components/nav/workbench-nav.tsx
topher-lo: I just copied the existing implementation:
https://github.com/TracecatHQ/tracecat/pull/328/files#diff-ffb1964b3632a8147cb3d858dfef3644c33d0e1493daca03283765543c6c53c6L36

frontend/src/components/workbench/panel/workflow/form.tsx
topher-lo: Updated

frontend/src/app/workspaces/[workspaceId]/workflows/[workflowId]/page.tsx
Comment (topher-lo) [reply frontend/src/app/workspaces/[workspaceId]/workflows/[workflowId]/page.tsx:13]: Screenshot 2024-08-21 at 12 18 48 PM

frontend/src/components/nav/workbench-nav.tsx
daryllimyt: fairs, if it works without can we remove the refresh?

frontend/src/app/workspaces/[workspaceId]/workflows/[workflowId]/page.tsx
Comment (daryllimyt) [reply frontend/src/app/workspaces/[workspaceId]/workflows/[workflowId]/page.tsx:13]: ok, i see

frontend/src/components/nav/workbench-nav.tsx
topher-lo: I'll try

topher-lo: Confirm works without

daryllimyt (approved):
LGTM

PR #331, opened 2024-08-21
feat(engine+ui): Move case management to workspace level

daryllimyt:
# Features
- Add link from each case to the workflow that created it

# Changes
- Create case management and case events services
- Move case management to workspace level in UI

# Todo
- Remove legacy `CaseAction` and `CaseContext` db schemas
- Remove some legacy zod schemas for cases

# Testing
Tested timeline, able to change status and add comments

# Screens

https://github.com/user-attachments/assets/36a525e7-0b41-46e8-a1bf-4bc08c5c0055


topher-lo (approved):
LGTM but still need to do migrations before next release. Let's do an early 0.8.0 before the weekend

PR #344, opened 2024-08-22
ci(infra): Post 0.7.x Fargate Terraform Stack

topher-lo:
Note: this PR modifies @iknowjason's PR #299 to align with versions 0.7 onwards

## What changed
- The majority of the stack was developed by Jason for versions < 0.6.x
- Set all assign_public_ip to False (no longer needed as we are point ALB directly to Caddy, then Caddy reverse proxy connects to UI and API via REST using service connect)
- 0.7 onwards adds Caddy and replaces a subdomain api / app strategy with an /api path.
- Consolidate into one hosted zone and one target listener to port 80 (caddy)
- Add caddy service with service connect to API and UI
- Consolidate variables and local env vars (separated by services as inspired by the original stack by @iknowjason)
- Consolidate security groups to: `alb`, `caddy`, `core` (for all core services apart from databases), `core_db` (tracecat services to tracecat database) , and `temporal_db` (temporal service to temporal database)
- Use A record instead of CNAME to point to hosted zone to ALB given AWS recommendations (given only on target listener exists in version 0.7.+)
- IMPORTANT: uses `managed_master_user_password` option in RDS resource to avoid plaintext passwords being logged https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance#managed-master-passwords-via-secrets-manager-default-kms-key

## QA
- `terraform init` and `terraform apply`
- All services start and are healthy
- You can access Tracecat UI via the configured domain
- Can create workflow (API and core db operational)
- Can run workflow (temporal and temporal worker operational)


daryllimyt (approved):
:shipit:

PR #345, opened 2024-08-22
chore(engine): Run migration for 0.7.3

daryllimyt:
## What changed
- Run migrations (see Alembic revision message)
- Fixes entrypoint alembic CLI
- Drop outdated CaseAction and CaseContext tables


topher-lo (approved):
LGTM

PR #354, opened 2024-08-24
ci(infra): Add persistence to EC2 stack

topher-lo:
## Why
Current stack does not have persistence. If the EC2 instance dies, so do the the encryption keys (saved in the `.env` file) and postgres data

## What changed
- Add persistence via EFS
- Update `user_data.tpl` to only run `env.sh` if `.env` doesn't yet exist.

## QA
- Ran terraform init and apply
Screenshot 2024-08-25 at 1 39 04 PM

- Check access, can create workflow, ran workflow successfully (simple two action reshape)
Screenshot 2024-08-25 at 1 41 26 PM

- Stopped instance, restarted instance, checked that workflow still exists
Screenshot 2024-08-25 at 1 43 20 PM

- Check EFS exists
Screenshot 2024-08-25 at 1 42 06 PM


daryllimyt (approved):
LGTM

PR #361, opened 2024-08-27
Sublime Integration

lued:
## Description

Implemented a playbook and integration for Sublime Email Security. There is a bug with the way the for loops iterate through all the messages since we can't perform nested loops. Just PRing so we can add a feature and provide value for others.

There is a change to the way webhooks are implemented with this PR - since the only way to get free webhook alerts from sublime is through slack's legacy Outgoing Webhook which uses query syntax, not JSON. 

Documentation may be lacking. Feel free to reject if you guys want me to write it up.


topher-lo:
Closing in favor of new PR to write sublime integrations into action templates

PR #377, opened 2024-09-03
feat(engine+ui): Implement runtime environments for secrets

### LINKED ISSUE #271: [FEATURE IDEA] Multi-client integrations within same workflow
I want to use multiple instances (i.e. different inputs / credentials) of the same type of integration (e.g. list SIEM alerts) in a workflow.

> Example: real world example would be that I am an M365 customer, and I acquire another business that is also an M365 customer. We are very likely not going to be able to transition all mailboxes and users off of that environment on day 1, but if I'm using a workflow where on a certain type of alert I want to reset a user's password and then contact them, how would I interact with both M365 environments in the same workflow since all users won't exist in one or the other?

_Originally posted by @mattdurant in https://github.com/TracecatHQ/tracecat/issues/268#issuecomment-2251099046_

daryllimyt:
# Features
- Implement runtime environments and Improve API for runtime configuration (generalization) in the general form of `DSLRunArgs` and `DSLConfig`. Runtime environment form an isolation boundary for certain resources in Tracecat, e.g. credentials.
- Implement `Configuration` field in the workbench-workflow view, where you can set runtime configuration values (`DSLConfig`). This is where users can set the environment value for a workflow.
- Implement runtime environments for secrets
    - Update `Secret` schema to take an environment string
    - Add unique constraint on (secret.name, secret.owner_id, secret.environment). Throws if all 3 fields match.
    - This influences how the secrets manager (SM) behaves.
        - At runtime, SM will search for the current environment set in the workflow context path `ENV.environment`.
        - If the above isn't found, it will throw an error
- Implement UI to create and modify secret environments
- A workflow may override the runtime environment for a child workflow in `core.workflow.execute` by specifying `environment` as an argument. 
  - This supports the use case of managing multiple copies of the same secret over multiple environments or clients/managed users (closes #271). 
  - If `environment` isn't provided, the child workflow will use its default environment specified in its workflow definition

# Changes
- Make `ActionStatement` accept generic `Mapping` for better DX when working with child workflows
- Fix some mypy linting errors
- Add graceful retries when connecting to temporal client
- Move `WorkspaceUserRole`, `WorkspaceUserOrServiceRole` and `AsyncDBSession` dependencies into tracecat.api.dependencies
- Set `exclude_unset=True` for `jsonable_encoder` in `PydanticORJSONPayloadConverter`. We do this to preserve information about what's fields are set when crossing a Temporal boundary. This way, Temporal can reconstruct which fields were set in `DSLRunArgs` when it was pass into the `DSLWorkflow 

# Testing
Added tests for the following use cases
- [x] Set environment at start and verify that it's correct
- [x] Configuring workflow environment directly
    - [x]  Test that we can set the runtime environment for a workflow. The workflow should use the environment set in the DSL config.
    - [x] Test that we can override the runtime environment for a workflow from its run_args. The workflow should use the environment passed in the run_args.
    - [x] Test that we can set the default runtime environment for a workflow. The workflow should use the default runtime environment 'default'.
- [x] Child workflow run
  - [x] Check that child workflow env can be set by the parent and is correct
    - [x] Single child
    - [x] Loop
  - [x] Check that child workflow env is set to DSL default environment if it isn't set by the parent
    - [x] Single child
    - [x] Loop
- [x] Check that we can grab secrets from different environments
- [x] Unit (update unit tests for sandbox)
- [x] Integration (ran as a workflow)


frontend/src/components/workbench/panel/workflow/form.tsx
```diff
@@ -55,6 +56,17 @@ import { CustomEditor } from "@/components/editor"
 const workflowConfigFormSchema = z.object({
   title: z.string(),
   description: z.string(),
+  config: z.string().transform((val, ctx) => {
+    try {
+      return YAML.parse(val) || {}
+    } catch (error) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: "Invalid YAML format",
```
topher-lo:
```suggestion
        message: "Invalid YAML format. Please check workflow definition for errors.",
```

frontend/src/lib/hooks.ts
```diff
@@ -633,12 +633,22 @@ export function useSecrets() {
...
-        title: "Failed to add new secret",
-        description: "An error occurred while adding the new secret.",
-      })
+    onError: (error: TracecatApiError) => {
+      switch (error.status) {
+        case 409:
+          console.error("Secret already exists", error)
+          toast({
+            title: "Secret already exists",
+            description: "A secret with matching name and environment exists.",
```
topher-lo:
```suggestion
            description: "Secrets with the same name and environment are not supported.",
```

frontend/src/lib/hooks.ts
```diff
@@ -633,12 +633,22 @@ export function useSecrets() {
...
+          })
+          break
+        default:
+          console.error("Failed to create secret", error)
+          toast({
+            title: "Failed to add new secret",
+            description: "An error occurred while adding the new secret.",
```
topher-lo:
```suggestion
            description: "Please contact support for help.",
```

tests/conftest.py
```diff
@@ -308,7 +308,12 @@ def test_workspace(
         if e.response.status_code == 409:
             logger.info("Test workspace already exists")
             workspace = test_config_manager.get_workspace()
-            yield WorkspaceMetadataResponse(**workspace)
+            if not workspace:
+                raise ValueError(
+                    "Unexpected error when trying to get workspace. It either doesn't exist, or there was a conflict."
+                    "Please check the logs for more information, or try viewing the database directly."
```
topher-lo, line 314:
```suggestion
                    "Please contact us for support (Cloud) or check the logs for more information (self-hosted)."
```

tests/conftest.py
```diff
@@ -308,7 +308,12 @@ def test_workspace(
         if e.response.status_code == 409:
             logger.info("Test workspace already exists")
             workspace = test_config_manager.get_workspace()
-            yield WorkspaceMetadataResponse(**workspace)
+            if not workspace:
+                raise ValueError(
+                    "Unexpected error when trying to get workspace. It either doesn't exist, or there was a conflict."
```
topher-lo:
```suggestion
                    "Unexpected error when retrieving workspace. Workspace either doesn't exist or there was a conflict."
```

tracecat/actions/core/workflow.py
```diff
@@ -35,6 +29,16 @@ async def execute(
             description="The inputs to pass to the child workflow.",
         ),
     ],
+    environment: Annotated[
+        str | None,
+        Field(
+            description=(
+                "The child workflow's target execution environment. "
+                "This is used as an isolation boundary for credentials and other secrets."
```
topher-lo:
```suggestion
                "This is used to isolate secrets across different environments."
```

tracecat/auth/sandbox.py
```diff
@@ -35,38 +33,51 @@ def __init__(
...
-    def __exit__(self, exc_type, exc_value, traceback) -> None:
+    def __exit__(
+        self,
+        exc_type: type[BaseException],
+        exc_value: BaseException,
+        traceback: TracebackType,
+    ) -> None:
         self._unset_secrets()
         if exc_type is not None:
             logger.error("An error occurred", exc_info=(exc_type, exc_value, traceback))
```
topher-lo: More information error with suggested action to follow up?

tracecat/auth/sandbox.py
```diff
@@ -106,99 +117,37 @@ def _unset_secrets(self):
...
+                SearchSecretsParams(
+                    names=list(unique_secret_names), environment=self._environment
+                )
+            )
+
+        if len(unique_secret_names) != len(secrets):
+            missing_secrets = unique_secret_names - {secret.name for secret in secrets}
             raise TracecatCredentialsError(
-                "Failed to retrieve secrets", detail=missing_secret_names
+                "Failed to retrieve secrets",
```
topher-lo:
```suggestion
                "Failed to retrieve secrets. Please contact support for help.",
```

tracecat/dsl/models.py
```diff
@@ -60,11 +63,23 @@ def title(self) -> str:
...
+    )
     enable_runtime_tests: bool = Field(
         default=False,
         description="Enable runtime action tests. This is dynamically set on workflow entry.",
     )
+    environment: ExpressionStr = Field(
+        default=DEFAULT_SECRETS_ENVIRONMENT,
+        description=(
+            "The workflow's target execution environment. "
+            "This is used as an isolation boundary for credentials and other secrets."
```
topher-lo:
```suggestion
            "This is used to isolate secrets across different environments."
```

tracecat/secrets/service.py
```diff
@@ -61,54 +63,79 @@ async def _delete_secret(self, secret: Secret) -> None:
...
+        try:
+            return result.one()
+        except MultipleResultsFound as e:
+            if raise_on_error:
+                raise MultipleResultsFound(
+                    "Multiple secrets found when searching by ID"
+                ) from e
+        except NoResultFound as e:
+            if raise_on_error:
+                raise NoResultFound("Secret not found when searching by ID") from e
```
topher-lo:
```suggestion
                raise NoResultFound("Secret not found when searching by ID. Please check that the ID was correctly input.") from e
```

tracecat/secrets/service.py
```diff
@@ -61,54 +63,79 @@ async def _delete_secret(self, secret: Secret) -> None:
...
-        secret = result.one_or_none()
-        if not secret and raise_on_none:
-            raise NoResultFound("Secret not found when searching by name")
-        return secret
+        try:
+            return result.one()
+        except MultipleResultsFound as e:
+            if raise_on_error:
+                raise MultipleResultsFound(
+                    "Multiple secrets found when searching by name"
```
topher-lo:
```suggestion
                    "Multiple secrets found when searching by name. Expected one secret only."
```

tracecat/secrets/service.py
```diff
@@ -61,54 +63,79 @@ async def _delete_secret(self, secret: Secret) -> None:
...
+                ) from e
+        except NoResultFound as e:
+            if raise_on_error:
+                raise NoResultFound("Secret not found when searching by name") from e
```
topher-lo:
```suggestion
                raise NoResultFound("Secret not found when searching by name. Please double check that the name was correctly input.") from e
```

topher-lo (changes requested):
Minor suggestions regarding error messages.

In general, we should take inspiration from resend: https://resend.com/docs/api-reference/errors.

Overall, code seems to match with write up.

Great implementation 🚢

tests/conftest.py
daryllimyt: This is conftest - i wrote this message more for devs

topher-lo (approved):

tracecat/auth/sandbox.py
daryllimyt: Doesn't catch a specific error here, so can't really add anything constructive

daryllimyt:
```suggestion
            logger.error("An error occurred inside AuthSandbox. If you are seeing this, please contact support.", exc_info=(exc_type, exc_value, traceback))
```

PR #380, opened 2024-09-05
feat(integration): Add mongodb doc find one getter 

benacumen:
## Description

This PR adds an integration for MongoDB. This integration allows you to query a MongoDB instance.

By supplying a database name, collection name, search field and search value, you can return either the entire resulting document, or by supplying a comma-separated list of fields, you can return specific fields from the document.

## Steps to QA

1. Verify code is accurate 
2. Verify functionality of integration (MongoDB can be trialed for free)

### DIFF
--- a/tracecat/actions/integrations/database/mongodb.py
+++ b/tracecat/actions/integrations/database/mongodb.py
@@ -1,15 +1,17 @@
 """MongoDB Integration.
-    
-    This integration will be used to query MongoDB for data.
-    
-    Authentication: MongoDB Connection String
-    
+
+This integration will be used to query MongoDB for data.
+
+Authentication: MongoDB Connection String
+
 """
 
 import os
+from typing import Annotated, Any
+
 from pymongo import MongoClient
+
 from tracecat.registry import Field, RegistrySecret, registry
-from typing import Annotated, Any, Optional
 
 mongo_secret = RegistrySecret(
     name="mongodb",
@@ -19,10 +21,11 @@ mongo_secret = RegistrySecret(
 """MongoDB Secret.
 
 - name: `mongodb`
-- key: 
+- key:
     - `MONGODB_CONNECTION_STRING`
 
 """
+
 
 @registry.register(
     default_title="Query MongoDB",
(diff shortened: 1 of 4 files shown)


tracecat/actions/integrations/database/mongodb.py
```diff
@@ -0,0 +1,87 @@
...
+    if return_field:
+        # Split the comma-separated fields and build the projection dict
+        fields = return_field.split(",")
+        projection = {field.strip(): 1 for field in fields}
+    else:
+        projection = None  # No projection, return entire document
+
+    result = collection.find_one({search_field: search_value}, projection)
+
+    if result:
```
topher-lo, line 83: @benacumen is the error message `return {"error_msg": "No results returned from MongoDB."}` required? A null return in mongodb and other database clients for empty results is considered expected behavior.

Could we just return `result` as is without the `error_msg`? This aligns with the return type: `dict[str, Any] | None` so we stay as close to expected behavior with `pymongo` as possible. The conditional `if result` won't be necessary either in this case i.e. we let the user deal with conditionals within Tracecat instead of having to "check" for `result.error_msg == "something"`. They can just assume (as in pymongo world) that `not result.error_msg` means no results found.

Thanks!

topher-lo (changes requested):
LGTM. Just one small suggestion

tracecat/actions/integrations/database/mongodb.py
```diff
@@ -0,0 +1,87 @@
...
+        result["_id"] = str(result["_id"])  # Convert ObjectId to String
+        return result
+    else:
+        return {"error_msg": "No results returned from MongoDB."}
```
topher-lo:
```suggestion
        return {}
```

@benacumen thanks for reporting the `dict[str,, Any] | None` error. In the meantime, shall we do this instead? Since `bool({})` is still falsy and evaluates to false in the conditional expression

topher-lo (changes requested):

topher-lo (approved):
LGTM!

PR #383, opened 2024-09-07
ci(infra): Drop EC2 Terraform stack

topher-lo:
## What changed?
- Removed EC2 terraform deployment option
- Neither production ready nor was it easy to maintain (EC2 user data runs docker compose with minimal logs for debugging)
- More users have existing VMs that they SSH into and then set up Tracecat from inside the VM
- Added better guidance directly in the docker compose docs


daryllimyt (approved):
🚢

PR #384, opened 2024-09-07
feat(ui): Add suggestive action to user invite fail error

topher-lo:
Self-explanatory. Part of effort to help users help themselves directly in the product.


daryllimyt (approved):
🚢

PR #392, opened 2024-09-10
feat(integrations): Capture ReadTimeout exceptions in httpx as "expected" errors

topher-lo:
As per title
Screenshot 2024-09-10 at 1 00 03 PM


daryllimyt (approved):
🚢

PR #394, opened 2024-09-11
feat(integration): Add jwt token getter to core http action

topher-lo:
## Use-case
- Limacharlie API auth requires two calls: first call gets JWT token, second call makes the operational API call.
- Current httpx action does not support jwt calls without logging jwt to outputs in another httpx core action

## Alternatives considered
- Write a Limacharlie integration to the registry. Decided against as there are too many Limacharlie APIs.
- I want to give flexibility to the end user to modify the templated http action without contacting support / writing python and opening a PR


tracecat/actions/core/http.py
```diff
@@ -20,6 +20,15 @@ class HTTPResponse(TypedDict):
     data: str | dict[str, Any] | list[Any] | None
 
 
+def get_jwt_token(
+    url: str,
+    json: dict[str, str] | None = None,
+    headers: dict[str, str] | None = None,
+) -> str:
+    token = httpx.get(url, json=json, headers=headers).json()["token"]
+    return token
```
daryllimyt: This should be async, otherwise you block the entire worker thread across concurrent workflows/actions in the GET

daryllimyt (changes requested):
small change

tracecat/actions/core/http.py
daryllimyt: Perhaps add some error handling on this as well

daryllimyt (approved):

Issue #375, opened 2024-09-03
Kubernetes readiness


pmig:
**Is your feature request related to a problem? Please describe.**
Tracecat has currently no easy installation method.

**Describe the solution you'd like**
Create a Glasskube Kubernetes package.

**Additional context**
Kubernetes package can be created based on https://docs.tracecat.com/self-hosting/deployment-options/aws-ecs

topher-lo:
@pmig This would be an incredible addition! Been following Glasskube for a while now and seems super promising.

topher-lo:
Adding a few more resources that will be relevant:
- Temporal's helm chart: https://github.com/temporalio/helm-charts

Also would be interested in knowing how Glasskube (if) simplifies upgrades, persistence, and secrets rotation

pmig:
In order to use temporal helm charts we created an upstream fix to make sure upgrading the helm chart will not result in a failure. As database migrations are performed as job that is not cleaned up correctly, the second update will not go through as the original job is still there. Adding a ttl on the job resolves the issue. See:

 - https://github.com/temporalio/helm-charts/pull/555

PR #396, opened 2024-09-11
ci(infra): add glasskube support

christophenne:
## Description

This PR adds kubernetes manifests for tracecat and its components, such that it can be installed via glasskube.

See #375


topher-lo:
Too much drift with core. Will revisit when issue comes up again. For now, Fargate and Docker compose are the only officially supported deployment options.

PR #401, opened 2024-09-15
feat(integration): Support OAuth2 flows in http core action

topher-lo:
## What changed
- Support passing OAuth parameters into http core action
- API design inspired by Cribl collector templates and Tines OAuth2 credentials type
- Updated Limacharlie playbooks (set `token_response_key` to `token`)

### DIFF
--- a/tracecat/actions/core/http.py
+++ b/tracecat/actions/core/http.py
@@ -5,6 +5,7 @@
 from typing import Annotated, Any, Literal, TypedDict
 
 import httpx
+from authlib.integrations.httpx_client import AsyncOAuth2Client
 from loguru import logger
 from pydantic import Field, UrlConstraints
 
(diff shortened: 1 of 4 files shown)


tracecat/actions/core/http.py
```diff
@@ -88,28 +116,72 @@ async def http_request(
...
+            client_secret=oauth2_client_secret,
+            grant_type=oauth2_grant_type,
+            token_response_key=token_response_key,
+            scope=oauth2_scope,
+            headers=token_request_headers,
+            payload=oauth2_payload,
+        )
+
+    if access_token is not None:
         key = auth_header.keys()[0]
```
daryllimyt: can `auth_header` ever be an empty list? is there always 1 item in here?

daryllimyt (approved):
LGTM, needs clarification on one part

PR #402, opened 2024-09-16
ci: add publishing glasskube manifest on tag

kosmoz:
## Description

This PR adds a new GitHub workflow that builds the Glasskube deployment manifest and uploads it as release artifact. 

**Important:** Requires #396

Related to: #375
Requires: #396

## Steps to QA

Difficult, but in a fork you can create a new release (with a tag like "a.a.a") and check the actions tab (you can cancel any other workflows, they are not needed).

### DIFF
--- a/.github/workflows/release-glasskube.yml
+++ b/.github/workflows/release-glasskube.yml
@@ -1,0 +1,23 @@
+name: Publish Glasskube Manifest
+on:
+  push:
+    tags:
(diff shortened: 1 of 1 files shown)


topher-lo:
Too much drift with core. Will revisit when issue comes up again. For now, Fargate and Docker compose are the only officially supported deployment options.

PR #408, opened 2024-09-23
feat(engine)!: Implement actions registry and template action management

daryllimyt:
# Goals
- ~Process-level isolation across action executions (better handling of multi-environment race conditions)~
  - Solved this using custom `secrets` manager in tracecat registry sdk
  - We wil only require process-level isolation if we require cpu intensive jobs
- Make `Registry` instance based, not singleton
- `Registry` is now versioned
- Workflow definitions are now also versioned.
- `Registry` will contain arbitrary UDFs. The base registry that ships with `tracecat` only contains the UDFs we manage
  - We will start managing different versions of registries moving forward
- Split registry into 2 parts:
  1. The marking logic that decorates UDFs -> `tracecat_registry` package
    - The registry package should be super lightweight and serves as the "SDK" that you can use to register custom UDFs
  3. The loading logic that binds UDFs to a `Registry` instance -> `tracecat.registry` module
- ~Implement registry action executor using `cloudpickle`~ 
  - We decouple action implementations from the worker. Enforce that workers' main responsibilities are just orchestration of tasks, resolving expressions, ensuring that actions get run until completion. 
    - Added benefit of having programming language agnostic tasks, or running 
- Implement `RegistryManager` to manage lifecycle of different `Registry` instances.
- Serve registry service from API service to minimize devops overhead. If scaling requirements change, we can detach this from the application.
  - Expose endpoints for registry CRUD operations
  - Expose endpoints to run an action in a worker process pool

# Todos
- [x] Make API calls from worker to API to run actions
- [x] Do something about registry udfs that depend on platform code (e.g. `core.transform.filter`) (we're allowing it)


registry/pyproject.toml
```diff
@@ -30,7 +30,8 @@ dependencies = [
     "dnspython==2.6.1",
     "httpx==0.27.0",
     "ldap3==2.9.1",
-    "polars-lts-cpu==1.2.0",
+    "openai==1.30.3",
+    "polars==1.7.1",
```
daryllimyt: UNDO this

PR #433, opened 2024-10-15
fix(engine): Add retries to recreate DB engine if test conn raises an exception

topher-lo:
## Rationale

Our current database connection handling doesn't adequately address potential connection failures, especially after the initial engine creation. This can lead to unexpected errors and poor user experience when database connections become stale or are interrupted. Additionally, we were using a deprecated method for executing queries with SQLModel.

Moreover, if the database password changes, we also cannot recover from a connection failure.

## What changed

1. Updated both `get_session` and `get_async_session` functions to test the connection immediately after creation.
2. Implemented logic to recreate the engine if a connection test fails.
3. Replaced the deprecated `session.execute()` method with `session.exec()` for SQLModel compatibility.
4. Added retry decorators to `get_session_context_manager` and `get_async_session_context_manager` for additional resilience.
5. Ensured consistent error handling and retry logic between synchronous and asynchronous database operations.

Key code changes:
- Added connection testing in `get_session` and `get_async_session`
- Updated query execution to use `session.exec(select(1))` instead of `session.execute("SELECT 1")`
- Implemented engine recreation on connection failure
- Added retry decorators to context manager functions

## How to QA

Go into Fargate. Force DB key rotation. Check if workflows still run without restarting the API container.

### DIFF
--- a/tracecat/db/engine.py
+++ b/tracecat/db/engine.py
@@ -203,24 +203,10 @@ async def get_async_session() -> AsyncGenerator[AsyncSession, None]:
 
 
 def get_session_context_manager() -> contextlib.AbstractContextManager[Session]:
-    try:
-        return contextlib.contextmanager(get_session)()
-    except SQLAlchemyError:
-        logger.warning(
-            "Database error occurred. Attempting to recreate engine with potentially updated configuration."
-        )
-        get_engine(force_recreate=True)
-        raise
+    return contextlib.contextmanager(get_session)()
 
 
-async def get_async_session_context_manager() -> (
+def get_async_session_context_manager() -> (
     contextlib.AbstractAsyncContextManager[AsyncSession]
 ):
-    try:
-        return contextlib.asynccontextmanager(get_async_session)()
-    except SQLAlchemyError:
-        logger.warning(
-            "Database error occurred. Attempting to recreate async engine with potentially updated configuration."
-        )
-        await get_async_engine(force_recreate=True)
-        raise
+    return contextlib.asynccontextmanager(get_async_session)()
--- a/tracecat/db/engine.py
+++ b/tracecat/db/engine.py
@@ -1,16 +1,14 @@
 import contextlib
 import json
 import os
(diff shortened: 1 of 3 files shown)


tracecat/db/engine.py
```diff
@@ -122,33 +130,59 @@ def _create_async_db_engine() -> AsyncEngine:
...
 
 
-def get_async_engine() -> AsyncEngine:
+@retry(
+    stop=stop_after_attempt(3),
+    wait=wait_fixed(1) + wait_random(0, 1),
+    retry=retry_if_exception_type(SQLAlchemyError),
+    reraise=True,
+)
+async def get_async_engine(force_recreate=False) -> AsyncEngine:
```
daryllimyt, line 162: Missing type hints

daryllimyt: missing type hints

tracecat/db/engine.py
```diff
@@ -122,33 +130,59 @@ def _create_async_db_engine() -> AsyncEngine:
...
 
 
-def get_engine() -> Engine:
+@retry(
+    stop=stop_after_attempt(3),
+    wait=wait_fixed(1) + wait_random(0, 1),
+    retry=retry_if_exception_type(SQLAlchemyError),
+    reraise=True,
+)
+def get_engine(force_recreate=False) -> Engine:
```
daryllimyt: missing type hints

tracecat/db/engine.py
```diff
@@ -157,7 +191,7 @@ def get_session_context_manager() -> contextlib.AbstractContextManager[Session]:
     return contextlib.contextmanager(get_session)()
 
 
-def get_async_session_context_manager() -> (
+async def get_async_session_context_manager() -> (
```
daryllimyt: this looks wrong as we are creating an async context manager as the output of the function, not invoking it as a coroutine
```suggestion
def get_async_session_context_manager() -> (
```

daryllimyt (changes requested):

PR #437, opened 2024-10-17
ci(infra): Fix issue with no snapshot terraform apply

topher-lo:
Fixes plan error when no snapshots exists (i.e. cold start Terraform apply)

image


daryllimyt (approved):
Not familiar with this area but trust that this works. LGTM!

PR #446, opened 2024-10-20
fix(integration): Incorrect Jira referencing secret

topher-lo:
## Fixes
Incorrect reference of secrets (it.s SECRETS. not secrets.) and 2. name for the formula `to_base64` not `str_to_b64`


daryllimyt (approved):
LGTM

PR #447, opened 2024-10-21
feat(engine): Add SAML SSO endpoints

topher-lo:
A glorious day

https://github.com/user-attachments/assets/b360ecdb-83d1-4e72-b99f-e00e8b2f1058


frontend/src/app/auth/saml/acs/route.ts
```diff
@@ -0,0 +1,53 @@
...
+    console.error("No set-cookie header found in response")
+    return NextResponse.redirect(new URL("/auth/error", getDomain(request)))
+  }
+
+  console.log("Redirecting to / with GET")
+  const redirectUrl = new URL("/", getDomain(request))
+  const redirectResponse = NextResponse.redirect(redirectUrl, {
+    status: 303 // Force GET request
+  })
+  console.log("Setting set-cookie header:", setCookieHeader)
```
daryllimyt:
```suggestion
```
Might have sensitive data?

frontend/src/components/auth/saml.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Button } from "@/components/ui/button"
+import { Icons } from "@/components/icons"
+
+type SamlSSOButtonProps = ComponentPropsWithoutRef<typeof Button>
+export function SamlSSOButton(props: SamlSSOButtonProps) {
+  const [isLoading, setIsLoading] = useState<boolean>(false)
+  const handleClick = async () => {
+    try {
+      setIsLoading(true)
+      window.location.href = authConfig.samlAuthorizationUrl
```
daryllimyt: Should we not serve this from the backend, and allow this to be configured from within the app? I see that this points to a backend url, should we not just hit this endpoint instead (we have autogenerated client code for this)? It made sense to redirect to the IDP when we were using ssoready, but we're now redirecting the page to the backend url

frontend/src/config/auth.ts
```diff
@@ -5,4 +5,5 @@ export const authConfig = {
     .split(",")
     .map((x) => x.toLowerCase()),
   staleTime: 5 * 60 * 1000, // 5 minutes
+  samlAuthorizationUrl: env("NEXT_PUBLIC_SAML_AUTHORIZATION_URL") || "http://localhost/api/auth/saml/login",
```
daryllimyt: see above comment

tracecat/auth/saml.py
```diff
@@ -0,0 +1,289 @@
...
+            config = Saml2Config()
+            config.load(saml_settings)
+
+    client = Saml2Client(config)
+    return client
+
+
+@router.get("/login", name=f"saml:{auth_backend.name}.login")
+async def login():
+    client = create_saml_client()
```
daryllimyt: can we factor this out into a dependency like:

```python
SamlClient = Annotated[<client type>, Depends(create_saml_client)]
```

Then in the route:
```python
async def login(client: SamlClient):
   ...

File: tracecat/auth/saml.py [hunk 5]
```diff
@@ -0,0 +1,289 @@
...
+    request: Request,
+    *,
+    SAMLResponse: str = Form(...),  # Changed from saml_response to SAMLResponse
+    user_manager: UserManagerDep,
+    strategy: AuthBackendStrategyDep,
+):
+    """Handle the SAML SSO response from the IdP post-authentication."""
+
+    # Get email in the SAML response from the IdP
+    client = create_saml_client()
```
daryllimyt: same as above

tracecat/auth/users.py
```diff
@@ -193,11 +233,16 @@ async def list_users(*, session: SQLModelAsyncSession) -> list[User]:
 
 
 def default_admin_user() -> UserCreate:
+    if not (email := os.getenv("TRACECAT__SETUP_ADMIN_EMAIL")):
+        raise ValueError("TRACECAT__SETUP_ADMIN_EMAIL is not set")
+    if not (password := os.getenv("TRACECAT__SETUP_ADMIN_PASSWORD")):
+        raise ValueError("TRACECAT__SETUP_ADMIN_PASSWORD is not set")
+
```
daryllimyt, line 240: i know i wrote this, but does this break the app on restart if you set the initial admin user at the start, then throw away these env vars?

tracecat/auth/saml.py
```diff
@@ -0,0 +1,289 @@
...
+@router.get("/login", name=f"saml:{auth_backend.name}.login")
+async def login():
+    client = create_saml_client()
+    _, info = client.prepare_for_authenticate()
+    try:
+        headers = info["headers"]
+        # Select the IdP URL to send the AuthN request to
+        for k, v in headers:
+            if k == "Location":
+                redirect_url = v
```
daryllimyt: nit: perhaps more concise, and it'll raise `StopIteration` if no `location` is found
```suggestion
        redirect_url = next(v for k, v in headers if k == "location")
```

tracecat/auth/saml.py
```diff
@@ -0,0 +1,289 @@
...
+    if SAML_IDP_METADATA_URL is None:
+        with generate_saml_metadata_file() as tmp_metadata_path:
+            # Add the local metadata file to the settings
+            saml_settings["metadata"] = {"local": [tmp_metadata_path]}
+            config = Saml2Config()
+            config.load(saml_settings)
+    else:
+        # Save the cert to a temporary file
+        with tempfile.NamedTemporaryFile(
+            mode="w+", suffix=".crt", delete=False
```
daryllimyt: wondering if its ok to skip cleaning up these certs? a nice pattern could be we turn this function into a fastapi dependency and `yield` inside this context manager - then when the fastapi router is done handling this route, it will cleanup the file

tracecat/auth/saml.py
```diff
@@ -0,0 +1,289 @@
...
+        tree = ET.ElementTree(root)
+        tree.write(tmp_file, encoding="unicode", xml_declaration=True)
+        tmp_file.flush()
+        tmp_file_path = tmp_file.name
+
+    try:
+        yield tmp_file_path
+    finally:
+        # Clean up the temporary file
+        os.unlink(tmp_file_path)
```
daryllimyt: nit: you could probably just do this, to the same effect
```suggestion
    # Create a temporary file
    with tempfile.NamedTemporaryFile(mode="w+", suffix=".xml") as tmp_file:
        # Write the XML to the temporary file
        tree = ET.ElementTree(root)
        tree.write(tmp_file, encoding="unicode", xml_declaration=True)
        tmp_file.flush()
        tmp_file_path = tmp_file.name
        yield tmp_file_path
```

frontend/src/app/auth/saml/acs/route.ts
```diff
@@ -0,0 +1,53 @@
...
+  const backendUrl = new URL(buildUrl("/auth/saml/acs"))
+  const backendFormData = new FormData()
+  backendFormData.append('SAMLResponse', samlResponse)
+
+  // Forward the request to the FastAPI backend
+  const backendResponse = await fetch(backendUrl.toString(), {
+    method: 'POST',
+    body: backendFormData,
+  })
+  console.log("Backend response:", backendResponse)
```
daryllimyt:
```suggestion
```
Might have sensitive data?

daryllimyt (changes requested):
Few things to go over but amazing work done here

tracecat/auth/users.py
topher-lo: Yeah good point it does. How can I check if an admin user already exists?

topher-lo: This should only run when there are no users:
```tracecat/api/setup_defaults.py
async def setup_defaults(session: AsyncSession, admin_role: Role):
    users = await list_users(session=session)
    if len(users) == 0:
        # Create admin user only if there are no users
        await get_or_create_default_admin_user()
```

PR #452, opened 2024-10-23
feat(integration)!: Rewrite LDAP and email functions

topher-lo:
## What changed
- Moved all LDAP and SMTP configs to secrets manager or action input
- Refactor logic in email action
  - Dropped premature class abstraction
  - Use smtp server context manager (more idiomatic)
  - Clean up repeated code in conditionals
- Renamed `disable_ad_user` and `enable_ad_user` to `disable_activate_directory_user` and `enable_active_directory_user`
- Moved `LDAP_USE_SSL` and `LDAP_TYPE` to action inputs as `use_ssl` and `is_active_directory`
- !!! Exceptions are no longer capture then return as dict

## NOTE
Users previously expecting exceptions to NOT fail the action will need to wait for our continue run on exception feature to be released in version 0.13.0 (1st week November).

Usage:
- Under the control flow tab, toggle "Continue on exception raised" to True.
- `${{ ACTIONS.some_failed_action.error }}`
- `.error` returns a dictionary with `{"error": "ValueError", "msg": str(err)}`


topher-lo:
@mattdurant Would be great to get your thoughts on this! The core automation platform / registry is now stable, so we're working on cleaning up existing integrations, building out new ones, and adding independent control flow logic (e.g. continue on exceptions) in the coming weeks.

mattdurant:
> @mattdurant Would be great to get your thoughts on this! The core automation platform / registry is now stable, so we're working on cleaning up existing integrations, building out new ones, and adding independent control flow logic (e.g. continue on exceptions) in the coming weeks.

All for this, this was modified from the existing Resend integration that was in place around 0.10 when I found the project, so the style is outdated compared to other integrations.

daryllimyt (approved):
🚢

PR #456, opened 2024-10-23
feat: Add configurable action retries

daryllimyt:
# Changes
- Add `ActionRetryPolicy` to configure maximum attempts (default: 1) and timeout (default: 300s)
- By default, the action will fail immediately unless max attempts is > 1
- If it's 0, will retry indefinitely (temporal semantics)
- Clean up some error handling logic in the run action activity as we transform the error in the registry client

# Testing
- Manual testing with http action failing to hit endpoint

# Screens

Screenshot 2024-10-23 at 01 46 28
Screenshot 2024-10-23 at 01 46 38

### DIFF
--- a/frontend/src/client/schemas.gen.ts
+++ b/frontend/src/client/schemas.gen.ts
@@ -122,10 +122,10 @@ export const $ActionResponse = {
 
 export const $ActionRetryPolicy = {
     properties: {
-        maximum_attempts: {
+        max_attempts: {
             type: 'integer',
-            title: 'Maximum Attempts',
-            description: 'Number of attempts if the action fails.',
+            title: 'Max Attempts',
+            description: 'Total number of execution attempts. 0 means unlimited, 1 means no retries.',
             default: 1
         },
         timeout: {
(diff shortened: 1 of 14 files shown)


topher-lo:
~@daryllimyt can we rename it to `max_retries`. `maximum_attempts` is a bit too verbose and less common imo (Datadog, Retool, prefect use `max retries`)? Does Temporal call it `maximum_attempts`?~

Understood that `maximum_attempts` is Temporal nomenclature also it's used by boto3. I guess `max_retries` is more a Python thing? Anyway, still nice to make it shorter `max_attempts`

frontend/src/client/schemas.gen.ts
```diff
@@ -117,6 +120,25 @@ export const $ActionResponse = {
     title: 'ActionResponse'
 } as const;
 
+export const $ActionRetryPolicy = {
+    properties: {
+        maximum_attempts: {
+            type: 'integer',
+            title: 'Maximum Attempts',
```
topher-lo, line 127:
```suggestion
            title: 'Max Retries',
```

frontend/src/client/schemas.gen.ts
```diff
@@ -117,6 +120,25 @@ export const $ActionResponse = {
...
+            description: 'Number of attempts if the action fails.',
```
topher-lo, line 128:
```suggestion
            description: 'Number of retries if the action fails.',
```

topher-lo (approved):

PR #476, opened 2024-11-02
feat(integrations)!: Drop jwt and oauth support in http action

topher-lo:
## Breaking Changes
- http jwt and oauth params have been removed
- Need to use Wiz and Checkpoint `get_auth_token` UDFs for authentication
- Future tools that require jwt and oauth auth, users need to open an issue / PR to implement `get_auth_token` UDF for that specific tool.

## Integrations Contract
- Maintainers manage and expose the auth logic (e.g. `get_jwt_tokens`) for tooling
- Users use Action Templates to call the token getters and call core endpoints for the tool

## Benefit
- Users don't have to worry about authentication logic
- Maintainers don't have to worry about implementing an app per API endpoint (we just need to focus on auth where Python is much more flexible to debug and maintain)


registry/tracecat_registry/integrations/limacharlie.py
```diff
@@ -0,0 +1,47 @@
+"""Limacharlie authentication.
+
+Docs: https://docs.limacharlie.io/apidocs/introduction
+"""
+
+import httpx
+
+from tracecat_registry import RegistrySecret, registry, secrets
+
+LIMACHARLIE_AUTH_URL = "https://jwt.limacharlie.io"
```
daryllimyt: can we make this consistent - either all base API URLs defined as a global var like this, or defined inside the udf

daryllimyt (changes requested):
small nit, feel free to ignore

daryllimyt (approved):

PR #500, opened 2024-11-09
feat(integration): Support custom HTTP certs

topher-lo:
## Why
Tracecat currently does not support self-signed certificates for http calls

## What changed
- Added custom SSL cert support to the httpx `AsyncClient` call: https://www.python-httpx.org/advanced/ssl/ via secrets manager

## Usage
- Add secret named `ssl` with keys `SSL_CLIENT_CERT`, `SSL_CLIENT_KEY` (optional), `SSL_CLIENT_PASSWORD` (optional).

Note: if you need multiple self-signed certs or do not want every the default environment to use your self-signed client-side cert for http calls, you'll need to run the workflow under a specific secret environment.


registry/tracecat_registry/base/core/http.py
```diff
@@ -2,16 +2,37 @@
...
 from pydantic import Field, UrlConstraints
 
-from tracecat_registry import logger, registry
+from tracecat_registry import RegistrySecret, logger, registry, secrets
 
 RequestMethods = Literal["GET", "POST", "PUT", "DELETE"]
 JSONObjectOrArray = dict[str, Any] | list[Any]
 
+http_secret = RegistrySecret(
+    name="http",
```
daryllimyt: perhaps `ssl` is a more suitable name? given all the keys are `SSL_*`. It's fine if the secret name != the function/namespace name imo

registry/tracecat_registry/base/core/http.py
```diff
@@ -37,19 +58,23 @@ async def http_request(
...
         Field(description="URL query parameters"),
-    ] = None,
+    ]
+    | None = None,
     form_data: Annotated[
         dict[str, Any],
         Field(description="HTTP form encoded data"),
-    ] = None,
+    ]
+    | None = None,
```
daryllimyt: any reason these are changed? These affect how the schema is shown in the UI

registry/tracecat_registry/base/core/http.py
```diff
@@ -72,7 +97,33 @@ async def http_request(
...
+                    key_file.flush()
+                    key_file_path = key_file.name
+
+            cert = [
+                cert_file_path,
+                key_file_path,
+                secrets.get("SSL_CLIENT_PASSWORD"),
+            ]
+            # Drop None values
+            cert = tuple([c for c in cert if c is not None])
```
daryllimyt: unnecessary listcomp
```suggestion
            cert = tuple(c for c in cert if c is not None)
```

daryllimyt (changes requested):

registry/tracecat_registry/base/core/http.py
```diff
@@ -23,6 +44,8 @@ class HTTPResponse(TypedDict):
     namespace="core",
     description="Perform a HTTP request to a given URL.",
     default_title="HTTP Request",
+    display_group="Core",
```
daryllimyt:
```suggestion
```

daryllimyt (approved):
:shipit:

PR #501, opened 2024-11-09
fix(integrations): Incorrect `secrets` API usage in get token actions

topher-lo:
- Drop `await`
- Get secret.get("SECRET_KEY") not secret.get("secret_name")


daryllimyt (approved):
🚢

PR #506, opened 2024-11-10
fix: Correct usage of validate dsl

daryllimyt:
# Changes
- Proper usage of `tracecat.validation.service.validate_dsl`


topher-lo (approved):
Looks good to me?

PR #508, opened 2024-11-11
feat+fix: Optional secrets

daryllimyt:
# Changes
- Fix issue with http request node requiring that `ssl` secret must be defined, when it should be optional
- To address this, implement optional secrets that skip assertion for optional `RegistrySecret`s, if not defined in the SM.

## Summary by CodeRabbit

- **New Features**
	- Introduced support for optional secrets in the `AuthSandbox` class, allowing for more flexible secret management.
	- Added a new `optional` property to the `$RegistrySecret` schema, enhancing the flexibility of secret definitions.
	- Implemented a new Docker Compose configuration for production deployment, facilitating the setup of multiple services.

- **Bug Fixes**
	- Improved error handling for missing secrets, providing clearer feedback on which secrets are not found.

- **Tests**
	- Enhanced testing coverage for the `AuthSandbox` class with new tests for handling required and optional secrets.

### DIFF
--- a/registry/tracecat_registry/base/core/http.py
+++ b/registry/tracecat_registry/base/core/http.py
@@ -16,6 +16,7 @@ JSONObjectOrArray = dict[str, Any] | list[Any]
 ssl_secret = RegistrySecret(
     name="ssl",
     optional_keys=["SSL_CLIENT_CERT", "SSL_CLIENT_KEY", "SSL_CLIENT_PASSWORD"],
+    optional=True,
 )
 """HTTP SSL certificate secrets.
 
--- a/docker-compose.prod.override.yml
+++ b/docker-compose.prod.override.yml
@@ -1,0 +1,22 @@
+services:
+  api:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    environment:
+      TRACECAT__APP_ENV: production
+
+  worker:
+    build:
+      context: .
+      dockerfile: Dockerfile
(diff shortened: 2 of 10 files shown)


registry/tracecat_registry/base/core/http.py
```diff
@@ -16,6 +16,7 @@
 ssl_secret = RegistrySecret(
     name="ssl",
     optional_keys=["SSL_CLIENT_CERT", "SSL_CLIENT_KEY", "SSL_CLIENT_PASSWORD"],
```
topher-lo, line 18:
```suggestion
    keys=["SSL_CLIENT_CERT"],
    optional_keys=["SSL_CLIENT_KEY", "SSL_CLIENT_PASSWORD"],
```

Has this combo been tested (i.e. `optional=True` with non null keys and optional keys?)

topher-lo (changes requested):

registry/tracecat_registry/base/core/http.py
daryllimyt: added more tests, and see:
https://github.com/TracecatHQ/tracecat/pull/508/files#diff-c466a7ce56c24160eb03320ad4f34f491c4482b72aa6d44335ad3d1bcc85f29eR74-R92

^ at commit time if a secret is optional but it's defined, we check that all required keys are defined, otherwise return validation errors

PR #514, opened 2024-11-12
fix(integration): Add sentinel value for LDAP connection attribute

mattdurant:
## Description

With the refactoring to registry and removal of async, accessing _connection before it's instantiated throws an exception.  This ensures that it exists when the check is performed before binding.

Didn't open an issue, minor fix.

## Steps to QA

Using any of the functions provided by this integration would throw exceptions without this fix.


daryllimyt (approved):
LGTM - we can ignore the failing pytest run, appears to be flaky. Thanks @mattdurant!

PR #534, opened 2024-11-20
feat(integration): Velociraptor integration

zaneGittins:
## Description

This integration runs a vql query against a Velociraptor server using a base64 encoded config as the secret. The configuration needs to be base64 encoded before adding it as a secret in Tracecat to preserve formatting.
You can use the following command to do so: `cat api.config.yaml | base64 -w` 

## Secrets
This action requires the following secrets for authentication:

* `velociraptor_ssl`

### DIFF
--- a/registry/tracecat_registry/integrations/velociraptor.py
+++ b/registry/tracecat_registry/integrations/velociraptor.py
@@ -1,0 +1,102 @@
+"""Generic interface to PyVelociraptor class.
+Author: Zane Gittins
+"""
+
+import json
+import grpc
+import yaml
+import base64
+
+from typing import Annotated, Any
+from pydantic import Field
+from pyvelociraptor import api_pb2
+from pyvelociraptor import api_pb2_grpc
+from tracecat_registry import RegistrySecret, registry, secrets
+
+velociraptor_secret = RegistrySecret(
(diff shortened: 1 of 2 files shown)


registry/tracecat_registry/integrations/velociraptor.py
```diff
@@ -0,0 +1,102 @@
...
+"""
+
+@registry.register(
+    default_title="Execute Velociraptor query",
+    description="Run VQL queries against a Velociraptor server via the gRPC api.",
+    display_group="Velociraptor",
+    namespace="integrations.velociraptor",
+    secrets=[velociraptor_secret],
+)
+
```
topher-lo:
```suggestion
```

registry/tracecat_registry/integrations/velociraptor.py
```diff
@@ -0,0 +1,102 @@
...
+async def velociraptor_query(
```
topher-lo:
```suggestion
async def run_velociraptor_query(
```

topher-lo (changes requested):
Small function name change

topher-lo (approved):

topher-lo (approved):

PR #535, opened 2024-11-21
feat(integrations): Add new string functions

topher-lo:


tracecat/expressions/functions.py
```diff
@@ -98,6 +98,46 @@ def format_string(template: str, *values: Any) -> str:
     return template.format(*values)
 
 
+def capitalize(x: str) -> str:
+    """Capitalize a string."""
+    return x.capitalize()
+
+
+def titlize(x: str) -> str:
```
daryllimyt: nit: either [`titleize`](https://www.google.com/search?sca_esv=7025c20cef8c26a5&q=titleize&spell=1&sa=X&ved=2ahUKEwi8sfbM7u6JAxVCFjQIHZNwDeYQBSgAegQIFhAB&biw=1554&bih=1019&dpr=2) (google autocorrects titlize to this) or `to_title` or just `title` (like how `capitalize` is just capitalize).

Claude:
> Both "titleize" and "titlize" exist as terms, but "titleize" is more common and generally preferred. It means to convert text to title case (capitalizing major words). The shorter form "titlize" is less frequently used and could be ambiguous.
> If you're working with code, many programming languages and libraries use "titleize" or "title_case" in their method names. For example, Ruby uses titleize, while Python uses title().

daryllimyt:
```suggestion
def titleize(x: str) -> str:
```

tracecat/expressions/functions.py
```diff
@@ -496,8 +531,15 @@ def eval_jsonpath(
...
 _FUNCTION_MAPPING = {
     # String transforms
+    "capitalize": capitalize,
+    "endswith": endswith,
+    "lowercase": lowercase,
     "slice": slice_str,
+    "split": split,
+    "startswith": startswith,
     "strip": strip,
+    "titlize": titlize,
```
daryllimyt:
```suggestion
    "titleize": titleize,
```

daryllimyt (approved):

PR #540, opened 2024-11-22
refactor(engine): Reorganize functions

topher-lo:
Functions are not grouped by usage category. This reorganizes the functions and adds inline comments for each section.


daryllimyt (approved):
LGTM, land when CI is done

PR #548, opened 2024-11-25
ci: Add auto release on tagged image push completion [DRAFT]

daryllimyt:
crude WIP, open to suggestions. Kinda annoying having to manually release when we create a tag. Also releasing before tagging is also not ideal since some users do try to install during this window. This is generated by cursor (claude) and should only run on tag creation


topher-lo:
Will revisit. Build times much faster now

PR #549, opened 2024-11-25
feat(engine): Replace jsonpath with lambda in intersect function

topher-lo:
Why? Lambda is easier to use.

### DIFF
--- a/tracecat/expressions/functions.py
+++ b/tracecat/expressions/functions.py
@@ -662,15 +662,16 @@ def or_(a: bool, b: bool) -> bool:
 
 
 def intersect[T: Any](
-    items: Sequence[T], collection: Sequence[T], jsonpath: str | None = None
+    items: Sequence[T], collection: Sequence[T], python_lambda: str | None = None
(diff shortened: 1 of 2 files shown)


tracecat/expressions/functions.py
```diff
@@ -662,15 +662,16 @@ def or_(a: bool, b: bool) -> bool:
...
+    """Return the set intersection of two sequences as a list. If a Python lambda is provided, it will be applied to each item before checking for intersection."""
     col_set = set(collection)
-    if jsonpath:
-        return list(
-            {item for item in items if eval_jsonpath(jsonpath, item) in col_set}
-        )
-    return list({item for item in items if item in collection})
+    if python_lambda:
+        fn = _build_safe_lambda(python_lambda)
+        result = [item for item in items if fn(item) in col_set]
```
daryllimyt: This should be a set comprehension
```suggestion
        result = {item for item in items if fn(item) in col_set}
```

daryllimyt (approved):

PR #587, opened 2024-12-04
fix(infra): Align network configuration for api and ollama in docker-compose

minhmannh2001:
- Updated ollama service to use the same network as api service.

## Description
This PR fixes a bug where the api service could not connect to the ollama service, causing workflows using the ai_action step with the llama3.2:1b model to fail. The issue was due to the ollama service not being part of the same network as the api service. By aligning the network configuration in the docker-compose file, requests from the api service to the ollama service will now succeed.

## Screenshots/Recordings

Sample workflow
image

Error logs
image

## Steps to QA

1. Follow the deploy tutorial, using the latest docker-compose.yaml file
2. Run a workflow with an ai_action step using the llama3.2:1b model. You can recreate the workflow shown in the above image.
3. Confirm that the workflow completes successfully and the ai_action step does not fail.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -173,6 +173,8 @@ services:
     container_name: ollama
     tty: true
     restart: unless-stopped
+    networks:
+      - core
     image: ollama/ollama:${OLLAMA__VERSION}
     ports:
       - 11434:11434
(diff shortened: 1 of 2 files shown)


topher-lo (approved):
LGTM

PR #595, opened 2024-12-10
[ORPHAN] feat(engine): Add enum type to expectations

topher-lo:


topher-lo:
Messed up branch. Reopening in another PR

PR #637, opened 2024-12-20
feat(integration): SentinelOne list Threats + update Threat status

benacumen:
## Description

Integrations to list Threats and update Threat Status in SentinelOne.


topher-lo (approved):
🚢

PR #644, opened 2024-12-21
feat+fix(engine)!: Support enums and defaults in Action Templates

topher-lo:
## What changed?
- Fixes issue where arguments with defaults do not show up in the action run. Fixed by validated arguments (i.e. creating the expectation model) before running the template action.
- Updates `validate_args` method to return `model_schema_inputs` with json mode (i.e. enum instances are coerced into string)
- Drop unused `validate` conditional in `_run_action_direct` (this path isn't used anywhere in the code)
- Replaced all references to `tracecat_registry._internal.exceptions.TracecatValidationError` with `tracecat.types.exceptions.TracecatValidationError`

Note: we had to make `key` in `TracecatValidationError` optional and serializable in order to bypass Ray's issue reraising instances of exceptions. https://github.com/ray-project/ray/issues/8948. Unsure why but unless all __init__ variables for the exception are optional, ray seems to keep raising the `TypeError` with missing `key`. See reference for example.

Screenshot 2024-12-22 at 4 51 03 PM

## References
ray serialization error


tracecat/registry/actions/models.py
```diff
@@ -148,6 +148,8 @@ def validate_args[T](self, *args, **kwargs) -> T:
                 f"Unexpected error when validating input arguments for bound registry action {self.action!r}. {e}",
                 key=self.action,
             ) from e
+        else:
+            return validated_args
```
daryllimyt: no need try-else control flow, just return this directly within the `try` block, if an exception is raised it doesn't return anyways

tracecat/expressions/expectations.py
```diff
@@ -87,10 +96,42 @@ def reference_type(self, name) -> str:
...
+            value_lower = value.lower()
+            if value_lower in seen_values:
+                raise ValueError(f"Duplicate enum value: {value}")
+
+            seen_values.add(value_lower)
+            enum_values[value] = value
 
-def parse_type(type_string: str) -> Any:
+        # Convert to upper camel case (e.g., "user_status" -> "UserStatus")
+        enum_name = "".join(word.title() for word in self.field_name.split("_"))
```
daryllimyt, line 120: might want to introduce a prefix for the enum, in case it clashes with other symbols in the namespace since it's user defined

tracecat/executor/service.py
daryllimyt, line 1: you shouldn't have to make change to this module - validation is done outside [here](https://github.com/TracecatHQ/tracecat/blob/5c82674d038a2219c96f1f5929f3aa162dde661b/tracecat/executor/router.py#L76-L104) in its own temporal activity before

tracecat/expressions/expectations.py
```diff
@@ -25,16 +26,18 @@
 
 INTEGER: "int"
 STRING: "str"
+STRING_LITERAL: "\"" /[^"]*/ "\""
```
daryllimyt: optionally support single quotes as well? see expressions lark grammar  for an example

tracecat/executor/service.py
```diff
@@ -86,24 +86,23 @@ async def coro():
...
     """
-    if validate:
-        # Optional, as we already validate in the caller
-        args = action.validate_args(**args)
     if action.is_template:
-        # This should not be reached
+        # Defensive check
         raise ValueError("Templates cannot be executed directly")
+
+    validated_args = action.validate_args(**args)
```
daryllimyt, line 92: if an exception (RegistryValidationError) is raised, does it get shown nicely in the UI?

tracecat/executor/service.py
```diff
@@ -145,12 +144,15 @@ async def run_single_action(
...
-        return await _run_action_direct(action=action, args=args)
+        logger.info("Running template action async", action=action.name)
+        result = await run_template_action(action=action, args=args, context=context)
+    else:
+        logger.trace("Running UDF async", action=action.name)
+        flat_secrets = flatten_secrets(secrets)
+        with env_sandbox(flat_secrets):
+            result = await _run_action_direct(action=action, args=args)
+
+    return result
```
daryllimyt, line 155: unnecessary imo, I find returning earlier much easier to follow

tracecat/executor/service.py
```diff
@@ -86,24 +86,23 @@ async def coro():
...
     """Execute the UDF directly.
 
     At this point, the UDF cannot be a template.
     """
-    if validate:
-        # Optional, as we already validate in the caller
-        args = action.validate_args(**args)
     if action.is_template:
-        # This should not be reached
+        # Defensive check
```
daryllimyt: Can we some notion of `unreachable`? Defensive loses some of this meaning imo

tracecat/expressions/expectations.py
```diff
@@ -87,10 +96,42 @@ def reference_type(self, name) -> str:
...
+        logger.trace("Enum type:", name=enum_name, values=enum_values)
+        return Enum(enum_name, enum_values)
 
-def parse_type(type_string: str) -> Any:
+    @v_args(inline=True)
+    def STRING_LITERAL(self, value) -> str:
+        # Remove quotes from the value
+        value = value.strip('"').strip("'")
+        # Coerce to string
+        return str(value)
```
daryllimyt: is this not already a string if you can strip it?

daryllimyt (changes requested):

tracecat/executor/service.py
topher-lo: kk
```suggestion
        # This should not be reachable
```

tracecat/executor/service.py
topher-lo: Unsure. How can I test / check this?

tracecat/expressions/expectations.py
topher-lo: kk

tracecat/expressions/expectations.py
topher-lo: For example?

tracecat/expressions/expectations.py
topher-lo: Oh yeah.

tracecat/registry/actions/models.py
topher-lo: indeed

tracecat/executor/service.py
topher-lo: This is related to parsing to inject defaults not validation.

topher-lo: See discussion here: https://github.com/TracecatHQ/tracecat/pull/652

tracecat/expressions/expectations.py
topher-lo: kk

topher-lo: Added `Enum` as prefix to enum class name

tracecat/executor/service.py
topher-lo: Done.

tracecat/registry/actions/models.py
```diff
@@ -390,7 +391,7 @@ def from_validation_result(
         )
 
     @staticmethod
-    def from_dsl_validation_error(exc: TracecatValidationError):
+    def from_dsl_validation_error(exc: RegistryValidationError):
```
daryllimyt: validation likely needs some consolidation, but i don't think this needs to be changed - this is for build-time validation, we explicitly catch `TracecatValidationError` exception groups. `RegistryValidationError` doesn't inherit from this so the returned struct will be malformed. Unless i'm missing something here

daryllimyt (changes requested):
minor change

daryllimyt (approved):
LGTM

PR #649, opened 2024-12-21
feat(engine): Store backend and EE executor router

daryllimyt:


topher-lo (commented):
EE license looks fine. But we need to come to consensus about the "number of user seats" part. I think this is not necessary.

topher-lo:
Stale.

PR #652, opened 2024-12-21
feat(engine): Use parsed workflow trigger inputs

topher-lo:
Requires #644 for enum expectations.

## Why
- Previously the workflow expects Pydantic model was only used for validation not parsing / model coercion (see distinction in pydantic docs: https://docs.pydantic.dev/1.10/usage/models/ and screenshot below)
- This means defaults in expects weren't injected into the trigger inputs for workflow runs
- Related issue for template actions (#644) where pydantic model wasn't used to parse args

## What changed
- Add multiple parameterized unit tests for trigger inputs with expects (with defaults, without defaults, with enums)
- Added `payload` to `ValidationResult`
- Passed validated payload (aka parsed trigger inputs that conforms to expects pydantic model) into execution context.

## References
Screenshot 2024-12-21 at 1 07 15 PM

### DIFF
--- a/tests/unit/test_workflows.py
+++ b/tests/unit/test_workflows.py
@@ -39,6 +39,7 @@ from tracecat.logger import logger
 from tracecat.secrets.models import SecretCreate, SecretKeyValue
 from tracecat.secrets.service import SecretsService
 from tracecat.types.auth import Role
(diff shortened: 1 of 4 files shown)


topher-lo:
Not priority. Will revisit.

PR #678, opened 2024-12-28
feat(ui): Add workflow alias to workbench navbar

daryllimyt:
Show the current workflow alias in the navbar (and nothing if there's none) without having to view the workflow panel.

# Screens
See `workflows.root` in navbar breadcrumbs
image


daryllimyt:
@topher-lo Verified that the workflow alias resizes nicely without wrapping as it's inside the container

PR #681, opened 2024-12-30
fix(engine): Propagate executor error messages safely

daryllimyt:
# Changes
- Structure errors as `RegistryActionErrorInfo` and return them when returning from the ray executor
- Reraise these as `WrappedExecutionError` and handle appropriately at the route level

# Screens
Before
Screenshot 2024-12-30 at 01 50 50

After
Screenshot 2024-12-30 at 01 51 03


topher-lo (approved):
Elegant

PR #694, opened 2025-01-02
feat(engine): Parameterize git host for custom repository 

topher-lo:
## Why
- We currently hardcode github.com as a host for custom repositories
- We have users that have self-hosted gitlab in production (this is blocking for them)

## What changed
- Added an env var `TRACECAT__ALLOWED_GIT_DOMAINS=github.com,gitlab.com,bitbucket.org`
- Renamed `parse_github_url` to `parse_git_url` with new match patterns with multiple hosts.
- Return `host` (extra return value) as well from `parse_git_url`
- Dynamically add new hosts to SSH client given what's parsed when installing remote repository

## QA
Tested with GitLab deploy tokens
Screenshot 2025-01-02 at 9 42 23 PM

## Note
The Tracecat secret is still called `github-ssh-key` to prevent breaking changes. Will keep this for now and only revisit when organization settings are implemented.


tracecat/registry/repository.py
```diff
@@ -612,32 +623,45 @@ def safe_url(url: str) -> str:
...
+    Parse a Git repository URL to extract organization, package name, and branch.
+    Handles Git SSH URLs with 'git+ssh' prefix and optional '@' for branch specification.
 
     Args:
-        url (str): The GitHub URL to parse.
+        url (str): The repository URL to parse.
 
     Returns:
-        tuple[str, str, str]: A tuple containing (organization, package_name, branch).
+        tuple[str, str, str, str]: A tuple containing (host, organization, package_name, branch).
```
daryllimyt, line 643: returns `GitUrl`, but can fix later if urgent

daryllimyt (approved):
LGTM, one docs change but can fix later as it's not important

Issue #505, opened 2024-11-10
Centralized errors / workflow runs reporting


mattdurant:
**Is your feature request related to a problem? Please describe.**
Unhandled exceptions are not easily reported on currently, if you have a workflow that has errored out in a way that is unexpected, unless you add error handling to the workflow itself, there is no centralized place to report or alert on errors.

**Describe the solution you'd like**
A few ideas:

1. If a workflow ends because of an exception, have an alerting configuration/workflow that runs.  Idea here is that there is a system-defined "error" workflow that we could use to do whatever we need to with an error.

2. Logging and a report of same errors that can be viewed or scheduled.

3. Single "runs" view that shows all workflows, not just the currently open one.  View can be filtered on whether the run was successful or not.

**Describe alternatives you've considered**
The only way currently to tell if an error occurred is to open each workflow and look at the previous runs or inspect logs on the worker container for errors.

**Additional context**

topher-lo:
To clarify, in your mind, this centralized list of failures consolidates all errors across all workflows in a workspace correct?

I think this makes sense

topher-lo:
Haven't seen option 3 in SOAR UI. But it's the default UI/UX for data workflow orchestrators like prefect and airflow. And it makes...a lot of sense.

Especially cause Tracecat encourages smaller composable workflows.

A list of all workflow runs is the most elegant option. Also gives nice filtering / search capabilities.

mattdurant:
Yes, as an example, since we moved our production instance from somewhere around 0.9 to 0.13, some of our workflows were broken due to namespaces renamed or configuration for udfs changing.  There was no way to know this unless proactively unless each workflow was inspected to look at the recent runs.  Highlighted a bigger issue such as a secret expiring and causing an integration to stop working and not knowing.

mattdurant:
> Haven't seen option 3 in SOAR UI. But it's the default UI/UX for data workflow orchestrators like prefect and airflow. And it makes...a lot of sense.
> 
> Especially cause Tracecat encourages smaller composable workflows.
> 
> A list of all workflow runs is the most elegant option. Also gives nice filtering / search capabilities.

I think a big piece is going to be whether we can also have the system notify on recurring errors with a workflow.  I think we discussed something like this at one point, being able to see the "health" of a workflow.  Notifying on the health of a workflow declining would be helpful.

topher-lo:
[...]

PR #697, opened 2025-01-03
feat(engine): Implement error handler workflow functionality in DSLWorkflow

daryllimyt:
Progresses #505

# Changes
- Make workflows return some _sensible error info_. Current impl is a dict of action ref to `ActionErrorInfo`. This improves the final workflow execution event, and the error info shown to callers of child workflows.
- Add error handling functionality to `DSLWorkflow`. 
  - A `Workflow` object may specify a workflow ID or alias as an error handler
  - This must be a workflow that exists
  - This workflow gets run if an error occurs in the primary workflow, and receives error info

## Implementation details
  - We wrap the existing workflow run logic with exception handling that triggers a child workflow that is the error handler.
  - This needs to be embedded inside `DSLWorkflow` because of scheduled and pull-based workflows, where the DSL is not known until it's pulled from the DB after the workflow starts.
  - The error handler child wf receives the _sensible error info_ as trigger input, alongside some additional metadata from the failing workflow

  
  
# Screens
Improved workflow error info
image

Improved child workflow error info
image

Error handler workflow trigger inputs
image

  
  
  
# Testing
- Add integration test that creates error handler and wf failing wf
 - It then runs the failing wf and uses the `WorkflowExecutionsService` to verify that
   - The failing wf ran
   - The error handler was initialized with correct data
   - The error handler started and completed
- Tset for invalid handler that has no match

### DIFF
--- a/frontend/src/client/schemas.gen.ts
+++ b/frontend/src/client/schemas.gen.ts
@@ -912,15 +912,36 @@ export const $EventGroup = {
             title: 'Action Id'
         },
         action_ref: {
-            type: 'string',
+            anyOf: [
+                {
+                    type: 'string'
+                },
+                {
+                    type: 'null'
+                }
+            ],
             title: 'Action Ref'
         },
         action_title: {
-            type: 'string',
+            anyOf: [
+                {
+                    type: 'string'
+                },
+                {
+                    type: 'null'
+                }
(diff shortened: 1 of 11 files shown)


topher-lo (approved):
LGTM

PR #699, opened 2025-01-03
fix(security): Microsoft entra nameattribute parsing

topher-lo:
## What changed
Made SAML response parsing more robust by removing parsing AttributeTemplate (optional) and trying multiple response attributes to find the email value.

## Additional fixes
Found bug where `logout` endpoint is not included with saml.

## QA
https://github.com/user-attachments/assets/83436cc1-4573-4b2b-9c56-03f87008d41a

## Breaking
- Removed SAML certificate env vars (only rely on metadata URL now)
- Still need to look into redirect after login (but it's probably the IdP side issue)


daryllimyt:
re: `logout` endpoint, you'll have to do something like [this](https://github.com/TracecatHQ/tracecat/blob/9617b5d63c503c3e34e8071dfcfa326d8ffc1c0f/tracecat/auth/users.py#L237-L271) as done for OAuth

daryllimyt (approved):
LGTM 🚢

PR #701, opened 2025-01-07
chore(integration)!: Drop velociraptor integration

topher-lo:
## Why
The python pip package doesn't meet community guidelines.

## Note
We have plans to build our own gRPC to Tracecat UDF convertor to enable folks to maintain their own gRPC integrations without relying on third party packages.


zaneGittins:
Hey @topher-lo, pyvelociraptor has received some updates, now v0.1.11 and using grpcio==1.58.3, do you think we may be able to add this back?

Although it seems that these may conflict:
- pyvelociraptor==0.1.11 requires cryptography==45.0.4
- tracecat requires cryptography==44.0.1

PR #707, opened 2025-01-08
chore(ui): Delete unused elements in workflow trigger and alias

topher-lo:
- Drop status section and dropdown in workflow trigger
- Drop field description for alias to make consistent with other fields UI
- Added tooltip (same description also shows if no alias is added)
- Renamed "Error handler" to "Error workflow"

Screenshot 2025-01-08 at 8 00 00 AM
Screenshot 2025-01-08 at 7 59 55 AM
Screenshot 2025-01-08 at 7 59 49 AM

### DIFF
--- a/frontend/src/components/workbench/panel/workflow-panel.tsx
+++ b/frontend/src/components/workbench/panel/workflow-panel.tsx
@@ -32,7 +32,6 @@ import {
 import {
   Form,
   FormControl,
-  FormDescription,
   FormField,
   FormItem,
   FormLabel,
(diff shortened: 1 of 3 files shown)


frontend/src/components/workbench/panel/workflow-panel.tsx
```diff
@@ -276,10 +276,6 @@ export function WorkflowPanel({
                                 />
                               )}
                             </div>
-                            <FormDescription className="text-xs">
-                              A name that can be used to uniquely identify this
-                              workflow.
-                            </FormDescription>
```
daryllimyt, line 285: could we maybe have a tooltip that tells you what aliases are used for?

topher-lo: Can do

daryllimyt (approved):
🚢

PR #711, opened 2025-01-08
feat(ui): Add template tab and fix state actions on node click

topher-lo:
## Before
- Fixed state issues with error saving state and tabs selected persisting between nodes
- Fixed tab switcher content overflowing and the content (e.g. editor) remains the same size
Screenshot 2025-01-08 at 12 17 54 PM

## After
Screenshot 2025-01-08 at 1 11 05 PM

## Others
- I deleted the toast when an action has malformed YAML. The state is really janky and removes immersion i.e. too aggressive (gives me unnecessary cortisol).
- ^I think this is feature of the past when we didn't have the really nice SaveStateIcon

### DIFF
--- a/frontend/src/lib/hooks.tsx
+++ b/frontend/src/lib/hooks.tsx
@@ -144,10 +144,6 @@ export function useAction(
     },
     onError: (error) => {
       console.error("Failed to update action:", error)
-      toast({
-        title: "Failed to save action",
-        description: "Could not update your action. Please try again.",
-      })
       setIsSaving(false)
     },
   })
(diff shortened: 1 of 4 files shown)


frontend/src/lib/hooks.tsx
```diff
@@ -144,10 +144,6 @@ export function useAction(
     },
     onError: (error) => {
       console.error("Failed to update action:", error)
-      toast({
-        title: "Failed to save action",
-        description: "Could not update your action. Please try again.",
-      })
```
daryllimyt, line 150: any reason why this was removed? is it clear enough without?

daryllimyt (approved):
LGTM, one clarification

frontend/src/lib/hooks.tsx
topher-lo: The state is janky and distracting. I think it's clear enough without because it'll get caught on "save".

PR #729, opened 2025-01-10
security: Drop stack trace from failure message

topher-lo:
## What changed
- Drop stack trace in failure handler
- Only show filename, function, and lineno in ExecutorActionErrorInfo if app env is "development"
- Also I think it's cleaner too from a UX perspective

Screenshot 2025-01-10 at 12 19 37 PM


daryllimyt (approved):
🚢

PR #758, opened 2025-01-15
feat(engine): Basic implementation for pull-based executor and remove internal executor endpoints

daryllimyt:
# Description
- Use Ray's RuntimeEnv to manage custom udf repository, instead of having to push to executor to sync updates.
- Instead, when a user syncs the repo, we still pull the repo and update the action interfaces for the builder and registry, but we don't allow pushing to executor.
- We update the executor so that it checks the db repo for a SHA.
- For stability, we enforce that we pass a SHA with the git url into the ray RuntimeEnv. If no SHA is available, we don't install the remote repo as a dep.
  - Ray caches the dependencies depending on the hash of the RuntimeEnv. If a change is observed, it will automatically `pip install` the updated version. For this reason, we always pass a SHA into the git url.

# Changes
- Remove sync executors and validation from executor

### DIFF
--- a/frontend/src/client/services.gen.ts
+++ b/frontend/src/client/services.gen.ts
@@ -78,8 +78,6 @@ import type {
   RegistryRepositoriesGetRegistryRepositoryResponse,
   RegistryRepositoriesListRegistryRepositoriesResponse,
   RegistryRepositoriesReloadRegistryRepositoriesResponse,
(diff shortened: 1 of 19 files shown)


topher-lo:
Note that we can further tighten our Docker subnets and AWS security groups after this change.

PR #778, opened 2025-01-21
feat(integrations): Support datetime or string in datetime functions

topher-lo:
## Rationale
We JSON serialize all inputs post-Pydantic validation before sending it over to the executor. This means datetimes are passed into each step in an action template as a JSON supported dtype (string, integer, float, object) only. There are no native JSON datetime objects.

## What changed
- All datetime functions e.g. `format_datetime` run `to_datetime` (which converts an ISO 8601 datetime string into a datetime object) if the input is a string.
- Added more tests to check that important datetime functions e.g. `to_datetime` and `format_datetime` correct support both datetime and ISO string with and without timezone data.
- Added `parse_datetime` function, which supports python `strftime` formatting
- Dropped `to_timestamp_str` (the functionality and naming is strange, likely a hallucination, and I don't think anybody is using it)

### DIFF
--- a/tests/unit/test_functions.py
+++ b/tests/unit/test_functions.py
@@ -942,10 +942,26 @@ def test_difference(a: Sequence[Any], b: Sequence[Any], expected: list[Any]) -> 
 @pytest.mark.parametrize(
     "input_val,unit,expected",
     [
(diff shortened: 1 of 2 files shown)


daryllimyt (approved):
LGTM, maybe add testcases with "Z" at the end?

Issue #779, opened 2025-01-22
Login form doesn't show up


szymon-romanko:
**Describe the bug**
When following installation instructions from wiki for [docker compose](https://docs.tracecat.com/self-hosting/deployment-options/docker-compose), app starts successfully but login form does not display properly

Log shows error message: `Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='# Enables SAML on startup', input_type=str]`
Which is a result of this [commit](https://github.com/TracecatHQ/tracecat/commit/ea638054eb377cbbddbd42a60ad0f82b6836b664#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8cR83)

Setting this line in `.env` to `TRACECAT__SETTING_OVERRIDE_SAML_ENABLED=false  # Enables SAML on startup` fixes the issue

**To Reproduce**
1. Run commands from wiki
```shell
curl -o env.sh https://raw.githubusercontent.com/TracecatHQ/tracecat/0.21.6/env.sh
curl -o .env.example https://raw.githubusercontent.com/TracecatHQ/tracecat/0.21.6/.env.example
chmod +x env.sh && ./env.sh
curl -o Caddyfile https://raw.githubusercontent.com/TracecatHQ/tracecat/0.21.6/Caddyfile
curl -o docker-compose.yml https://raw.githubusercontent.com/TracecatHQ/tracecat/0.21.6/docker-compose.yml
docker compose up
``` 
2. Go to web browser and try to log into tracecat
3. After clicking "Sign In" there is no login form (see screenshots)

**Expected behavior**
There should be a way for user to log in or sign up 

**Screenshots**
Login form doesn't show up
![Image](https://github.com/user-attachments/assets/b6e417fb-8557-4417-b2d1-4875b3fc3dd3)

**Environment (please complete the following information):**

- OS: Ubuntu 24.04.1 LTS
- CPU architecture: x86
- Docker version 27.3.1
- Docker compose 2.29.7
- Are you using Docker desktop? - No

**Additional context**
<details><summary>Full error message</summary>
<p>

```
api                   | INFO:     172.21.0.3:47104 - "GET /users/me HTTP/1.1" 401 Unauthorized
api                   | 2025-01-22 09:00:20.089770Z [16] | ERROR     tracecat.api.common:generic_exception_handler:13 - Unexpected error | {'exc': 1 validation error for AppInfo
api                   | saml_enabled
[...]

PR #780, opened 2025-01-22
build(fix): Syntax error with comment and setting override

szymon-romanko:
## Description

Fixes bug https://github.com/TracecatHQ/tracecat/issues/779

## Steps to QA

See *To Reproduce* section in the issue


topher-lo:
Thanks for opening this up!

topher-lo (approved):

PR #782, opened 2025-01-23
fix(api): Show public (nextjs) SAML SP ACS endpoint and hardcode API ACS endpoint

topher-lo:
## What changed
- We previously showed the /api/auth/saml/acs endpoint in the settings page. This is not right. The user should set the /auth/saml/acs endpoint in their IdP instead
- Having an environment variable called `SAML_SP_ACS_URL` is super misleading (it previously referred to the API ACS endpoint, but it isn't even configurable....)

## Possibly fixes
- Warning that `/api/auth/saml/acs` endpoint is not in path. This kind of makes sense since we set the /api endpoint in the saml configs NOT the public SAML SP ACS endpoint (without the /api)

## QA
Tested Okta. Tested Entra (still doesnt redirect but transcendence 🌠 )


daryllimyt (approved):
LGTM

PR #783, opened 2025-01-23
docs(integrations): Clearer descriptions for `payload` and `form_data`

topher-lo:
I also get confused sometimes between `params`, `payload`, and `form_data` and their associated methods. I don't think changing `payload` to `body` is a good idea, since technically `form_data` is also inside the request body. Anyway i think these descriptions clarify things. Does for me.

Prior art: I looked at curl / httpx's official descriptions of `data` / `json` arguments.


daryllimyt (approved):
LGTM

PR #807, opened 2025-01-28
fix(ui): Scrolling making workflow builder minify

topher-lo:
## Fix Command Menu Scroll Position During Search

This PR fixes an issue where the command menu's scroll position would jump unexpectedly when deleting characters in the search input. The scroll position should remain at the top regardless of whether the user is typing new characters or deleting them.

### Problem
When deleting characters in the search input, more items become visible in the filtered list. The browser tries to maintain the relative scroll position of currently visible elements, causing an undesirable jumping effect.

### Solution
We now force the scroll position to stay at the top after any input value changes by:
1. Updating the input value
2. Using `requestAnimationFrame` to wait for DOM updates
3. Explicitly setting the scroll position to 0 on the command list element

### Technical Details
- Using `requestAnimationFrame` ensures our scroll reset happens after React's re-render but before the next paint
- We target the `[cmdk-list]` element directly as it's the actual scrollable container
- The timing of the scroll reset prevents any visible jumping

### Testing
- [x] Typing characters keeps scroll at top
- [x] Deleting characters keeps scroll at top
- [x] No visible jumping when filtering results
- [x] Smooth user experience during search

### DIFF
--- a/frontend/src/components/workbench/canvas/selector-node.tsx
+++ b/frontend/src/components/workbench/canvas/selector-node.tsx
@@ -114,7 +114,17 @@ export default React.memo(function SelectorNode({
           ref={inputRef}
           className="!py-0 text-xs"
           placeholder="Start typing to search for an action..."
-          onValueChange={(value) => setInputValue(value)}
+          onValueChange={(value) => {
+            // First update the value
+            setInputValue(value)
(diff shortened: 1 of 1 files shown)


daryllimyt (approved):
LGTM

PR #882, opened 2025-02-21
feat(engine): Implement `core.table.lookup` action

daryllimyt:
# Description
- Implement table lookup from workflows. We expose a `core.table.lookup` interface that's handled by native `DSLWorkflow` logic, as opposed to a udf implementation. See `core.workflow.execute` for reference.
- Update action node icon

# Testing
- Added integration tests for calling in workflows


topher-lo (approved):
:shipit:

PR #912, opened 2025-02-26
chore(infra): Standardized restart policy

szymon-romanko:
## Description
In `docker-compose.yml` different services had different `restart_policy` set
This is small fix that standardizes them all to `unless-stopped`

**Before:**
```caddy - unless-stopped  
api - unless-stopped  
worker - unless-stopped  
executor - unless-stopped  
ui - unless-stopped  
postgres_db - always  
temporal_postgres_db - always  
temporal - always  
temporal_ui - not specified, defaults to no
```

## Steps to QA

Follow default installation steps

1. `git clone https://github.com/TracecatHQ/tracecat.git`
2. `./env.sh`
3. `docker compose up -d`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -156,7 +156,7 @@ services:
   postgres_db:
     image: postgres:16
     container_name: postgres_db
-    restart: always
+    restart: unless-stopped
     networks:
       - core-db
     shm_size: 128mb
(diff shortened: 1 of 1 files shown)


topher-lo (approved):
LGTM!

PR #915, opened 2025-02-27
chore: Fix up llm type hints

daryllimyt:


topher-lo:
Will do in another PR

PR #923, opened 2025-03-04
feat(core): Add lookup multiple rows action

daryllimyt:
# API Design
- Consider also `lookup_many`


topher-lo:
I think lookup_many aligns better with convention.

daryllimyt:
Should we rename "Insert Row" to "Insert Record" as well? And what's the preference for `default_title`--capitalize first in sentence or titlecase?

EDIT: I've made the above changes too

topher-lo (approved):

PR #925, opened 2025-03-05
feat(integrations)!: Break LLM calls into individual providers

topher-lo:
## Breaking
- Dropped `core.ai_action` to reduce cost of maintaining routing logic
- We will focus on using models directly to achieve goals (agentic use-cases) best suited to specific models

## TODO in future PR
- We do NOT support structured outputs in the calls (yet)
- For structured outputs, we need a way to convert YAML -> dict[str, Any] to Pydantic BaseModel (we already do this with `expects`. We should reuse that functionality and syntax).
- Eventually would it be possible to have a context called `ACTIONS.http_request.expects` to quickly pass in the `expects` for some action?
- Add `gemini` and `anthropic` (via bedrock) providers

## Summary

1. Added `llm.openai.call` and `llm.ollama.call` actions that are simple wrappers around `async_xyz_call`.
2. Reorganized the LLM code into a dedicated sublib
3. Updating dependencies to newer versions
4. Improving model handling (replaced json mode with structured outputs) for both OpenAI and Ollama
5. Aded newer models (4o, o1, o3-mini, mistral, llama3.3) to enums

### Code Organization
- Moved LLM functionality from a single file to a dedicated module structure:
  - Created `tracecat/llm/__init__.py` for exports
  - Created `tracecat/llm/ollama.py` for Ollama-specific functionality
  - Removed the old monolithic LLM implementation

### LLM Provider Improvements
- **Ollama**:
  - Added more specific model size tags to avoid ambiguity
  - Improved model validation and local model checking
  - Enhanced error handling for model availability
  - Added functions to list and check local models

- **OpenAI**:
  - Updated to support newer API patterns (structured outputs, new chat response types)

### Testing
- Added comprehensive smoke tests in `tests/unit/test_llms.py`:
  - Tests for basic prompts
  - Tests for system prompts
  - Tests for conversation memory
  - Added "LLM-as-judge" verification pattern

### Dependency Updates
- Updated several dependencies:
  - `openai` from 1.30.3 to 1.65.2
  - `ollama` from 0.3.3 to 0.4.7
  - `OLLAMA__VERSION` from 0.3.12 to 0.5.12
  - Fixed version constraints for several other packages

## Related Issues

This PR appears to be part of ongoing improvements to the LLM functionality in Tracecat. While no specific issues are directly referenced, it addresses the need for better organization and testing of LLM capabilities.

## Steps to QA

Run the new LLM smoke tests to verify functionality with both providers:
 ```bash
 pytest tests/unit/test_llms.py -v
 ```


topher-lo:
> * Eventually would it be possible to have a context called `ACTIONS.http_request.expects` to quickly pass in the `expects` for some action?

@daryllimyt thoughts?

tests/unit/test_llms.py
```diff
@@ -0,0 +1,166 @@
...
+async def test_user_prompt(call_llm_params: tuple[str, Callable]):
+    provider, call_llm = call_llm_params
+    prompt = "What is the capital of France?"
+    kwargs = {"prompt": prompt}
+    kwargs = {
+        **kwargs,
+        **load_api_kwargs(provider),
+    }
+    response = await call_llm(**kwargs)
+    assert response is not None
```
daryllimyt: maybe 
```python
assert "paris" in response.lower()
```

tests/unit/test_llms.py
```diff
@@ -0,0 +1,166 @@
...
+    response = await call_llm(**kwargs)
+
+    # TODO: Create a simple router for content extraction
+    # Corner case dealing with len(response.choices) > 0 for OpenAI
+    expected_response_type = ExpectedChatResponseType[provider.upper()]
+    match expected_response_type:
+        case ExpectedChatResponseType.OPENAI:
+            response_content = response.choices[0].message.content
+        case ExpectedChatResponseType.OLLAMA:
+            response_content = response.message.content
```
daryllimyt: Instead of matching the actual type to the class, you can just match the response object to the type directly:
```python
match response:
    case ParsedChatCompletion(): # Note that this is an instance/invocation of the class
         # `response` is now type narrowed as `ParsedChatCompletion`
    case ChatResponse():
        ...
 ```

File: tracecat/llm/openai.py [hunk 3]
```diff
@@ -0,0 +1,104 @@
...
+        ]
+    elif memory:
+        messages: list[ChatCompletionMessageParam] = [
+            *memory,
+            {"role": "user", "content": prompt},
+        ]
+    else:
+        messages: list[ChatCompletionMessageParam] = [
+            {"role": "user", "content": prompt}
+        ]
```
daryllimyt: Could be more concise
```suggestion
    messages: list[ChatCompletionMessageParam] = []
    
    if system_prompt:
        messages.append({"role": "developer", "content": system_prompt})
    elif memory:
        messages.extend(memory)
    
    messages.append({"role": "user", "content": prompt})
```

daryllimyt: Also, this implementation seems like the system prompt and memory are mutually exclusive? So you can't use both in the same call - intentional?

tracecat/llm/openai.py
```diff
@@ -0,0 +1,104 @@
...
+    O1 = "o1"
+    # https://platform.openai.com/docs/models/o1-mini
+    O1_MINI = "o1-mini"
+    # https://platform.openai.com/docs/models/o3-mini
+    O3_MINI = "o3-mini"
+
+
+# Must support structured outputs
+# https://platform.openai.com/docs/guides/structured-outputs#supported-models
+DEFAULT_OPENAI_MODEL = OpenAIModel.GPT4O_MINI
```
daryllimyt: Could make these actual docstrings visible in the IDE like:
```
x: int
"""Docstring"""
```

tracecat/llm/ollama.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+async def check_model_valid(model: str, api_url: str | None = None) -> bool:
+    """Check if a model is valid."""
+    client = _get_ollama_client(api_url)
+    try:
+        await client.pull(model)
+    except Exception:
+        return False
+    return True
```
daryllimyt: is this used?

tracecat/llm/ollama.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+async def list_local_models(api_url: str | None = None) -> list[dict[str, Any]]:
+    """List all models available locally."""
+    client = _get_ollama_client(api_url)
+    models = await client.list()
+    return [model.model_dump() for model in models.models]
```
daryllimyt, line 61: Would DX be better downstream if we kept this as pydantic models?

tracecat/llm/ollama.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+async def list_local_model_names(api_url: str | None = None) -> list[str | None]:
+    """List all model names available locally."""
+    models = await list_local_models(api_url)
+    return [model["model"] for model in models]
+
+
+async def is_local_model(model: str, api_url: str | None = None) -> bool:
```
daryllimyt, line 71: lets add lru_cache here?

tracecat/llm/ollama.py
```diff
@@ -0,0 +1,129 @@
...
+            {"role": "system", "content": system_prompt},
+            {"role": "user", "content": prompt},
+        ]
+    elif memory:
+        messages = [
+            *memory,
+            {"role": "user", "content": prompt},
+        ]
+    else:
+        messages = [{"role": "user", "content": prompt}]
```
daryllimyt: similar comments from openai section below also apply here

tracecat/llm/ollama.py
```diff
@@ -0,0 +1,129 @@
...
+        "🧠 Calling LLM chat completion",
+        provider="ollama",
+        model=model,
+        prompt=prompt,
+    )
+
+    kwargs = {
+        "model": model,
+        "messages": messages,
+        "format": format.model_json_schema() if format else None,
```
daryllimyt: does passing `None` here have the same effect as omitting the kwarg?

tracecat/llm/ollama.py
```diff
@@ -0,0 +1,129 @@
...
+    for model in models:
+        try:
+            await client.pull(model)
+        except Exception as e:
+            logger.warning(
+                "Failed to pull model",
+                model=model,
+                error=e,
+            )
+    return responses
```
daryllimyt, line 54: interesting, isn't this operation IO bound? does it actually overwhelm ollama?

daryllimyt (changes requested):
Looks good, some minor clarifications. Nice use of overloads

tests/unit/test_llms.py
daryllimyt: I know this is just a test, but the use of `ExpectedChatResponseType` as an enum mapping to type seems somewhat unconventional

tracecat/llm/ollama.py
topher-lo: It's for type hinting. Ollama doesn't expose the `Model` pydantic model...

tracecat/llm/ollama.py
topher-lo: Some users in network constrained environments faced some issues pulling multiple models at the same time. Yes it's IO bound but Ollama loads models in chunks and sometimes it losses track of some chunks and errors occur.

Found pulling one by one fixed it.

tests/unit/test_llms.py
topher-lo: I wanted  an explicit mapping from the provider to the expected response type. This is to defend again possible changes in response types. E.g. openai's `beta.chat.completions.parse` endpoint returns a different response type than the standard `chat.create` endpoint.

Anyway happy to just take your suggestion tbh

tracecat/llm/openai.py
topher-lo: This is correct. Memory should contain system prompt already.

tracecat/llm/ollama.py
topher-lo: They are mutually exclusive, not sure if I can use the "concise" solution

tracecat/llm/ollama.py
topher-lo: No. Will remove

tracecat/llm/ollama.py
topher-lo: No idea. Used conditional instead to be safe.

PR #937, opened 2025-03-11
feat(ui): allow JSONB insert and adding the collapsable JSON

### LINKED ISSUE #931: Cannot add JSONB value to lookup table via UI
Expected object got string.

Image

Jarro01X:
## Checklist

- [X] Read [CONTRIBUTING.md](https://github.com/TracecatHQ/tracecat/blob/main/CONTRIBUTING.md).
- [X] PR title is short and non-generic (see previously [merged PRs](https://github.com/TracecatHQ/tracecat/pulls?q=is%3Apr+is%3Aclosed) for examples).
- [X] PR only implements a single feature or fixes a single bug.
- [X] Tests passing (`uv run pytest tests`)?
- [X] [Lint](https://docs.astral.sh/ruff/) / [pre-commits](https://pre-commit.com/) passing (`pre-commit run --all-files`)?

## Description

Adding a fix that makes it possible to add JSONB to tables, and the ability to collapse or retract items longer then 15 lines

## Related Issues

Fixes #931 

## Screenshots
See comments below

## Steps to QA

1. Go to tables
2. Create a table with JSONB column
3. Add a valid JSONB
4. Make sure it is longer then 15 lines in case you want to test collapse and retractability


frontend/src/components/tables/table-view.tsx
```diff
@@ -33,11 +33,32 @@ export function DatabaseTable({ table: { columns } }: { table: TableRead }) {
...
+                      setIsExpanded(!isExpanded);
+                    }}
+                  >
+                    {isExpanded ? "↑" : "↓"}
+                  </button>
+                )}
+                <pre className="text-xs overflow-hidden">
+                  {value && formatJsonWithLimit(value, isExpanded ? undefined : 15)}
+                </pre>
+              </div>
```
daryllimyt: I'm thinking if we could just replace all this with `react-json-view`'s JSON viewer (the one we use throughout the UI). An example of this being used can be found [here](https://github.com/TracecatHQ/tracecat/blob/3aca25830862154e3aba4928e49c1bb3d7975c6a/frontend/src/components/executions/event-details.tsx#L483-L520). 

The nice thing about just using this would be that all the niceties like copying, list count, expand/collapse are all built-in and it works for any JSON data, so we don't have to typecheck anymore here.

daryllimyt (changes requested):
Added comments suggesting we reuse the json viewer we've been using all this while. Would be good to have your thoughts here too @topher-lo

topher-lo:
Agreed on the JSON viewer. Also we should consider only showing prettified JSON if expanded. Otherwise the tables will get too with <5-10 rows and large nested jsons (which is expected for LLM memory and security related logs)

Jarro01X:
oh sorry, I hadn't noticed you guys were using that. Let me go ahead and make the adjustments

Jarro01X:
@topher-lo done, the JSON gets prettified very nicely if you choose the Nested option. This is probably some of the best UI/UX I've ever seen for viewing JSONs, this is great! Much better now honestly

Screenshot_107

Jarro01X:
Alright there we go, everything is running well on my end. The branch hadn't been updated so running the new/modified tests that are in the CI/CD pipeline wasn't working

topher-lo:
Looking good. Just waiting on @daryllimyt for final review

frontend/src/components/tables/table-view.tsx
```diff
@@ -10,6 +10,8 @@ import { useListRows } from "@/lib/hooks"
 import { DataTable } from "@/components/data-table"
 import { TableViewAction } from "@/components/tables/table-view-action"
 import { TableViewColumnMenu } from "@/components/tables/table-view-column-menu"
+import { JsonViewWithControls } from "../workbench/events/events-selected-action"
+import { TooltipProvider } from "@radix-ui/react-tooltip"
```
daryllimyt: Prefer shadcn tooltip over the raw radix one
```suggestion
import { TooltipProvider } from "@/components/ui/tooltip"
```

frontend/src/components/tables/table-view.tsx
```diff
@@ -10,6 +10,8 @@ import { useListRows } from "@/lib/hooks"
 import { DataTable } from "@/components/data-table"
 import { TableViewAction } from "@/components/tables/table-view-action"
 import { TableViewColumnMenu } from "@/components/tables/table-view-column-menu"
+import { JsonViewWithControls } from "../workbench/events/events-selected-action"
```
daryllimyt: Could we:

1. Move `JsonViewWithControls` into a separate file under `/components/json-viewer.tsx`
2. Import it in thie file and in `/workbench/events/events-selected-action`  using its absolute path prefixed with @ i.e. `@/components/json-viewer.tsx`

daryllimyt (changes requested):
Looks good, just a couple of small adjustments

daryllimyt (approved):
LGTM! 🚢

PR #951, opened 2025-03-17
fix(integrations): Fix Okta user lookup query + result parsing

daryllimyt:
# Description
- Wrap inputs.email with quotations
- Don't return the first element, but instead take .data and return the list

### DIFF
--- a/registry/tracecat_registry/templates/detect/lookup_user_by_email/okta.yml
+++ b/registry/tracecat_registry/templates/detect/lookup_user_by_email/okta.yml
@@ -20,7 +20,7 @@ definition:
     - ref: search_query
       action: core.transform.reshape
       args:
-        value: profile.email eq ${{ inputs.email }}
+        value: profile.email eq "${{ inputs.email }}"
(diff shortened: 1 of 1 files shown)


registry/tracecat_registry/templates/detect/lookup_user_by_email/okta.yml
```diff
@@ -30,4 +30,5 @@ definition:
           Authorization: "SSWS ${{ SECRETS.okta.OKTA_API_TOKEN }}"
         params:
           search: ${{ FN.url_encode(steps.search_query.result) }}
-  returns: ${{ steps.search_user_by_email.result[0] }}
+  # core.http_request returns data, headers, and status_code
+  returns: ${{ steps.search_user_by_email.result.data }}
```
daryllimyt: Do we still want to grab the first result only? @topher-lo

topher-lo: Yeah there should only be one user

PR #968, opened 2025-03-21
feat(integrations): Add `core.sql.query` udf action

daryllimyt:
# Security
- Passing raw SQL here is highly unsafe--it's reliant on the user to perform the sanitization. cc @topher-lo.
# Solution
- Make the transaction read only for `postgresql`
- Added additional validation to check for statement delimiters and take 
- Zeropath scan seems ok

# Description
- Add `core.sql.query`, which currently only allows you to query with a SELECT statement.
- Added a lot of sanitization/validation for security
- Only supports `postgresql` driver atm. Uses `psycopg[binary]` for simplicity
- Requires SQL_HOST, SQL_PORT, SQL_USER, SQL_PASS env vars.
- Database name is optional

# Guardrails
We sanitize on any loopback addresses and especially `postgres_db`. Note that we also don't allow underscores in the url anyways (this is according to some RFCs, can't recall).

# Testing
Added unit tests.
## QA
Tested with my local Tracecat `postgres_db` exposed through tcp:// over ngrok.

Screenshot 2025-03-22 at 00 37 02

Example of malicious query being blocked by postgres read only

Screenshot 2025-03-22 at 01 44 31

### DIFF
--- a/frontend/src/components/icons.tsx
+++ b/frontend/src/components/icons.tsx
@@ -4,6 +4,7 @@ import {
   BoxIcon,
   Building2Icon,
   Cpu,
+  Database,
   Globe,
   ListChecks,
   Mail,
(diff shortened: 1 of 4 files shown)


topher-lo:
Closing due to risk exposure.

PR #990, opened 2025-04-01
feat(engine): Implement core object store functionality

daryllimyt:
# Motivation
- Avoid returning large payloads from Temporal activities by storing them in Minio.

# Todos
- [ ] Additional testing matrix with different store configurations 

# Description
- Phase 1: Roll out object store related functionality behind feature flag. Ensures that we remain fully backward compatible and tested
- Phase 2: Complete additional object store features like presigned urls in the Ui for downloading large payloads, object expiration, conditional storage of payloads over a certain size threshold. 
- Phase 3: Security. TBH this is part of the matrix testing config. Just need to double check that we can use a non-root user account to r/w into minio


topher-lo:
Stale. Also no longer urgent due to compression.

PR #995, opened 2025-04-04
feat(engine): Executor return value size limit

daryllimyt:
# Motivation
- Prevent executor from returning paylaods larger than 128KB
- Added `pydantic_core.to_jsonable_python` as fallback serializer for non-standard types


tracecat/executor/router.py
```diff
@@ -39,7 +43,16 @@ async def run_action(
     log.info("Starting action")
 
     try:
-        return await dispatch_action_on_cluster(input=action_input, session=session)
+        result = await dispatch_action_on_cluster(input=action_input, session=session)
+        serialized = orjson.dumps(result, default=to_jsonable_python)
+        logger.warning("Serialized result", serialized=serialized)
```
topher-lo:
```suggestion
```

topher-lo (approved):

PR #998, opened 2025-04-05
feat(ui): Add hotkeys to toggle events panel and action panel

daryllimyt:
# Hotkeys
- CMD+E for events panel (left side)
- CMD+U for action panel (right side)


topher-lo (approved):
clean

PR #1006, opened 2025-04-08
refactor!: Drop ollama service support

topher-lo:
We do not want to be tied to Ollama (e.g. vLLM is a faster more production ready inference server).

## What changed
- Dropped ollama from docker compose
- Dropped ollama env vars
- Dropped pull model on executor startup
- Dropped ollama from tests

Users that want to use Ollama can still do so by spinning up their own server then specifying `base_url` in the Ollama UDF integration.


docker-compose.dev.yml
```diff
@@ -58,7 +58,6 @@ services:
       TRACECAT__REMOTE_REPOSITORY_URL: ${TRACECAT__REMOTE_REPOSITORY_URL}
       # AI
       TRACECAT__PRELOAD_OSS_MODELS: ${TRACECAT__PRELOAD_OSS_MODELS}
```
daryllimyt: do we still need these?

tracecat/config.py
```diff
@@ -181,7 +181,6 @@
     (models := os.getenv("TRACECAT__PRELOAD_OSS_MODELS")) and models.split(",")
 ) or []
```
daryllimyt: and this?

daryllimyt (approved):
Couple of clarifications

PR #1026, opened 2025-04-10
feat(integrations): Add stdout parser and podman stderr exceptions

topher-lo:


topher-lo:
Deprecating. We're dropping podman for k8s

PR #1052, opened 2025-04-15
feat(integrations): Splunk search job and results

topher-lo:
An Action Template that:
1. Posts search query with http request
2. Extract search ID with simple python apply
3. Polls search job endpoint until `dispatchState` is `DONE`
4. Gets results

### DIFF
--- a/registry/tracecat_registry/templates/detect/query_events/splunk.yml
+++ b/registry/tracecat_registry/templates/detect/query_events/splunk.yml
@@ -61,7 +61,7 @@ definition:
         headers:
           Authorization: Bearer ${{ SECRETS.splunk.SPLUNK_API_KEY }}
         poll_condition: >
-          lambda x: '<s:key name="dispatchState">DONE</s:key>' in x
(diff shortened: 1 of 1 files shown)


registry/tracecat_registry/templates/detect/query_events/splunk.yml
```diff
@@ -0,0 +1,77 @@
...
+    - ref: get_search_status
+      action: core.http_poll
+      args:
+        url: ${{ inputs.base_url }}/services/search/jobs/${{ steps.search_id.result }}
+        method: GET
+        verify_ssl: ${{ inputs.verify_ssl }}
+        headers:
+          Authorization: Bearer ${{ SECRETS.splunk.SPLUNK_API_KEY }}
+        poll_condition: >
+          lambda x: '<s:key name="dispatchState">DONE</s:key>' in x
```
topher-lo:
```suggestion
          lambda x: '<s:key name="dispatchState">DONE</s:key>' in x.get('data')
```

PR #1063, opened 2025-04-17
Add Anthropic support via Bedrock

cschmidt0121:
## Checklist

- [ ] Tests passing (`uv run pytest tests`)?

## Description

I wanted to use Claude via Bedrock in a way similar to how tracecat supports Ollama/OpenAI, so I implemented that. It only supports role-based/environment variable based auth; I'm not sure if others would want to store keys as secrets (AWS doesn't recommend this). 

Sorry, couldn't get tests to run. Not sure if the registry directory structured changed or what, but I don't _think_ this will break any tests

```    from tracecat.registry.actions.models import (
tracecat/registry/actions/models.py:20: in <module>
    from tracecat_registry import RegistrySecret
E   ModuleNotFoundError: No module named 'tracecat_registry'```


## Screenshots / Recordings
![Screenshot 2025-04-17 at 11 44 28 AM](https://github.com/user-attachments/assets/1e454dbf-84df-4a5b-85a0-072570763011)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ dependencies = [
     "uv==0.4.10",
     "uvicorn==0.29.0",
     "virtualenv==20.27.0",
+    "anthropic[bedrock]==0.49.0",
 ]
 dynamic = ["version"]
 
--- a/registry/tracecat_registry/llm/anthropic_bedrock.py
+++ b/registry/tracecat_registry/llm/anthropic_bedrock.py
@@ -1,0 +1,52 @@
+from typing import Annotated, Any
+
+from typing_extensions import Doc
+
+from tracecat.llm.anthropic_bedrock import (
+    async_anthropic_bedrock_call,
+    DEFAULT_ANTHROPIC_MODEL,
+)
+from tracecat_registry import registry
+
+
+@registry.register(
+    default_title="Call Anthropic Bedrock (messages)",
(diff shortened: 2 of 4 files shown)


topher-lo:
Thanks for opening this! Much needed. Is it not possible to have one client for both bedrock and anthropic direct?

EverdredIIDX:
> Thanks for opening this! Much needed. Is it not possible to have one client for both bedrock and anthropic direct?

Well, on the SDK side they _are_ two different clients (`AsyncAnthropic` by `AsyncAnthropicBedrock`). On the surface their functions seem pretty much identical, but I wouldn't be surprised if there were some subtle differences. More importantly though auth will be different. Anthropic customers will want to use an API key while most AWS users will want to use role-based auth. 

I will defer to you on that though! I implemented exactly what I needed and then stopped there :D

topher-lo:
> > Thanks for opening this! Much needed. Is it not possible to have one client for both bedrock and anthropic direct?
> 
> Well, on the SDK side they _are_ two different clients (`AsyncAnthropic` by `AsyncAnthropicBedrock`). On the surface their functions seem pretty much identical, but I wouldn't be surprised if there were some subtle differences. More importantly though auth will be different. Anthropic customers will want to use an API key while most AWS users will want to use role-based auth.
> 
> I will defer to you on that though! I implemented exactly what I needed and then stopped there :D

I'm going to built upon this PR 🫡

PR #1078, opened 2025-04-21
build(app): Add uv.lock

daryllimyt:
- **Remove uv.lock from .gitignore**
- **Add uv.lock**

# Description
- We should start using `uv sync` as it reads from uv.lock and skips dependency resolution.


daryllimyt:
Closed in favor of #1312

PR #1082, opened 2025-04-22
fix: Drop multiindex support

topher-lo:
## What changed
- Only support single index now
- Use sqlalchemy inspect to get indexes
- Renamed "natural key" to "unique index"
- Changed key icon to database spark
- Added ValueError (reraised as HTTP 400 error at router) exception to table service if update column with `is_index=True` and an index already exists
- Catch 400 exception in frontend updateColumn use effect

## Fixes
Silent bug where multi index path was reachable by user.

## QA
- Create new table
- Create index
- Create another index (errors as notification)

Screenshot 2025-04-22 at 5 46 36 PM

### DIFF
--- a/tracecat/tables/service.py
+++ b/tracecat/tables/service.py
@@ -438,22 +438,6 @@ class BaseTablesService(BaseService):
         await self.session.flush()
 
     @require_access_level(AccessLevel.ADMIN)
-    async def create_multi_unique_index(
-        self, table: Table, column_names: list[str]
-    ) -> None:
-        """Create a unique index on multiple columns."""
-        full_table_name = self._full_table_name(table.name)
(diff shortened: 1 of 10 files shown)


tracecat/tables/service.py
```diff
@@ -450,14 +429,30 @@ async def create_unique_index(self, table: Table, column_names: list[str]) -> No
...
+                    sanitized_column,  # Column to index
                 ),
             )
         )
 
         # Commit the transaction
         await self.session.flush()
 
+    @require_access_level(AccessLevel.ADMIN)
+    async def create_multi_unique_index(
```
daryllimyt: intentional?

daryllimyt (approved):
lgtm

PR #1086, opened 2025-04-23
fix(ui): Handle oauth/saml login loading states gracefully

daryllimyt:
When logging in with oauth/saml we flicker to the login page for a split second

# Screens
## Before

https://github.com/user-attachments/assets/479878ac-e335-456b-8bcd-362e7fb93492

## After

https://github.com/user-attachments/assets/a1a63f2b-916a-4eb8-9382-bb827bf295e5


topher-lo (approved):
Smooth

PR #1109, opened 2025-05-07
feat(integrations): Add openapi to action template script

daryllimyt:
Blocked by https://github.com/TracecatHQ/tracecat/pull/1111
# Description
- Script that takes OAS3.0 spec and converts it into Tracecat action templates. Requires tracecat and tracecat_registry to be installed.
- Not perfect but will do most of the conversion work
- You can configure the generation by passing `--config`. 
  - See schema and example inputs in the Pydantic models
  - An example config file is provided.

# Learnings
- Having a notion of an `undefined` or `Unset` type/sentinel would have made this much easier. We don't have a nice way to omit params in action template `args`
  
# Testing
- Tested locally with successful sync of jira and sentinelone openapi specs. Performed successful query of jira users endpoint.
- Able to ingest ms graph openapi spec as well. Didn't test with the full spec but able to ingest a subset of `/admin*` endpoints

### DIFF
--- a/scripts/openapi_to_template_example_config.yaml
+++ b/scripts/openapi_to_template_example_config.yaml
@@ -1,0 +1,26 @@
+# Example config for generating templates from an OpenAPI spec
+endpoints:
+  # include:
+  #   like:
+  #     - "/pet/{petId}*"
+  # exclude:
+  #   like:
+  #     - "**/api-doc/api-details/**"
+definition_overrides:
+  # Namespace will be used as a prefix, not
+  namespace: pet_store
+  # All other fields that appear here, override
+  # These override the default values for the actions
+  display_group: PetStore
+
+# How to inject auth logic
+auth:
+  secrets:
+    - name: petstore
+      keys: ["API_KEY"]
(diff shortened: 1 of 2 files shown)


scripts/openapi_to_template_example_config.yaml
```diff
@@ -0,0 +1,26 @@
+# Example config for generating templates from an OpenAPI spec
+endpoints:
+  # include:
+  #   like:
+  #     - "/pet/{petId}*"
+  # exclude:
+  #   like:
+  #     - "**/api-doc/api-details/**"
+definition_overrides:
+  # Namespace will be used as a prefix, not
```
topher-lo, line 10: Clarification? Otherwise lgtm

topher-lo (approved):

scripts/openapi_to_template_example_config.yaml
daryllimyt, line 10:
```suggestion
  # Namespace will be used as a prefix instead of override
```

PR #1114, opened 2025-05-08
feat(integrations): Slack ChatOps starter kit

topher-lo:


tracecat/expressions/functions.py
```diff
@@ -746,7 +751,9 @@ def now() -> datetime:
 
 def utcnow() -> datetime:
     """Return the current timezone-aware datetime."""
-    return datetime.now(UTC)
+    return datetime.now(
+        UTC,
+    )
```
daryllimyt:
```suggestion
    return datetime.now(UTC)

```

tracecat/webhooks/router.py
```diff
@@ -59,10 +63,12 @@ async def incoming_webhook(
         trigger_type=TriggerType.WEBHOOK,
     )
     if echo:
+        if empty_echo:
+            return Response(status_code=200)
         try:
             response["payload"] = await request.json()
         except Exception as e:
-            logger.warning("Failed to echo request payload body", error=str(e))
+            logger.warning("Failed to decode request payload body", error=str(e))
```
daryllimyt: This code path is only related to echo and is an internal facing log
```suggestion
            logger.warning("Failed to decode request payload body during echo", error=str(e))
```

docker-compose.dev.yml
```diff
@@ -147,6 +147,7 @@ services:
       ]
 
   ui:
+    # image: ghcr.io/tracecathq/tracecat-ui:${TRACECAT__IMAGE_TAG:-0.34.3}
```
daryllimyt:
```suggestion
```

daryllimyt (commented):
Some changes

daryllimyt (changes requested):
Some changes

daryllimyt (approved):

PR #1116, opened 2025-05-11
fix(ui): Add defensive graph pruning on canvas init + layout

daryllimyt:
# Motivaiton
Aim to eliminate ghost edges which makes UX awkward. Solving at the root cause would be much more challenging and involved than retroactively patching the graph object.
# Changes
- **We don't actually use pendingDeleteEdges**
- **Drop ghost edges in graph prune step**
- **Simplify ephemeral node pruning logic**
- **pruneGraphObject -> pruneReactFlowInstance**
- **Defensive graph prune on layout + initialize**

# Description
- Update graph pruning logic to drop ghost edges
- Add tests for grpah pruning logic (TODO: Add to CI)
- Run pruning on canvas initialization and onLayout handler
- Simplify `isEphemeral` check in prune step

### DIFF
--- a/frontend/src/components/workbench/canvas/canvas.tsx
+++ b/frontend/src/components/workbench/canvas/canvas.tsx
@@ -40,7 +40,7 @@ import Dagre from "@dagrejs/dagre"
 import { MoveHorizontalIcon, MoveVerticalIcon, PlusIcon } from "lucide-react"
 
 import { useDeleteAction } from "@/lib/hooks"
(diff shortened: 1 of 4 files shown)


topher-lo:
This is godly

## Before
Screenshot 2025-05-10 at 11 56 53 AM

## After

Screenshot 2025-05-11 at 6 35 36 PM

topher-lo (approved):

PR #1122, opened 2025-05-12
feat(ui): Split run if by operators

daryllimyt:
We're blocked by our inability to handle multiline action node handles well. It'll look bad until this is resolved.

# Vision
The pill overflows past the incoming edge arrowhead and the node body.
image


daryllimyt:
Revisit when this becomes insufferable

PR #1125, opened 2025-05-12
chore(build): Remove deprecated env vars

topher-lo:
## What this PR does
- Drops `TRACECAT__REMOTE_REPOSITORY_X` env vars
- Moves allowed email attributes env var into action
- Drops settings override vars (logic still intact but not needed in infra-as-code e.g. still can be added in emergencies via task definition directly)

## Why this PR is needed
Our environment variables are quite stable now. Want to keep it this way as we're now supporting three different deployment options:
- AWS via Terraform
- Helm chart
- Docker compose


deployments/aws/ecs/locals.tf
```diff
@@ -43,15 +43,10 @@ locals {
       TRACECAT__APP_ENV                               = var.tracecat_app_env
       TRACECAT__AUTH_ALLOWED_DOMAINS                  = var.auth_allowed_domains
       TRACECAT__AUTH_TYPES                            = var.auth_types
-      TRACECAT__SETTING_OVERRIDE_SAML_ENABLED         = var.setting_override_saml_enabled
-      TRACECAT__SETTING_OVERRIDE_OAUTH_GOOGLE_ENABLED = var.setting_override_oauth_google_enabled
-      TRACECAT__SETTING_OVERRIDE_BASIC_AUTH_ENABLED   = var.setting_override_basic_auth_enabled
```
daryllimyt, line 48: not sure if we should remove these, as the main reason of having TRACECAT__SETTING_OVERRIDE_XXX was to override these. If we remove these and we (hopefully don't) need this, we will have to patch these again. I'm happy to remove these from docker, because it's easier to reset or override db values easily, but for tf deploys we don't have a way to do so

daryllimyt (approved):
LGTM, one clarification on deleting the auth method overrides

PR #1150, opened 2025-05-26
feat(integrations): Parse semi-colon sep `Content-Type` in webhook

topher-lo:
Our matching algorithm on `Content-Type` is to strict. Will fail if users add multiple content-type (e.g. `application/x-ndjson: charset=utf-8)

This PR splits up all specified content types and does a if contains check.

### DIFF
--- a/tracecat/webhooks/dependencies.py
+++ b/tracecat/webhooks/dependencies.py
@@ -116,38 +116,43 @@ async def parse_webhook_payload(
     if not body:
         return None
 
-    match content_type:
-        case "application/x-ndjson" | "application/jsonlines" | "application/jsonl":
-            # Newline delimited json
-            try:
-                lines = body.splitlines()
-                result = [orjson.loads(line) for line in lines]
-            except orjson.JSONDecodeError as e:
-                logger.error("Failed to parse ndjson payload", error=e)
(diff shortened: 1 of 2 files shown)


tracecat/webhooks/dependencies.py
```diff
@@ -116,38 +116,43 @@ async def parse_webhook_payload(
...
-                    detail="Invalid json payload",
-                ) from e
+    # Normalize content type for comparison (lowercase and strip whitespace)
+    normalized_content_type = content_type.lower().strip() if content_type else ""
+
+    if any(
+        ct in normalized_content_type
+        for ct in ["application/x-ndjson", "application/jsonlines", "application/jsonl"]
+    ):
+        # Newline delimited json
```
daryllimyt: nit: why not something like:
```python
def parse_content_type(content_type: str) -> tuple[str, dict[str, str]]:
    mime_type, *params = content_type.strip().split(";")
    metadata = {}
    for param in params:
        key, value = param.strip().split("=", 1)
        metadata[key] = value.strip('"')
    return mime_type, metadata
```
    
to parse the content type + metadata? Would we ever have to handle this content-type metadata?

daryllimyt (approved):

PR #1202, opened 2025-06-18
refactor(registry+app): Move field components to registry

daryllimyt:
# Motivation
This is the logically correct place to add these annotations as they are tightly coupled to UDFs. 

# TODO
Make the fields frozen

## Summary by cubic
Moved field component classes from the app to the registry package to centralize and simplify field definitions.

- **Refactors**
  - Relocated all field component classes to `tracecat_registry._internal.fields`.
  - Updated imports across the codebase to use the new location.


topher-lo:
Stale.

PR #1310, opened 2025-07-30
feat(api+integrations): Insert many rows

topher-lo:
## Summary by cubic
Added support for inserting multiple rows into a table at once, with optional upsert to update existing rows based on unique indexes.

- **New Features**
  - New `insert_rows` API for batch inserts with upsert support.
  - Upsert uses unique indexes and preserves existing values if new data omits columns.

- **Bug Fixes**
  - Prevents overwriting existing column values with NULL during upsert when rows have missing fields.


topher-lo:
@cubic-dev-ai review again

PR #1362, opened 2025-08-21
fix(alertmedia): Lat & Long can be zero

bishopdotdev:
Fix for lat and long allowed to be zero.
    

---

## Summary by cubic
Fix update_trip to include latitude and longitude when the value is 0. This prevents zero coordinates from being omitted, so trips at the equator or prime meridian are updated correctly.

### DIFF
--- a/packages/tracecat-registry/tracecat_registry/templates/tools/alertmedia/update_trip.yml
+++ b/packages/tracecat-registry/tracecat_registry/templates/tools/alertmedia/update_trip.yml
@@ -99,9 +99,9 @@ definition:
                   destination_address["city"] = city
               if country_code:
                   destination_address["country"] = country_code
-              if latitude:
+              if latitude or latitude == 0:
                   destination_address["latitude"] = latitude
-              if longitude:
+              if longitude or longitude == 0:
                   destination_address["longitude"] = longitude
               if destination_address:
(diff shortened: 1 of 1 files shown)


packages/tracecat-registry/tracecat_registry/templates/tools/alertmedia/update_trip.yml
```diff
@@ -99,9 +99,9 @@ definition:
                   destination_address["city"] = city
               if country_code:
                   destination_address["country"] = country_code
-              if latitude:
+              if latitude or latitude == 0:
```
topher-lo:
```suggestion
              if latitude is None:
```

packages/tracecat-registry/tracecat_registry/templates/tools/alertmedia/update_trip.yml
```diff
@@ -99,9 +99,9 @@ definition:
...
                   destination_address["latitude"] = latitude
-              if longitude:
+              if longitude or longitude == 0:
```
topher-lo:
```suggestion
              if longitude is None:
```

topher-lo (changes requested):
More pythonic to check for None-ness directly.

Issue #1380, opened 2025-08-27
HTTP request action treats 302 redirects as errors when follow_redirects=false


xkratos:
Current [core http](https://github.com/TracecatHQ/tracecat/blob/main/packages/tracecat-registry/tracecat_registry/core/http.py) has some issue.

### Current Behavior
When using the HTTP request action with follow_redirects: false, any 3xx response is treated as an error and the task fails. Instead, the action should return the full raw response so the workflow can handle redirects manually. This includes preserving all headers (e.g., Location, all Set-Cookie headers), the body (even if empty), reason phrase, and the final request URL.

### Proposed Fix
When follow_redirects is false:
Treat all 3xx (300–308) as successful results, not errors.

PR #1381, opened 2025-08-27
feat(integrations): Core HTTP request action now only raises for HTTP ≥ 400

xkratos:
## Description

This PR changes the HTTP request action’s error semantics to only raise for HTTP status codes ≥ 400.
Previously, non-2xx (including 3xx redirects) were treated as errors in some cases, which prevented workflows from inspecting redirect responses when follow_redirects: false.

What’s changed
Error policy: 1xx/2xx/3xx no longer raise; only 4xx/5xx raise.

## Related Issues
HTTP request action treats 302 redirects as errors when follow_redirects=false  #1380

    

---

## Summary by cubic
Update the HTTP request action to only raise on HTTP 4xx/5xx. Redirects (3xx) no longer raise, letting workflows inspect 302 responses when follow_redirects=false and addressing #1380.

### DIFF
--- a/packages/tracecat-registry/tracecat_registry/core/http.py
+++ b/packages/tracecat-registry/tracecat_registry/core/http.py
@@ -589,7 +589,8 @@ async def http_request(
                     data=form_data,
                     files=httpx_files_param,
                 )
-            response.raise_for_status()
+            if response.status_code >= 400:
+                response.raise_for_status()
         except httpx.HTTPStatusError as e:
             error_message = _http_status_error_to_message(e)
             logger.error(


topher-lo:
Hey @xkratos great idea. A few points about execution:

- Prefer not to override httpx default raise for status behaviors
- Would prefer to use an argument for users to control this behavior and instead of hard coding > 400. Maybe do a check status_code in instead? Or a Boolean (raise on 400 above?)

We could all this new argument error codes and its set to None by default (ie current logic)

xkratos:
> Prefer not to override httpx default raise for status behaviors.

I think this behavior is a little bit wired, other lib or language don't treat it as an error.
eg: Python – Requests, Go – net/http,Rust – reqwest, Java – HttpClient (Java 11+), Kotlin/Java – OkHttp.
I could do this it's a bit verbose.
```
           # raise_on_redirect is the new argument.
            if not response.is_redirect or raise_on_redirect:
                response.raise_for_status()
```

topher-lo:
> > Prefer not to override httpx default raise for status behaviors.
> 
> I think this behavior is a little bit wired, other lib or language don't treat it as an error. eg: Python – Requests, Go – net/http,Rust – reqwest, Java – HttpClient (Java 11+), Kotlin/Java – OkHttp. I could do this it's a bit verbose.
> 
> ```
>            # raise_on_redirect is the new argument.
>             if not response.is_redirect or raise_on_redirect:
>                 response.raise_for_status()
> ```

I see if not aligned with other best practices. You're right httpx is weird. Happy to accept your PR.

topher-lo (approved):

PR #1411, opened 2025-09-05
feat(api)!: Poll condition now polls UNTIL condition is met

topher-lo:


packages/tracecat-registry/tracecat_registry/core/http.py
```diff
@@ -681,22 +683,32 @@ def user_defined_predicate(response: httpx.Response) -> bool:
...
-            headers=dict(response.headers.items()),
-            data=data,
-            status_code=response.status_code,
-        )
-        return predicate(args)
+        args = {
+            "headers": dict(response.headers.items()),
+            "data": data,
+            "status_code": response.status_code,
+        }
```
daryllimyt: any reason for removing the validation here?

PR #1412, opened 2025-09-08
feat(api): Case records CRUD

topher-lo:
## Summary by cubic
Add full CRUD for case-linked entity records, including list, create, link, update, unlink, and delete APIs, plus generated frontend SDK/types. Adds a per-case record limit with a new config flag.

- **New Features - New features added**
  - Backend: /cases/{case_id}/records endpoints
    - GET list, GET by id, POST create, POST /link, PATCH update, DELETE /unlink (soft), DELETE (hard)
    - Responses include entity metadata and record data (CaseRecordRead)
  - Frontend: generated types and SDK methods
    - caseRecordsListCaseRecords, Create, Get, Update, Delete, LinkEntityRecord, UnlinkCaseRecord

- **Migration**
  - New env var: TRACECAT__MAX_RECORDS_PER_CASE (default 50). Requests above the limit return 400.

### DIFF
--- a/tracecat/cases/records/models.py
+++ b/tracecat/cases/records/models.py
@@ -6,9 +6,11 @@ import uuid
 from datetime import datetime
 from typing import Any, Literal
 
-from pydantic import BaseModel, Field
+import orjson
+from pydantic import BaseModel, Field, field_validator
 
 from tracecat import config
+from tracecat.records.model import MAX_BYTES
 
 
 class CaseRecordCreate(BaseModel):
@@ -24,6 +26,15 @@ class CaseRecordCreate(BaseModel):
         default_factory=dict,
         description="Entity record data",
     )
+
+    @field_validator("data")
+    @classmethod
+    def enforce_value_sizes(cls, v: dict[str, Any]) -> dict[str, Any]:
+        """Match record model: limit each field's serialized size to 200KB."""
+        for key, value in v.items():
+            if len(orjson.dumps(value)) > MAX_BYTES:
+                raise ValueError(f"Field '{key}' value must be less than 200 KB")
+        return v
 
 
 class CaseRecordLink(BaseModel):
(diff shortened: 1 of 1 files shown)


tracecat/cases/records/models.py
```diff
@@ -0,0 +1,94 @@
...
+    entity_key: str = Field(
+        ...,
+        min_length=1,
+        max_length=255,
+        description="Key of the entity type",
+    )
+    data: dict[str, Any] = Field(
+        default_factory=dict,
+        description="Entity record data",
+    )
```
daryllimyt, line 28: should we enforce size limits here?

tracecat/cases/records/models.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class CaseRecordUpdate(BaseModel):
+    """Model for updating a case record's entity data."""
+
+    data: dict[str, Any] = Field(
+        ...,
+        description="Updated entity record data",
+    )
```
daryllimyt, line 55: size limits?

tracecat/cases/records/router.py
```diff
@@ -0,0 +1,435 @@
...
+        data=updated_record.record.data,
+        created_at=updated_record.created_at,
+        updated_at=updated_record.updated_at,
+    )
+
+
+@router.delete(
+    "/{case_record_id}/unlink",
+    status_code=status.HTTP_200_OK,
+)
```
daryllimyt: nit: imo this should be either:
- POST with /unlink, since unlink is a verb
- DELETE with /link, since this acts on the /link resource

topher-lo: Probably PATCH to keep consistent with entities /activate and /deactivate

tracecat/cases/records/models.py
topher-lo: Done but also we enforce these in the records service too

PR #1428, opened 2025-09-12
feat : Lookup table on multiple columns

y0no:
## Checklist

- [X] Read [CONTRIBUTING.md](https://github.com/TracecatHQ/tracecat/blob/main/CONTRIBUTING.md).
- [X] PR title is short and non-generic (see previously [merged PRs](https://github.com/TracecatHQ/tracecat/pulls?q=is%3Apr+is%3Aclosed) for examples).
- [X] PR only implements a single feature or fixes a single bug.
- [ ] Tests passing (`uv run pytest tests`)?
- [X] [Lint](https://docs.astral.sh/ruff/) / [pre-commits](https://pre-commit.com/) passing (`pre-commit run --all-files`)?

## Description

This PR aims to provide a way to lookup on table over multiple columns and values. The implementation is compatible with the old syntax and the new one.

Currently the tests are not implemented in case the feature is not validated. I think there is a better way to implement this feature for example with a filter field which accept an array of key=value strings..But I don't know how to do it without breaking the old syntax.

## Steps to QA

1. Add a lookup or lookup_many node on a workflow
2. Add two column names on column field
3. Add two value names on value field
4. See if a row is selected with multiple search criteria.

    

---

## Summary by cubic
Add multi-column filtering to table lookups. You can now pass lists of columns and values; single-column calls still work.

- **New Features**
  - lookup and lookup_many accept column: str | list[str] and value: Any | list[Any].
  - Validates one-to-one column/value pairs; raises a clear error on mismatch.
  - lookup now calls lookup_many with limit=1 for consistent behavior.
  - Sends column/value lists to TablesService.lookup_rows; existing limit checks remain.

### DIFF
--- a/packages/tracecat-registry/tracecat_registry/core/agent.py
+++ b/packages/tracecat-registry/tracecat_registry/core/agent.py
@@ -154,7 +154,7 @@ async def agent(
     model_name: Annotated[str, Doc("Name of the model to use.")],
     model_provider: Annotated[str, Doc("Provider of the model to use.")],
     actions: Annotated[
-        list[str] | str,
+        list[str],
         Doc("Actions (e.g. 'tools.slack.post_message') to include in the agent."),
         ActionType(multiple=True),
     ],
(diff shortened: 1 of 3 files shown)


topher-lo:
@y0no thanks for opening this up! I'm going to close this PR for now and use it as inspo

PR #1486, opened 2025-10-02
fix(api): Non-atomic case creation logic

topher-lo:
## Summary by cubic
- Make case creation atomic and improve custom field handling. If fields fail, the whole transaction rolls back, and error messages are clear and non-technical.
- On case creation (manual and UDF), custom field row is now also created and field defaults are now parsed and applied correctly. Previously, this caused confusing behavior where a boolean custom field (with default true) would be show as "false" on case creation but then becomes "true" on refresh.

- **Bug Fixes**
  - Removed manual commits in create_case/update_case; atomic via session/context manager.
  - Always create the case_fields row; apply defaults even when no fields are provided (uses get_row).
  - Handle TracecatNotFoundError on field updates; raise a friendly error that lists attempted field names and guidance.
  - Replace DB-jargon errors with clear messages for undefined/missing custom fields.
  - Strip PostgreSQL type casts and quotes from default values using parse_postgres_default.
  - Added tests covering rollback on field errors and user-friendly messages.

### DIFF
--- a/tests/registry/test_core_cases.py
+++ b/tests/registry/test_core_cases.py
@@ -1945,7 +1945,7 @@ class TestCoreCreateCaseErrorHandling:
         # Should NOT contain technical DB details
         assert "row_id" not in error_msg.lower()
         assert (
-            "table" not in error_msg.lower() or "case_fields" not in error_msg.lower()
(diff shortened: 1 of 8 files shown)


topher-lo:
@cubic-dev-ai review again please!

PR #1499, opened 2025-10-06
feat(integrations): Adding SSL Verification option to all GoPhish commands

bishopdotdev:
## Description

Updating GoPhish templates to allow users to specify verifying SSL in the HTTP request.

    

---

## Summary by cubic
Added a verify_ssl option to all GoPhish templates so users can disable SSL verification for internal or self-hosted instances. Default is true to keep current behavior unchanged.

- **New Features**
  - Added verify_ssl input to all GoPhish commands and passed it to core.http_request.
  - Supports self-signed certs and internal networks without SSL errors.


bishopdotdev:
Cubic didn't like this one 😅 lol

@daryllimyt & @topher-lo feel free to reject if you don't want to allow not using ssl verification but know this is our issue currently with our GoPhish instance 😅

bishopdotdev:
We can wait to use this once we have things fixed in our AWS for that, just know it might take some time.

topher-lo (approved):

PR #1505, opened 2025-10-08
feat(api): Case durations

topher-lo:
---

## Summary by cubic
Added case duration metrics with selectable first/last event anchors and safe timestamp handling to enable metrics like “time to first” vs “time to last” resolution. Definitions are stored in workspace settings and computed via a new CaseDurationService integrated into CasesService.

- **New Features**
  - New CaseDurationService to define and compute durations from case events.
  - Anchor selection: choose the first or last matching event for start/end.
  - Optional field filters and custom timestamp paths for anchors.
  - Workspace settings support case_durations; available as cases.durations.

- **Bug Fixes**
  - Prevent negative durations by ignoring end events that occur before the start when events are out of order.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,6 @@ repos:
           - --fix
           - --show-fixes # Show what files were fixed
           - --verbose
-        exclude: ^alembic/versions/
       - id: ruff-format
         verbose: true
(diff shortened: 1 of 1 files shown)


topher-lo:
@cubic-dev-ai review again

PR #1530, opened 2025-10-15
fix: Prevent duplicate validators in registry repository

daryllimyt:
## Summary
Fixes an issue where validators were being duplicated when attached to function parameters in the registry repository, and adds comprehensive unit tests for the fix.

## Changes
- **Fixed `attach_validators`**: Now checks for existing validators before adding new ones to prevent duplicates
- **Fixed `import_and_reload`**: Added fallback when module loader is missing to prevent ModuleNotFoundError
- **Added helper `_annotated_with_validators`**: Properly handles Annotated types and metadata deduplication
- **Added unit tests**: Comprehensive test coverage for validator attachment and module reloading edge cases

## Technical Details
The previous implementation would wrap every annotation in a new `Annotated` type without checking if validators were already present, leading to duplicate validators. The fix:
1. Extracts existing metadata from `Annotated` types
2. Checks for existing validators by class type
3. Only adds validators that aren't already present
4. Preserves the base type and other metadata

## Test Plan
- Run `uv run pytest tests/unit/test_registry_repository.py -v`
- Verify validators are attached only once when called multiple times
- Verify module reload fallback works when loader is missing

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/unit/test_registry_repository.py
+++ b/tests/unit/test_registry_repository.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+import importlib
+import importlib.machinery
+import sys
+from types import ModuleType
+from typing import Annotated, ForwardRef, get_args, get_origin
+
+import pytest
+
+from tracecat.expressions.validation import TemplateValidator
+from tracecat.registry.repository import attach_validators, import_and_reload
+
+
+def _metadata(annotation: object) -> list[object]:
+    origin = get_origin(annotation)
+    if origin is Annotated:
+        args = get_args(annotation)
+        return list(args[1:])
+    return []
+
+
+def test_attach_validators_idempotent() -> None:
+    def sample(x: int) -> int:  # noqa: ANN001 - test helper
+        return x
+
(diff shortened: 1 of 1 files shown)


daryllimyt:
### Registry Postmortem Summary
- Reloading the builtin registry tried `importlib.reload` even when the module had no loader, producing `ModuleNotFoundError("spec not found for 'tracecat_registry'")` and cascading submodule import failures.
- Every action bind called `attach_validators` on the same function object; we kept wrapping annotations with a fresh `TemplateValidator`, so repeated retries stacked validators until Pydantic hit `RecursionError`.
- Fixes: guard `import_and_reload` to skip reload without a loader, make validator attachment idempotent, and add unit coverage that locks the behavior down.

daryllimyt:
Also worth noting: this patch removes the root causes behind the intermittent 'Unknown action' 500s and ModuleNotFoundError traces we’ve been seeing in production, so those error reports should clear once it lands.

PR #1558, opened 2025-10-23
feat(core): Add ignore_if_exists option to registry create_table UDF

topher-lo:
## Summary
- add an `ignore_if_exists` flag to the registry `create_table` UDF and forward it to the tables service
- extend the `TableCreate` model and service logic so existing tables can be ignored during creation
- cover the new option with registry unit tests

## Testing
- pytest tests/registry/test_core_table.py::TestCoreCreateTable -q *(fails: ModuleNotFoundError: No module named 'tracecat_registry')*

------
https://chatgpt.com/codex/tasks/task_e_68f97e92b55c8333bb2ae16f2c20a2a9
    

---

## Summary by cubic
Add ignore_if_exists to the registry create_table UDF to skip creation when a table already exists. This makes table creation idempotent and returns the existing table instead of erroring.

- New Features
  - create_table UDF accepts ignore_if_exists (default: false).
  - TableCreate model now includes ignore_if_exists.
  - TablesService.create_table checks for existing table when flag is true; logs and returns it without creating.
  - Unit tests cover flag propagation and behavior.

### DIFF
--- a/tests/unit/test_tables_service.py
+++ b/tests/unit/test_tables_service.py
@@ -3,6 +3,7 @@ from decimal import Decimal
 from uuid import UUID, uuid4
 
 import pytest
+import sqlalchemy as sa
 from sqlalchemy.exc import DBAPIError, StatementError
 from sqlmodel.ext.asyncio.session import AsyncSession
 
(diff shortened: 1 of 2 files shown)


topher-lo:
Poor quality.

PR #1600, opened 2025-10-31
feat: Generate smarter action input defaults

daryllimyt:
We were simply doing the equivalent of `'\n'.join("{key} : " for key in required_keys)` which results in initial values all being `null`, which can fail validaiton leading to jarring UX. By generating smart defaults (zero value) based on the args schema, we avoid initially setting invalid values. 

## Summary
Replaces basic YAML field listing with intelligent default value generation for action inputs, improving UX by providing meaningful zero-values that respect field constraints.

## Changes
- Added `generate_zero_defaults()` utility to generate type-aware defaults for Pydantic models
- Supports primitives, collections, constraints (min_length, gt, ge), nested models, enums, and datetime types
- Updated action router to use generated defaults instead of empty field list
- Added comprehensive unit tests covering 40+ scenarios

## Test plan
- All existing tests pass
- New unit tests verify correct default generation for various field types and constraints

### DIFF
--- a/tracecat/workflow/actions/utils.py
+++ b/tracecat/workflow/actions/utils.py
@@ -21,7 +21,7 @@ from pydantic_core import PydanticUndefined
 _ZERO_SENTINEL = object()
 
 
-def _metadata_lookup(field_info, attr: str):
+def _metadata_lookup(field_info: Any, attr: str) -> Any:
     """Look up metadata attribute from field info."""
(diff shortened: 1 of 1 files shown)


daryllimyt:
@cubic-dev-ai rereview

daryllimyt:
@codex rereview

PR #1614, opened 2025-11-04
feat(core): Add raise_if_duplicate to create table UDF

topher-lo:
## Summary by cubic
Added a raise_on_duplicate option to the create_table UDF to control behavior when a table already exists. By default it raises a ValueError; set to False to return the existing table after a rollback.

- **New Features**
  - Handle DuplicateTableError wrapped in ProgrammingError.
  - Roll back failed transaction, then return existing table via get_table_by_name.

<sup>Written for commit 1bc4bb40d2bb6d65a1044bda3c2c02dba767c292. Summary will update automatically on new commits.</sup>

### DIFF
--- a/tests/registry/test_core_table.py
+++ b/tests/registry/test_core_table.py
@@ -457,10 +457,10 @@ class TestCoreCreateTable:
         assert result == mock_table.model_dump.return_value
 
     @patch("tracecat_registry.core.table.TablesService.with_session")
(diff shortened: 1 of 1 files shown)


topher-lo:
Screenshot 2025-11-03 at 11 08 58 PM
Screenshot 2025-11-03 at 11 08 52 PM

PR #1650, opened 2025-11-14
fix: bigint support for table service

jordan-umusu:
Migrate to BIGINT for integers columns in tables to prevent integer overflow.

---
## Summary by cubic
Switch INTEGER columns to BIGINT in the table service to prevent integer overflow. Applies to column creation, type updates, and SQL bindings.

- **Bug Fixes**
  - Map SqlType.INTEGER to BIGINT in DDL for create and alter column operations.
  - Bind INTEGER values as BigInteger in to_sql_clause to match the DB type.

<sup>Written for commit 977152502cdde7214c96eba8c330663c58486d30. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/tracecat-registry/tracecat_registry/integrations/openai_guardrails.py
+++ b/packages/tracecat-registry/tracecat_registry/integrations/openai_guardrails.py
@@ -2,14 +2,13 @@
 
 from __future__ import annotations
 
-from typing import Annotated, TypedDict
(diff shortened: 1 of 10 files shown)


jordan-umusu:
@cubic review

topher-lo (approved):
Will approve but let's add a one file alemvic migration to migrate all existing int32 table columns to binging

Issue #1676, opened 2025-11-19
Autogenerated OAuth provider ID has unsupported symbols


zisk:
**Describe the bug**
Custom OAuth providers created through the UI generate provider IDs with hyphens (e.g., `custom-secureapp`), which cannot be referenced in workflow template expressions because the expression parser only accepts identifiers with letters, digits, and underscores.

**Provide logs**
Workflow execution fails with:

```
TracecatExpressionError: Error parsing expression `SECRETS.custom-secureapp_oauth.CUSTOM-SECUREAPP_SERVICE_TOKEN`

Failed to parse expression: Unexpected token Token('FN_NAME_WITH_TRANSFORM', 'iven_oauth.CUSTOM') at line 1, column 19.
Expected one of:
    * LPAR
Previous tokens: [Token('TYPE_SPECIFIER', 'str')]
```

**To reproduce**
1. Create a custom OAuth provider via the UI:
   - Name: `SecureApp`
   - Grant Type: `client_credentials`
   - Configure endpoints and credentials

2. Create an action template that uses this provider:
   ```yaml
   type: action
   definition:
     title: Search SecureApp
     namespace: tools.secureapp
     name: search
     secrets:
       - type: oauth
         provider_id: custom-secureapp  # Generated by system
         grant_type: client_credentials
     steps:
       - ref: call_api
         action: core.http_request
         args:
           headers:
             Authorization: Bearer ${{ SECRETS.custom-secureapp_oauth.CUSTOM-SECUREAPP_SERVICE_TOKEN }}
   ```

3. Use the action in a workflow

4. Execute the workflow

**Expected behavior**

The OAuth token should be resolved from the custom provider and used in the HTTP request.

**Environment (please complete the following information):**

- Tracecat version: Current main branch (commit b33cecab as of 2025-11-18)
- OS macOS
- Where did you deploy Tracecat? Docker
- CPU architecture ARM
- Browser type Chrome

**Additional context**

### Provider ID Generation

**Location:** `tracecat/integrations/service.py:88-102`

The `_generate_custom_provider_id()` function uses:
```python
slug = slugify(base_source, separator="-") or uuid4().hex
if not slug.startswith("custom-"):
    slug = f"custom-{slug}"
```

This generates provider IDs like `custom-secureapp` (with hyphens).

### Expression Parser Grammar

**Location:** `tracecat/expressions/parser/grammar.py:80,102,116`

The expression grammar defines:
```lark
[...]

PR #1682, opened 2025-11-20
fix: use underscores in custom OAuth provider IDs and fix overflow

jordan-umusu:
## Summary
Fixed custom OAuth provider IDs to use underscores instead of hyphens (e.g., custom_secureapp vs custom-secureapp) for expression parser compatibility. Also constrained the custom provider dialog height to 85% viewport with scrolling. 

Issue noted in: #1676 
## Description
- Changed slugify() separator from - to _ in provider ID generation to work with expression parser
- Added max-h-[85vh] and overflow-y-auto to custom provider dialog to prevent full-screen height
image

---
## Summary by cubic
Switch custom OAuth provider IDs to underscores for parser compatibility, and constrain the custom provider dialog height with scrolling to prevent overflow. Migrate existing custom- IDs in the database to custom_ to keep data consistent.

- **Bug Fixes**
  - OAuth IDs: Generate custom IDs with underscores (custom_), use underscore suffixes for collisions, support legacy hyphen IDs in detection, and add tests for underscore-based secret paths.
  - UI: Limit the custom provider dialog to 85vh and enable vertical scrolling to avoid content overflow.

- **Migration**
  - Replace custom- with custom_ in oauth_integration and oauth_provider provider_id via Alembic migration; downgrade reverses the change.

<sup>Written for commit f7531fb13bc181867d04a337c1a5f5fafbe10457. Summary will update automatically on new commits.</sup>

### DIFF
--- a/tracecat/integrations/router.py
+++ b/tracecat/integrations/router.py
@@ -484,7 +484,7 @@ async def delete_integration(
 
     # For custom providers, delete the provider definition even if no integration exists
     if integration is None:
-        if provider_info.key.id.startswith(("custom_", "custom-")):
+        if provider_info.key.id.startswith("custom_"):
             # Delete the custom provider definition if it exists
             await svc.delete_custom_provider(provider_key=provider_info.key)
             return
--- a/alembic/versions/b873c83b8e0a_migrate_custom_provider_ids_hyphen_to_.py
+++ b/alembic/versions/b873c83b8e0a_migrate_custom_provider_ids_hyphen_to_.py
@@ -1,0 +1,75 @@
+"""migrate_custom_provider_ids_hyphen_to_underscore
+
+Revision ID: b873c83b8e0a
+Revises: c13c1c2f4d93
+Create Date: 2025-11-26 13:40:46.077313
+
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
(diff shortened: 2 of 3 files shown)


jordan-umusu:
@cubic review

topher-lo:
Closes #1676

tracecat/integrations/router.py
```diff
@@ -484,7 +484,7 @@ async def delete_integration(
 
     # For custom providers, delete the provider definition even if no integration exists
     if integration is None:
-        if provider_info.key.id.startswith("custom-"):
+        if provider_info.key.id.startswith(("custom_", "custom-")):
```
topher-lo: This might be unexpected behavior for the user. Can we make the user manually delete the old integration instead?

topher-lo (changes requested):
One comment then ready to merge.

tracecat/integrations/router.py
jordan-umusu: Semi-reviving this apologies but could you elaborate? This line allows either variation of the provider key that will/currently exists to delete the provider it creates. Otherwise custom providers without an integration using the old provider key prefix "custom-"  cannot be deleted.

tracecat/integrations/router.py
topher-lo, line 1: Revert given new one time migration?

topher-lo (changes requested):

PR #1693, opened 2025-11-21
feat(app+ui): Handle null action results in workflow events panel

daryllimyt:
## Summary
- display a descriptive message when an action result is null in the workflow events sidebar
- reuse the result viewer for both standard and session result tabs to avoid empty output
- handle binary and null payloads in the compression codec decode() path so they surface as None instead of empty panes

## Testing
- pnpm -C frontend lint

## Screens

https://github.com/user-attachments/assets/8eb51c0d-504a-4890-acc8-8842eb593a5d

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6920998857f4832083c156af8dad512a)

---
## Summary by cubic
Show a clear message when a workflow action returns null and use a shared result viewer to keep the events sidebar consistent. Also interpret Temporal binary/null and empty payloads as None to remove empty panes in both Result and Session Result tabs.

- **Bug Fixes**
  - Display “This action returned no result (null).” when action_result is null or undefined across both result views.
  - Treat binary/null encodings and empty payloads as None in payload extraction and compression decoding.

<sup>Written for commit d0a1afef6df484dd4b2695dc31322c9f47204876. Summary will update automatically on new commits.</sup>

### DIFF
--- a/frontend/src/components/builder/events/events-selected-action.tsx
+++ b/frontend/src/components/builder/events/events-selected-action.tsx
@@ -212,15 +212,45 @@ export function SuccessEvent({
       )
     case "result":
       return (
-        <JsonViewWithControls
-          src={event.action_result}
+        <ActionResultViewer
+          result={event.action_result}
+          eventRef={eventRef}
           defaultExpanded={defaultExpanded}
           defaultTab={defaultTab}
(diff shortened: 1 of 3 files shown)


daryllimyt:
@codex review

PR #1711, opened 2025-11-25
fix: agent history trimming/truncating

jordan-umusu:
## Summary
Replaces existing "pruning" logic with two pydantic-ai history processors to manage agent context: one truncates oversized tool outputs, and another trims history to fit within model-specific token limits while preserving tool call/return pairing.

## Description:
- Add truncate_tool_returns_processor to cap large tool outputs at configurable limit (default 80k chars)
- Add trim_history_processor to trim history based on model's context window (e.g., 200k for Claude, 400k for GPT5)
- Clean orphaned tool calls/returns that would cause API errors (e.g., tool_use without tool_result)
- Token counting uses ~4 chars/token heuristic on serialized messages
- Safe cut-point detection ensures trimming doesn't break tool call pairing
- Processors run in sequence: truncate first, then trim

---
## Summary by cubic
Replaces ad-hoc history pruning with two processors that truncate oversized tool outputs and trim message history to model-specific limits, while removing orphaned tool call/return pairs to prevent API errors.

- **Bug Fixes**
  - Clean orphaned tool calls/returns before trimming to avoid invalid histories.
  - Truncate large tool results (strings, dicts, lists) to a configurable limit (default 80k chars).
  - Trim history using token counts and safe cut points; preserves system prompts and valid tool sequencing.
  - Integrated processors into the agent; removed legacy prune_history path.
  - Minor UI fix: trim string defaults in Add Custom Field dialog.

- **Dependencies**
  - Update model context limits and raise default tool output limit to 80k chars.

<sup>Written for commit e84a6a91d7ebf1d043ac505150a2c111b0bc4435. Summary will update automatically on new commits.</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ dependencies = [
     "pyarrow==16.1.0",
     "pydantic-extra-types>=2.10.2,<3.0.0",
     "pydantic>=2.11.7,<3.0.0",
-    "pydantic-ai-slim[openai,anthropic,bedrock,google,mcp]==1.22.0",
+    "pydantic-ai-slim[anthropic,bedrock,google,mcp,openai]==1.22.0",
     "pysaml2==7.5.0",
     "python-slugify==8.0.4",
     "ray[default]==2.43.0",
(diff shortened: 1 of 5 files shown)


jordan-umusu:
@codex review

PR #1712, opened 2025-11-25
fix: swap to JSONB schema validation for case fields to support SELEC…

jordan-umusu:
## Summary
Added a new column "schema"  to case_fields to more flexibly support validation and SELECT/MULTI_SELECT types
## Description
- Add options field to custom field creation for SELECT/MULTI_SELECT types with client-side validation ensuring defaults are within defined options
- Store logical field type and options in CaseFields.schema JSONB column to distinguish SELECT/MULTI_SELECT from their physical TEXT/JSONB storage types

### DIFF
--- a/alembic/versions/3f1bacf5a8a9_jsonb_schema_validation_case_fields.py
+++ b/alembic/versions/3f1bacf5a8a9_jsonb_schema_validation_case_fields.py
@@ -1,7 +1,7 @@
 """jsonb schema validation case fields
 
 Revision ID: 3f1bacf5a8a9
-Revises: c13c1c2f4d93
+Revises: b873c83b8e0a
 Create Date: 2025-11-25 20:33:14.153722
 
 """
@@ -17,7 +17,7 @@ from tracecat.identifiers.workflow import WorkspaceUUID
 
 # revision identifiers, used by Alembic.
 revision: str = "3f1bacf5a8a9"
-down_revision: str | None = "c13c1c2f4d93"
+down_revision: str | None = "b873c83b8e0a"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 


jordan-umusu:
@codex review

jordan-umusu:
@cubic review

tracecat/cases/service.py
```diff
@@ -630,15 +634,15 @@ async def update_case(self, case: Case, params: CaseUpdate) -> Case:
...
+            # Check if workspace row exists for this case
+            existing_fields = await self.fields.get_fields(case)
+            if existing_fields is not None:
                 # Merge existing fields with new fields
-                existing_fields = await self.fields.get_fields(case) or {}
-                await self.fields.update_field_values(
-                    case_fields.id, existing_fields | fields
-                )
+                row_id = await self.fields._ensure_workspace_row(case.id)
+                await self.fields.update_field_values(row_id, existing_fields | fields)
```
daryllimyt: I think update_field_values() should actually take a Case (similar to create_field_values()), then you call _ensure_workspace_row() at the start with the returned row_id. We're leaking the internals of the CaseFieldsService here. Then you can drop the additional _ensure_schema_ready() call inside since it's being called twice

tracecat/cases/service.py
```diff
@@ -702,7 +706,11 @@ async def delete_case(self, case: Case) -> None:
 
 
 class CaseFieldsService(CustomFieldsService):
-    """Service that manages workspace-scoped case fields."""
+    """Service that manages workspace-scoped case fields.
+
+    CaseFields is the workspace-level definition storing schema metadata.
+    The actual per-case values are stored in a dynamic workspace table.
```
daryllimyt:
```suggestion
    The actual per-case values are stored in the workspace's custom fields schema in the `case_fields` table.
```

tracecat/cases/service.py
jordan-umusu: Agreed. In fact the logic here between the two is identical, merged the two to be upsert.

daryllimyt (approved):

PR #1769, opened 2025-12-04
fix(workflow): workflow cursor pagination

jordan-umusu:
## Summary
Use new server-side sorting for workflow pagination


jordan-umusu:
@cubic review

PR #1818, opened 2025-12-18
feat(api+integrations): Support array payloads and update payload UDF

topher-lo:
## Summary by cubic
Enable case payloads to be objects or arrays and add a UDF to append or replace payload data. This supports richer ingestion and safer incremental updates.

- **New Features**
  - Case payload now accepts object or array in API schemas, DB model (JSONB), and frontend viewer.
  - Added update_payload(case_id, payload, append=false): when append=true, dict+dict merges keys, list+list extends, and dict<->list converts and appends.

- **Migration**
  - Regenerate API/client types and update any code that assumed payload is always an object.
  - No backend data migration needed (JSONB unchanged); the UI already renders arrays and hides empty payloads.

<sup>Written for commit d33eaf077da75783f3feb3e332b476fc89a9b004. Summary will update automatically on new commits.</sup>

### DIFF
--- a/frontend/src/client/schemas.gen.ts
+++ b/frontend/src/client/schemas.gen.ts
@@ -3121,6 +3121,10 @@ export const $CaseCreate = {
           type: "object",
         },
         {
+          items: {},
+          type: "array",
+        },
+        {
           type: "null",
         },
       ],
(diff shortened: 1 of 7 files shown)


topher-lo:
Need to redo with new SDK API spec.

PR #1820, opened 2025-12-19
feat(registry): Add typed return types for case UDFs

daryllimyt:
## Summary
- Add SDK types module (`tracecat_registry.sdk.types`) with TypedDict definitions for UDF return types
- Update all case UDFs to use typed returns instead of `dict[str, Any]`
- Update table UDFs to use typed returns (`TableRowDict`, `TableReadDict`)
- Enables better IDE autocomplete and static type checking for SDK consumers

### Deadlock fix (case UDFs only)
- Fixed deadlock in `CasesService.create_case` by ensuring `case_fields` schema is initialized before taking locks on the `case` table
- Added advisory locking to `CustomFieldsService.initialize_workspace_schema` to serialize concurrent DDL operations
- Added test fixture with advisory lock to prevent deadlocks during parallel test execution

**Note:** The deadlock fix specifically addresses concurrent case creation scenarios. The advisory locking in `CustomFieldsService` is triggered by case operations, not general custom field operations.

## Test plan
- [x] Existing characterization tests pass with TypeAdapter validation
- [x] Type checking passes
- [x] Previously flaky deadlock tests now pass consistently

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/tracecat-registry/tracecat_registry/types.py
+++ b/packages/tracecat-registry/tracecat_registry/types.py
@@ -441,14 +441,14 @@ class CaseAttachmentRead(TypedDict):
     """Case attachment metadata."""
 
     id: UUID
-    case_id: str
-    file_id: str
+    case_id: UUID
+    file_id: UUID
     file_name: str
     content_type: str
     size: int
     sha256: str
-    created_at: str
-    updated_at: str
+    created_at: datetime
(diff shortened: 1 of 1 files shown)


daryllimyt:
@cubic review

PR #1846, opened 2025-12-29
feat(agent-harness): Add litellm service and MCP setup

jordan-umusu:
## Summary by cubic
Adds a LiteLLM gateway with per-request credentials and a secure two-tier MCP for scoped tool execution. Action execution now runs via a trusted FastMCP server with nsjail sandboxing and a per-job proxy MCP.

- **New Features**
  - LiteLLM gateway: embedded proxy with custom auth injecting workspace/org credentials at runtime (OpenAI, Anthropic, Bedrock, custom base_url/model). Per-request model and settings come from an executor-minted LLM token; builds response_format for structured outputs from output_type. Credential scope comes from the chat entity and is passed via RunAgentArgs.use_workspace_credentials.
  - Meta-tools MCP: trusted FastMCP server on a Unix socket exposing a generic execute_action tool with nsjail sandboxing, plus a per-job proxy MCP that limits tools, pulls schemas from the registry, forwards calls with separate JWTs for LLM vs MCP, and normalizes tool names for Claude.

- **Bug Fixes**
  - Isolated custom registry installs with --target and automatic path setup; enabled via registry-client feature flag.
  - Sanitized gateway logging and prevented LLM token leakage.

<sup>Written for commit c84282a902a149d257a1b64d2374f455004cb8a0. Summary will update on new commits.</sup>


jordan-umusu:
@cubic review

jordan-umusu:
@cubic re-review

jordan-umusu:
@codex review

jordan-umusu:
@cubic re-review

jordan-umusu:
@cubic rereview

jordan-umusu:
@cubic re-review

tracecat/agent/gateway.py
```diff
@@ -0,0 +1,373 @@
...
+    # Create a service role for the workspace
+    role = Role(
+        type="service",
+        user_id=None,
+        service_id="tracecat-llm-gateway",
+        workspace_id=workspace_uuid,
+    )
+
+    async with get_async_session_context_manager() as session:
+        service = AgentManagementService(session=session, role=role)
```
daryllimyt: nit: prefer .with_session()

tracecat/config.py
```diff
@@ -23,6 +23,12 @@
 )
 
 
+TRACECAT__EXECUTOR_URL = os.environ.get(
+    "TRACECAT__EXECUTOR_URL", "http://executor:8000"
+)
```
daryllimyt: looks unused

tracecat/agent/tokens.py
```diff
@@ -0,0 +1,292 @@
...
+        "sub": MCP_TOKEN_SUBJECT,
+        "iat": int(now.timestamp()),
+        "exp": int((now + timedelta(seconds=ttl)).timestamp()),
+        "role": role.model_dump(mode="json"),
+        "allowed_actions": allowed_actions,
+    }
+    if run_id:
+        payload["run_id"] = run_id
+    if session_id:
+        payload["session_id"] = session_id
```
daryllimyt: shouldn't these be required?

tracecat/agent/tokens.py
```diff
@@ -0,0 +1,292 @@
...
+    """Claims extracted from a verified LLM token.
+
+    These claims are set by the AgentExecutor when minting the token
+    and are immutable - the jailed runtime cannot modify them.
+    """
+
+    # Identity
+    workspace_id: str = Field(..., description="Workspace UUID as string")
+    session_id: str = Field(..., description="Agent session UUID as string")
+
```
daryllimyt: why not UUID over str?

tracecat/agent/mcp/executor.py
```diff
@@ -0,0 +1,242 @@
...
+    # Execute via ActionRunner with nsjail sandbox
+    runner = get_action_runner()
+    result = await runner.execute_action(
+        input=run_input,
+        role=role,
+        resolved_context=resolved_context,
+        tarball_uris=[tarball_uri] if tarball_uri else None,
+        timeout=float(timeout_seconds),
+        force_sandbox=True,
+    )
```
daryllimyt, line 102: can you use the EphemeralBackend instead, unless you need lower level control? It has a higher level API that abstracts a lot of the tarball management. If you need a way to control token minting maybe we can expose a hook

tracecat/agent/mcp/proxy_server.py
```diff
@@ -0,0 +1,152 @@
...
+            tool_name=tool_name,
+        )
+
+    logger.info(
+        "Created proxy MCP server",
+        tool_count=len(tools),
+    )
+
+    return create_sdk_mcp_server(
+        name="tracecat-registry",
```
daryllimyt, line 149: nit: `tracecat_registry` probably more conventional here? so paths would look like `mcp.tracecat_registry...`

tracecat/agent/mcp/utils.py
```diff
@@ -0,0 +1,129 @@
...
+            try:
+                ra = action_map.get(action_name)
+                if ra is None:
+                    logger.warning(
+                        "Action not found in registry",
+                        action_name=action_name,
+                    )
+                    continue
+
+                bound = svc.get_bound(ra, mode="execution")
```
daryllimyt: avoid using `get_bound` as it calls importlib and can cause ModuleNotFound issues, especially if you aren't calling it in a subprocess for execution. You can just query the RegistryAction.interface

tracecat/agent/gateway.py
```diff
@@ -0,0 +1,373 @@
...
+        else:
+            return await service.get_provider_credentials(provider)
+
+
+# -----------------------------------------------------------------------------
+# LiteLLM Auth Hooks
+# -----------------------------------------------------------------------------
+
+
+async def user_api_key_auth(request: Request, api_key: str) -> UserAPIKeyAuth:
```
daryllimyt, line 65: is it fine for this to be async, given that it's mainly doing cpu bound work?

tracecat/agent/gateway.py
```diff
@@ -0,0 +1,373 @@
...
+        self,
+        user_api_key_dict: UserAPIKeyAuth,
+        cache: DualCache,
+        data: dict,
+        call_type: CallTypesLiteral,
+    ):
+        """Inject credentials and model settings before LLM call."""
+        workspace_id: str = user_api_key_dict.metadata.get("workspace_id", "")
+        use_workspace_creds: bool = user_api_key_dict.metadata.get(
+            "use_workspace_credentials", False
```
daryllimyt: is this the correct default?

tracecat/agent/schemas.py
```diff
@@ -44,6 +44,10 @@ class RunAgentArgs(BaseModel):
     """Results for deferred tool calls from a previous run (CE handshake)."""
     is_continuation: bool = False
     """If True, do not emit a new user message; continue prior run with deferred results."""
+    use_workspace_credentials: bool = False
```
daryllimyt: maybe surprising

tracecat/agent/mcp/executor.py
daryllimyt: RunContext = WorkflowRunContext | AgentRunContext

tracecat/agent/mcp/proxy_server.py
jordan-umusu: left as is to differentiate between the mcp server name and the tool that follows which uses snake-case
ref: https://platform.claude.com/docs/en/agent-sdk/custom-tools

tracecat/agent/gateway.py
```diff
@@ -0,0 +1,367 @@
...
+)
+
+
+# -----------------------------------------------------------------------------
+# Credential Fetching via AgentManagementService
+# -----------------------------------------------------------------------------
+
+
+async def get_provider_credentials(
+    workspace_id: str,
```
daryllimyt: can this not be UUID?

tracecat/agent/tokens.py
```diff
@@ -0,0 +1,300 @@
...
+    now = datetime.now(UTC)
+    ttl = ttl_seconds or config.TRACECAT__EXECUTOR_TOKEN_TTL_SECONDS
+
+    payload: dict[str, Any] = {
+        "iss": MCP_TOKEN_ISSUER,
+        "aud": MCP_TOKEN_AUDIENCE,
+        "sub": MCP_TOKEN_SUBJECT,
+        "iat": int(now.timestamp()),
+        "exp": int((now + timedelta(seconds=ttl)).timestamp()),
+        "role": role.model_dump(mode="json"),
```
daryllimyt: just a note: `Role` contains access level etc, which may or may not be sensitive. can't be spoofed as we sign the jwt, but something to keep in mind that might want to be concealed as all fields in the jwt are visible. obviously if we don't include it here we then have to pull this info from the server if required. if not required for this we should omit anything potentially sensitive

jordan-umusu: agreed, the trusted mcp server will instead internalyl construct its role from a passed in workspace id and user id

PR #1891, opened 2026-01-14
feat(cases): inject task_id into TRIGGER context and add custom field editing

topher-lo:
## Summary by cubic
Workflows triggered from case tasks now inject case_id and task_id into TRIGGER inputs with schema-aware handling. Added an Edit Custom Field dialog to update field name, default value, and options for SELECT/MULTI_SELECT.

- **New Features**
  - TRIGGER inputs include case_id and, when present, task_id for no-schema triggers; for schema triggers, these are injected only if defined in the schema.
  - Custom fields table adds an “Edit field” action that opens a dialog to change the identifier, default value (type-aware), and options (with validation).
  - Updates invalidate the case fields list and show a success toast.

<sup>Written for commit 682378dc312661ebe1713c57b73a3824dab17891. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/components/cases/edit-custom-field-dialog.tsx
+++ b/frontend/src/components/cases/edit-custom-field-dialog.tsx
@@ -306,6 +306,13 @@ export function EditCustomFieldDialog({
             type: "manual",
             message: "A field with this name already exists",
           })
+        } else {
+          toast({
+            title: "Error updating field",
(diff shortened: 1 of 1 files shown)


topher-lo:
@cursor review

topher-lo:
@cubic review

PR #1900, opened 2026-01-16
fix(dsl): make gather always return CollectionObject

daryllimyt:
## Problem

`core.transform.gather` is treated as an interface action and the workflow uses a no-op activity to read the scheduler-produced value. Today the scheduler finalizes gather by building a Python list and wrapping it in `InlineObject(data=...)`, which:

- Re-inlines stream results even when they were externalized
- Risks Temporal event payload/history bloat for large scatter→gather workflows
- Keeps gather expression resolution outside of an activity boundary

## Fix

- Evaluate `gather.args.items` per-stream via `DSLActivities.evaluate_templated_object_activity`, storing each stream item as a `StoredObject` handle.
- Add `DSLActivities.finalize_gather_activity` to materialize those handles, apply `drop_nulls` + `error_strategy`, and store the final list as a `CollectionObject` manifest.
- Scheduler now records the gather `TaskResult.result` as that `CollectionObject` (including the skipped-scatter path), while preserving existing error strategy semantics (including fail-fast `RAISE`).

## Tests

- `uv run pytest -q tests/unit/test_workflows.py -k gather`
- `uv run ruff check tracecat/dsl/action.py tracecat/dsl/scheduler.py tests/unit/test_workflows.py`
- `uv run pyright tracecat/dsl/action.py tracecat/dsl/scheduler.py tests/unit/test_workflows.py`

---
## Summary by cubic
Updated core.transform.gather to return a StoredObject-backed collection and finalize it in an activity to cut Temporal history while preserving error handling (including fail-fast RAISE). Uses CollectionObject when externalization is enabled; falls back to InlineObject otherwise.

- **Refactors**
  - Evaluate gather items per stream via evaluate_templated_object_activity and store each as StoredObject.
  - Add finalize_gather_activity to materialize items, apply drop_nulls + error_strategy, and store the final list as a StoredObject (CollectionObject when externalized).
  - Scheduler records TaskResult.result as that StoredObject (including the skipped-scatter path) and raises when errors exist under RAISE.

- **Bug Fixes**
  - Guard CollectionObject creation when externalization is disabled; fall back to InlineObject in scatter input, collection sync, gather finalize, and subflow prep.
  - Add helpers to support both types: scheduler _get_collection_size and PreparedSubflowResult.get_trigger_input_at.

<sup>Written for commit d1e35068ec792ffa151134bbc8638e826dfd6dd2. Summary will update on new commits.</sup>

### DIFF
--- a/tests/unit/test_scatter_interval.py
+++ b/tests/unit/test_scatter_interval.py
@@ -183,8 +183,9 @@ async def test_scatter_without_interval(
     # Verify results
     assert await action_result(result, "gather") == [2, 3, 4, 5]
 
-    # Without interval, execution should be relatively fast (< 4 seconds for simple ops)
-    assert elapsed < 4.0, (
+    # Without interval, execution should be relatively fast
+    # Allow extra time for CI variability and result externalization overhead
+    assert elapsed < 10.0, (
         f"Execution without interval should be fast, but took {elapsed:.2f}s"
     )
 
@@ -255,8 +256,9 @@ async def test_scatter_with_zero_interval(
     # Verify results
     assert await action_result(result, "gather") == [2, 4, 6]
 
-    # With zero interval, should execute quickly (allow overhead for externalization)
-    assert elapsed < 4.0
+    # With zero interval, should execute quickly
+    # Allow extra time for CI variability and result externalization overhead
+    assert elapsed < 10.0
 
 
 @pytest.mark.anyio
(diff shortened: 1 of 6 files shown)


daryllimyt:
### Code review

Found 1 issue:

1. Dead code: Functions `_partition_errors`, `_is_error_info`, and `_as_error_info` are defined in `scheduler.py` but no longer called anywhere. These functions were moved to `action.py` (where they are actively used in `finalize_gather_activity`) but the old definitions were left behind.

https://github.com/TracecatHQ/tracecat/blob/7cacfedb6dcf475bb886594dd93153b009845da4/tracecat/dsl/scheduler.py#L1323-L1351

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1913, opened 2026-01-19
feat(registry): use RegistryIndex for UI action queries and fix template validation

daryllimyt:
## Summary
- Always use V2 sync flow, remove `REGISTRY_SYNC_V2` feature flag checks
- Stop populating `RegistryAction` table during sync (becomes ghost table)
- Query actions from `RegistryIndex` instead of `RegistryAction` for UI
- Fix template action validation in production executor path

### Registry Index Migration
- Add `list_actions_from_index()` and `get_action_from_index()` service methods
- Add `from_index()` and `from_index_and_manifest()` schema factory methods
- Update router endpoints to use `RegistryIndex` as source of truth
- Fetch implementation details from `RegistryVersion.manifest` on-demand

### Template Action Validation Fix
- Add validation of template action args against `expects` schema in `_execute_template_action()`
- This applies defaults and validates types (including enums) - previously missing in the production path
- Fix `ExpectedField` serialization bug where "no default" was serialized as `default: None`
- Use sentinel value `__TRACECAT_UNSET__` to distinguish required fields from optional fields with `None` default

## Why the validation fix was needed
The production executor path (`_execute_template_action`) was missing validation that existed in the deprecated path (`run_template_action`). Additionally, when `ExpectedField` objects were serialized through the manifest (JSON), the distinction between "no default specified" (required) and "default=None" (optional) was lost because both serialized to `{"default": null}`.

## Test plan
- [x] Verify registry sync creates `RegistryIndex` entries
- [x] Verify `/registry/actions` returns data from index
- [x] Verify `/registry/actions/{name}` returns full details with implementation
- [x] Verify template actions apply defaults correctly
- [x] Verify template actions validate required fields
- [x] Verify template actions validate enum values

### DIFF
--- a/tracecat/expressions/schemas.py
+++ b/tracecat/expressions/schemas.py
@@ -26,6 +26,10 @@ class ExpectedField(BaseModel):
     type: str
     description: str | None = None
     default: Any = _UNSET_SENTINEL
+    enum: list[str] | None = None
+    """Optional list of allowed values for this field."""
+    optional: bool | None = None
+    """Whether this field is optional (alternative to using default)."""
 
     def has_default(self) -> bool:
         """Check if a default value was explicitly specified."""


daryllimyt:
This was merged into fix/platform-registry-sync

PR #1914, opened 2026-01-20
build(deploy): migrate to dedicated migrations container

daryllimyt:
## Summary
- Replace `RUN_MIGRATIONS=true` on API startup with a dedicated migrations service/job
- Migrations run before other services start, ensuring DB schema is ready

## Changes

**Docker Compose:**
- Add `migrations` service running `alembic upgrade head`
- Update `depends_on` chains: `db → migrations → api/worker/executor`

**Helm:**
- Add `migrations-job.yaml` as pre-install/pre-upgrade hook
- ServiceAccount created as hook (weight -10) before migrations (weight -5)
- Remove `RUN_MIGRATIONS` from API env vars

## Test plan
- [x] Helm template renders correctly
- [x] QA'd migrations hook on OrbStack K8s cluster
- [x] Verified migrations complete before main deployments start

---
## Summary by cubic
Move database migrations to a dedicated container/job that runs before other services. This ensures the DB schema is ready and removes startup race conditions across Docker Compose and Helm.

- **Refactors**
  - Docker Compose: added a migrations service (alembic upgrade head) and gated api/worker/executor on its successful completion.
  - Helm: added a post-install/post-upgrade migrations Job with a wait-for-postgres init container.
  - Removed RUN_MIGRATIONS from API configs in Compose and Helm.

- **Migration**
  - Helm: no action needed; the migrations job runs automatically on install/upgrade.
  - Compose: stop setting RUN_MIGRATIONS and rely on the migrations service; ensure TRACECAT__DB_URI and TRACECAT__DB_SSLMODE are set.
  - Custom stacks: make app services depend on the migrations service_completed_successfully.

<sup>Written for commit 1b241dae1a74d21cd8fa8160aa6036fcfa19b021. Summary will update on new commits.</sup>

### DIFF
--- a/deployments/helm/tracecat/templates/migrations-job.yaml
+++ b/deployments/helm/tracecat/templates/migrations-job.yaml
@@ -6,7 +6,7 @@ metadata:
     {{- include "tracecat.labels" . | nindent 4 }}
     app.kubernetes.io/component: migrations
   annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook": post-install,post-upgrade
     "helm.sh/hook-weight": "-5"
     "helm.sh/hook-delete-policy": before-hook-creation
 spec:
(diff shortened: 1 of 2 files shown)


daryllimyt:
Addressed in 1b241dae1 - changed to `post-install,post-upgrade` hooks with an init container that waits for Postgres connectivity before running migrations. This ensures the job works with both internal CloudNativePG (default) and external Postgres configurations.

PR #1929, opened 2026-01-21
fix(ui): block input mid agent stream

jordan-umusu:
## Summary by cubic
Block typing while the agent is streaming or processing to prevent mid-response edits. Applies to both chat and Copilot panes.

- **Bug Fixes**
  - Disable input unless status is "ready" or "error" (allows retry after failures).
  - Centralize submit gating with canSubmit and sync submit button with input state.

<sup>Written for commit 78d3cbc2bf42478d9d9c94119e52b9c6db9613b0. Summary will update on new commits.</sup>


daryllimyt (approved):
lgtm though do note that we might need to enforce this server-side too

PR #1956, opened 2026-01-24
fix(tests): stabilize async DB engine lifecycle in tests

daryllimyt:
## Summary

Fixes "Event loop is closed" errors in CI by properly managing the async database engine lifecycle during tests.

**Problem**: The async engine connection pool holds references to the event loop where connections were created. With pytest-xdist parallel testing and pytest-anyio, different tests may run on different event loops. When SQLAlchemy tries to terminate connections created on a closed loop, it fails.

**Solution**:
- Add `reset_async_engine()` function to clear the global engine singleton
- Make `test_db_engine` fixture async and dispose + reset the engine after each test
- This ensures each test gets a fresh engine with connections on the current event loop

## Changes

- `tracecat/db/engine.py`: Add `reset_async_engine()` to reset the global `_async_engine`
- `tests/conftest.py`: Update `test_db_engine` fixture to properly dispose and reset engine per test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,7 +31,11 @@ from tests.database import TEST_DB_CONFIG
 from tracecat import config
 from tracecat.auth.types import AccessLevel, Role, system_role
 from tracecat.contexts import ctx_role
-from tracecat.db.engine import get_async_engine, get_async_session_context_manager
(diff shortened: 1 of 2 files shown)


daryllimyt:
Superseded by #1958 which contains only the test fix changes

Issue #2026, opened 2026-02-04
Incorrect Credentials Input Form on Google Suites-related Integrations


ItsHarta:
**Describe the bug**
Currently, Tracecat provides several built-in Google Suites integrations (e.g., Google Docs, Google Sheets). These integrations requires a Service Accounts credentials, which Google provides in JSON key file format.

However, currently the input form in Tracecat UI incorrectly renders OAuth credentials form, which accepts a Client ID and Client Secret instead of a Service Account `.json` key file.

After a quick investigation, this happened because the Service Account check `IsServiceAccountProvider` incorrectly performs strict checking `id === "google"`. All of the additional Google Suites integration uses a `google_[servicename]` prefix instead.
- Google Docs: `id: ClassVar[str] = "google_docs"`
- Google Sheets: `id: ClassVar[str] = "google_sheets"`

Meanwhile, the backend performed the correct checks and asks for JSON format since it was inherited from the `service_account.py`/**`Google Cloud (service account)`** integrations.

This should be a quick fix. I propose to update the validations against a set of string array instead.
Relevant files provided as Additional Context Below.

**Provide logs**
None
(Reproducible via UI)

**To reproduce**

1. Click **`Integrations`** sidebar.
2. From the available list, select one of Google Suites-related integrations
  2.1. In this case, i will use **`Google Sheets (Service account)`**.
4. Navigate to **`Configuration`** tab.
5. Notice that the **`Client Credentials`** Section asks for a **OAuth Credentials** instead of **Service Account Credentials**.
6. Paste a Google Service Account JSON. You will notice an error:
  `Client secret must be 512 characters or less`

**Expected behavior**
Google Suites-related integrations should render a Service Account Form, which is similar to **Client Credentials** section in **`Google Cloud (service account)`** integration.

**Screenshots**
Currently, the **`Client Credentials`** Form in **`Google Sheets (Service account)`** looks like this:

<img width="782" height="782" alt="Image" src="https://github.com/user-attachments/assets/a79799a3-23ac-4cdd-8867-4d478d13d80b" />

<img width="791" height="939" alt="Image" src="https://github.com/user-attachments/assets/7858ecb6-6f3d-462d-937a-860d5d3a140a" />

  
The correct input form should looks like this (Taken from `Google Cloud (service account)`:

<img width="779" height="659" alt="Image" src="https://github.com/user-attachments/assets/a2a4648d-8126-44ed-aea0-07d3d8d3480b" />

[...]

PR #2027, opened 2026-02-04
fix(integrations): Incorrect Credentials on Google Suites Integrations

ItsHarta:
## Checklist

## Description

Currently, Tracecat provides several built-in Google Suites integrations (e.g., Google Docs, Google Sheets). These integrations requires a Service Accounts credentials, which Google provides in JSON key file format.

However, currently the input form in Tracecat UI incorrectly renders OAuth credentials form, which accepts a Client ID and Client Secret instead of a Service Account .json key file.

## Related Issues

Fixes https://github.com/TracecatHQ/tracecat/issues/2026

## Steps to QA

1. Click Integrations sidebar.
2. From the available list, select one of Google Suites-related integrations
  2.1. In this case, i will use Google Sheets (Service account).
3. Navigate to Configuration tab.
4. Verify that the Client Credentials Section asks for a Service Account Credentials.

---
## Summary by cubic
Fixes the credentials form for Google Suites integrations to show Service Account fields instead of OAuth client fields. Google Docs and Sheets now render the correct inputs and validation.

- **Bug Fixes**
  - Treat "google", "google_sheets", and "google_docs" as service-account providers.
  - Show "Service account email" and use the larger secret max length (16384) for service accounts.

<sup>Written for commit 12117a50d6a71d34788350750ccaa0ce5a224c18. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/components/integrations/oauth-integration-details-dialog.tsx
+++ b/frontend/src/components/integrations/oauth-integration-details-dialog.tsx
@@ -52,7 +52,8 @@ export function OAuthIntegrationDetailsDialog({
     grantType,
   })
 
-  const isServiceAccountProvider = provider?.metadata.id === "google"
+  const serviceAccountProviders = ["google", "google_sheets", "google_docs"]
+  const isServiceAccountProvider = serviceAccountProviders.includes(provider?.metadata.id ?? "")
   const clientIdLabel = isServiceAccountProvider
     ? "Service account email"
     : "Client ID"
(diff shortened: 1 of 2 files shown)


topher-lo:
@ItsHarta LGTM. @jordan-umusu this logic shouldn't be hardcoded in the frontend. Let's fix this too so this doesn't happen again. Same idea for PK OAuth which requires an upload as well

ItsHarta:
> @ItsHarta LGTM. @jordan-umusu this logic shouldn't be hardcoded in the frontend. Let's fix this too so this doesn't happen again. Same idea for PK OAuth which requires an upload as well

mentioned this in the issue as well, would probably require an additional property that defines creds type (which is a major change and will break existing integrations). Since the 1.0-beta is close, I would recommend refactoring after beta release

PR #2040, opened 2026-02-06
fix(agent): AgentOutput v2

jordan-umusu:
## Summary by cubic
Moves structured output handling into the agent runtime/SDK and passes structured outputs and usage end-to-end. The gateway no longer injects response_format.

- **Refactors**
  - Add lightweight common/output_format helper and use it in the Claude runtime to build output_format from AgentConfig.output_type (supports primitives and JSON schema; safe for sandbox).
  - Filter SDK-internal lines like "(no content)" and "Stop hook feedback" from message history; set options.output_format when starting the agent.
  - Extend protocol/socket with result_structured_output; loopback/activity plumb structured_output, usage, and num_turns; durable path returns them in AgentOutput and populates usage fields.
  - Remove gateway response_format injection and output_type token claim; delete helper logic. Add output_type to SandboxAgentConfig and sandbox payloads; update types/parsers to accept str|dict.

- **Migration**
  - If you set output_type via LLM token/gateway before, set AgentConfig.output_type instead. Existing tokens with output_type are ignored; no gateway API changes.

<sup>Written for commit 6adaf968a69c2cba611cd75078783a000f7fd4bc. Summary will update on new commits.</sup>

### DIFF
--- a/tracecat/agent/common/output_format.py
+++ b/tracecat/agent/common/output_format.py
@@ -1,0 +1,72 @@
+"""Lightweight output format helpers for the Claude SDK runtime.
+
+This module has NO pydantic-ai dependencies and can be safely imported
+in sandboxed runtimes with minimal import footprint.
+"""
+
+from __future__ import annotations
+
+from typing import Any, NotRequired, TypedDict
+
+from tracecat.logger import logger
+
+
+class _ItemsSchema(TypedDict):
+    type: str
+
+
+class _JsonSchema(TypedDict):
+    type: str
+    items: NotRequired[_ItemsSchema]
+
+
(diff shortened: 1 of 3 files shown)


tracecat/agent/common/output_format.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from tracecat.logger import logger
+
+# JSON Schema mappings for primitive output types
+_PRIMITIVE_JSON_SCHEMAS: dict[str, dict[str, Any]] = {
```
daryllimyt: nit: would be nice if this was TypedDict

PR #2073, opened 2026-02-11
feat(org): add safe organization deletion with confirmation

daryllimyt:
## Summary by cubic
Adds safe organization deletion that requires typing the exact org name. Restricts self-deletion to org owners, logs the action (with organization resource type), and fully cleans up related resources.

- **New Features**
  - Frontend: admin delete modal requires exact org name; Delete disabled until it matches; sends confirmation.
  - APIs: admin and /organization DELETE accept confirm; return 400 on mismatch; /organization requires owner role and returns 403 for non-owners.
  - Backend: centralized confirmation validation; audit logs organization delete events (adds organization resource type); cleanup removes workspaces, schedules, case fields, memberships, ownerships, secrets, and registry data before deleting the org.
  - Admin CLI/client: CLI prompts for the org name (or uses it with --force); Python client delete_organization now requires a confirmation.

- **Migration**
  - Always send confirm that exactly matches the org name when deleting.
  - Update integrations to use adminDeleteOrganizationWithConfirmation({ orgId, confirm }) and client.delete_organization(org_id, confirmation=...).
  - Use an owner role to call /organization DELETE; platform admins cannot delete organizations via this endpoint.

<sup>Written for commit 90be0cdb0a4b25c0058534c758c9ed03a73dac33. Summary will update on new commits.</sup>

### DIFF
--- a/tracecat/organization/management.py
+++ b/tracecat/organization/management.py
@@ -187,7 +187,9 @@ async def delete_organization_with_cleanup(
             workspace_role=WorkspaceRole.ADMIN,
             access_level=AccessLevel.ADMIN,
         )
-        schedule_service = WorkflowSchedulesService(session=session, role=bootstrap_role)
+        schedule_service = WorkflowSchedulesService(
+            session=session, role=bootstrap_role
+        )
         for schedule in await schedule_service.list_schedules():
             await schedule_service.delete_schedule(schedule.id, commit=False)
 
@@ -205,7 +207,9 @@ async def delete_organization_with_cleanup(
             delete(Ownership).where(Ownership.resource_id.in_(workspace_resource_ids))
         )
 
-    await session.execute(delete(Ownership).where(Ownership.owner_id == organization.id))
+    await session.execute(
+        delete(Ownership).where(Ownership.owner_id == organization.id)
+    )
     await session.execute(
         delete(OrganizationSecret).where(
             OrganizationSecret.organization_id == organization.id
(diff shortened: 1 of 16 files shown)


tracecat/organization/service.py
```diff
@@ -324,6 +329,25 @@ async def add_member(
...
+            raise TracecatNotFoundError("Organization not found")
+
+        validate_organization_delete_confirmation(
+            organization, confirmation=confirmation
+        )
+        await delete_organization_with_cleanup(
+            self.session,
+            organization=organization,
+            operator_user_id=self.role.user_id,
+        )
```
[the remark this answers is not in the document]
daryllimyt: this should be fine as superusers can create a new org from the admin panel

tracecat/organization/management.py
```diff
@@ -134,6 +148,94 @@ async def create_organization_with_defaults(
     return org
 
 
+def validate_organization_delete_confirmation(
```
jordan-umusu, line 151: do we need this? this sort of validation is usually only in the frontend.

tracecat/organization/service.py
```diff
@@ -324,6 +329,26 @@ async def add_member(
         await self.session.refresh(membership)
         return membership
 
+    @audit_log(resource_type="organization", action="delete")
```
jordan-umusu, line 332: this type doesnt exist as an AuditResourceType atm

jordan-umusu (changes requested):

tracecat/organization/management.py
daryllimyt: Client-side confirmations are just UX. The **backend must enforce** the destructive guardrails because anything in the browser can be bypassed (curl, scripts, replayed requests, compromised session, buggy UI).

jordan-umusu (approved):

PR #2081, opened 2026-02-11
fix(expressions): short-circuit ternary branch evaluation

daryllimyt:
### Motivation
- Ternary expressions (``a if cond else b``) were evaluating both branches eagerly, which caused errors when the unselected branch raised (e.g. ``1/0 if False else 1``). 
- The intent is to evaluate only the selected branch after the condition is resolved to match expected short-circuit behavior and avoid side-effects from the unselected branch.

### Description
- Override the tree transformer to handle `ternary` nodes lazily by evaluating the condition first and then only the selected branch in `tracecat/expressions/parser/evaluator.py` via a new `_transform_tree` implementation. 
- Add a helper `_transform_child` to dispatch evaluation for child `Tree` or `Token` nodes. 
- Keep the existing `ternary` visitor signature intact for backward compatibility while ensuring the AST-level short-circuit is applied. 
- Add regression cases to `tests/unit/test_expressions.py` covering ``1 if True else 1/0`` and ``1/0 if False else 1`` to assert the unselected branch is not evaluated.

### Testing
- Ran `uv run ruff check tracecat/expressions/parser/evaluator.py tests/unit/test_expressions.py`, which passed. 
- Ran `uv run ruff format --check tracecat/expressions/parser/evaluator.py tests/unit/test_expressions.py`, which passed. 
- Executed a smoke script with `uv run python - <<'PY'` that evaluated example expressions and confirmed expected short-circuit behavior (e.g. ``1 if True else 1/0`` => `1`, ``1/0 if False else 1`` => `1`). 
- Attempted to run the targeted pytest for the expressions file, but the test run could not complete because the environment could not connect to PostgreSQL at `localhost:5432`, causing test setup failures; full test-suite validation should be run in CI or a dev cluster where the DB is available (see project `AGENTS.md`/README for test environment instructions).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_698cf42b8e8c8320871408cb93e187e1)

---
## Summary by cubic
Short-circuits ternary expressions so only the chosen branch is evaluated, preventing errors from the unselected branch (e.g., 1/0 if False else 1 now returns 1 safely).

- **Bug Fixes**
  - Handle ternary nodes lazily in _transform_tree: evaluate condition first, then only the selected branch.
  - Add _transform_child for Tree/Token dispatch; keep the existing ternary visitor signature.
  - Add regression tests for short-circuit cases (1 if True else 1/0, 1/0 if False else 1).

<sup>Written for commit 33834e2af436e0664a44066032735853864c5f2f. Summary will update on new commits.</sup>


jordan-umusu (approved):
LGTM

PR #2085, opened 2026-02-12
feat(invitations): workspace invitations v2

jordan-umusu:
---
## Summary by cubic
Unifies org and workspace invitations behind a single token-based InvitationService with new /invitations endpoints, a shared accept page, and copyable invite links. Workspace membership now uses a unified POST /workspaces/{id}/members and member lists include invited rows with status and expiry.

- **New Features**
  - Mounted unified /invitations router; added create/list/get-by-token/get-token/accept/revoke and list-my-pending endpoints; shared accept page now uses the unified hooks (use-invitations).
  - Workspace: server-routed POST /members; WorkspaceMember adds a status enum and supports invite-only rows (user_id nullable); UI can copy/revoke invite links; org invites support workspace_assignments; OrgMemberRead exposes token and API returns a member’s workspace memberships.

- **Bug Fixes**
  - Immediate permission updates by clearing in-process authz caches after role/membership changes, including group role assignments; client also invalidates workspaces and user-scopes.
  - Guarded all assignee selectors against invite-only rows and improved the workspace landing flow to prevent duplicate creation and handle loading/fetching states.

<sup>Written for commit a3f7a084c21ee5a291c1bed40ad13c9b06886f7f. Summary will update on new commits.</sup>

image
image

### DIFF
--- a/frontend/src/components/organization/org-members-table.tsx
+++ b/frontend/src/components/organization/org-members-table.tsx
@@ -7,7 +7,7 @@ import { FolderIcon, GlobeIcon, Trash2Icon, XIcon } from "lucide-react"
 import { useState } from "react"
 import { useFieldArray, useForm } from "react-hook-form"
 import { z } from "zod"
-import type { OrgMemberRead } from "@/client"
+import { invitationsGetInvitationToken, type OrgMemberRead } from "@/client"
 import { useScopeCheck } from "@/components/auth/scope-guard"
 import {
   DataTable,
@@ -501,9 +501,13 @@ export function OrgMembersTable() {
                               <>
                                 <DropdownMenuItem
                                   onSelect={async () => {
-                                    if (!member.token) return
+                                    if (!member.invitation_id) return
                                     try {
-                                      const url = `${window.location.origin}/invitations/accept?token=${member.token}`
+                                      const { token } =
+                                        await invitationsGetInvitationToken({
+                                          invitationId: member.invitation_id,
+                                        })
+                                      const url = `${window.location.origin}/invitations/accept?token=${token}`
                                       await navigator.clipboard.writeText(url)
                                       toast({
                                         title: "Copied",
--- a/frontend/src/components/workspaces/workspace-members-table.tsx
+++ b/frontend/src/components/workspaces/workspace-members-table.tsx
@@ -4,7 +4,11 @@ import { DotsHorizontalIcon } from "@radix-ui/react-icons"
 import { useQueryClient } from "@tanstack/react-query"
 import { Check, Copy } from "lucide-react"
 import { useCallback, useEffect, useMemo, useState } from "react"
-import type { WorkspaceMember, WorkspaceRead } from "@/client"
+import {
+  invitationsGetInvitationToken,
+  type WorkspaceMember,
+  type WorkspaceRead,
+} from "@/client"
 import { useScopeCheck } from "@/components/auth/scope-guard"
 import {
   DataTable,
@@ -60,6 +64,7 @@ export function WorkspaceMembersTable({
   workspace: WorkspaceRead
 }) {
   const queryClient = useQueryClient()
+  const canInviteMembers = useScopeCheck("workspace:member:invite")
   const canManageMembers = useScopeCheck("workspace:member:update")
   const canRemoveMembers = useScopeCheck("workspace:member:remove")
   const [selectedMember, setSelectedMember] = useState<WorkspaceMember | null>(
@@ -85,11 +90,20 @@ export function WorkspaceMembersTable({
 
   const [copied, setCopied] = useState<string | null>(null)
 
-  async function handleCopyInviteLink(token: string) {
-    const link = `${window.location.origin}/invitations/accept?token=${token}`
-    await navigator.clipboard.writeText(link)
-    setCopied(token)
-    setTimeout(() => setCopied(null), 2000)
+  async function handleCopyInviteLink(invitationId: string) {
+    try {
+      const { token } = await invitationsGetInvitationToken({ invitationId })
+      const link = `${window.location.origin}/invitations/accept?token=${token}`
+      await navigator.clipboard.writeText(link)
+      setCopied(invitationId)
+      setTimeout(() => setCopied(null), 2000)
+    } catch {
+      toast({
+        title: "Failed to copy invitation link",
+        description: "Could not retrieve the invitation token.",
+        variant: "destructive",
+      })
+    }
   }
 
   const handleChangeRole = useCallback(
@@ -265,13 +279,15 @@ export function WorkspaceMembersTable({
                     <DropdownMenuContent>
                       {isInvited ? (
                         <>
-                          {member.token && (
+                          {canInviteMembers && member.invitation_id && (
                             <DropdownMenuItem
                               onClick={() =>
-                                handleCopyInviteLink(member.token as string)
+                                handleCopyInviteLink(
+                                  member.invitation_id as string
+                                )
                               }
                             >
-                              {copied === member.token ? (
+                              {copied === member.invitation_id ? (
                                 <Check className="mr-2 size-4" />
                               ) : (
                                 <Copy className="mr-2 size-4" />
@@ -279,7 +295,7 @@ export function WorkspaceMembersTable({
                               Copy invitation link
                             </DropdownMenuItem>
                           )}
-                          {canRemoveMembers && (
+                          {canInviteMembers && (
                             <AlertDialogTrigger asChild>
                               <DropdownMenuItem
                                 className="text-rose-500 focus:text-rose-600"
--- a/tracecat/invitations/router.py
+++ b/tracecat/invitations/router.py
@@ -62,20 +62,19 @@ async def _check_scope(
     required_scope: str,
     workspace_id: WorkspaceID | None = None,
 ) -> None:
-    """Verify the user has the required scope, checking org then workspace scopes.
+    """Verify the user has the required scope, with workspace-context checks.
 
-    1. Check org-level scopes already on the role (free).
-    2. For workspace ops, compute workspace-specific scopes (cached 30s TTL).
+    For org-level operations, checks the role's resolved scopes.
+    For workspace operations, always computes effective scopes in the workspace
+    context, which includes both org-wide and workspace-specific assignments.
 
     Raises:
         HTTPException 403 if the user lacks the required scope.
     """
-    # Org-level scopes are already resolved on the role
-    if role.scopes and has_scope(role.scopes, required_scope):
-        return
-
-    # For workspace operations, compute workspace-specific scopes
-    if workspace_id is not None:
+    if workspace_id is None:
+        if role.scopes and has_scope(role.scopes, required_scope):
+            return
+    else:
         ws_role = role.model_copy(update={"workspace_id": workspace_id})
         ws_scopes = await compute_effective_scopes(ws_role)
         if has_scope(ws_scopes, required_scope):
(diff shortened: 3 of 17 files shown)


jordan-umusu:
@cubic review

tracecat/workspaces/service.py
```diff
@@ -39,6 +42,175 @@
...
+            user_id=user_id,
+            workspace_id=invitation.workspace_id,
+            role=invitation.role,
+        )
+        session.add(membership)
+
+        await session.commit()
+        await session.refresh(membership)
+    except TracecatValidationError:
+        # Re-raise validation errors without logging as failure (expected user errors)
```
daryllimyt: should this not rollback?

frontend/src/components/organization/org-members-table.tsx
```diff
@@ -528,31 +771,122 @@ function ChangeUserRoleDialog({
...
+  useEffect(() => {
+    setWsRoleChanges({})
+  }, [selectedUser?.user_id])
+
+  const handleConfirm = useCallback(async () => {
+    if (!selectedUser?.user_id) return
+    setIsSaving(true)
+    try {
+      // Update org role
+      onConfirm(newRole)
```
daryllimyt: High: `onConfirm(newRole)` is async but not awaited here. If the org-role update fails, this `try/catch` won’t reliably catch it, and workspace-role updates can still run, leaving inconsistent state. Recommend `await onConfirm(newRole)` at minimum, and ideally move org+workspace role changes into one backend transaction.

frontend/src/components/organization/org-members-table.tsx
```diff
@@ -514,10 +647,120 @@ export function OrgMembersTable() {
...
+  const [isPending, setIsPending] = useState(false)
+  const { workspaces } = useWorkspaceManager()
+
+  const handleSubmit = useCallback(async () => {
+    if (!member?.email || !workspaceId) return
+    setIsPending(true)
+    try {
+      await workspacesCreateWorkspaceInvitation({
+        workspaceId,
+        requestBody: { email: member.email, role: wsRole },
```
daryllimyt: High: assigning an existing org member to a workspace is currently email-based from the client. This should be a backend endpoint keyed by `user_id` so identity and assignment policy (direct membership vs invitation) are enforced server-side and idempotently.

frontend/src/components/workspaces/add-workspace-member.tsx
```diff
@@ -64,126 +70,195 @@ export function AddWorkspaceMember({
...
-        })
-      }
-      return
-    }
 
-    // We've found this user
-    try {
-      await addWorkspaceMember({
+      // Get the invitation token to build the magic link
+      const { token } = await workspacesGetInvitationToken({
```
daryllimyt: Medium: this is a two-step client orchestration (create invitation, then fetch token, then construct URL). It introduces partial-failure modes and duplicates URL assembly in the UI. Prefer returning a canonical shareable accept URL (or token) directly from invitation creation.

frontend/src/components/workspaces/workspace-members-table.tsx
```diff
@@ -50,69 +48,140 @@ import {
...
   )
+  const {
+    invitations,
+    isLoading: invitationsLoading,
+    error: invitationsError,
+    revokeInvitation,
+  } = useWorkspaceInvitations(workspace.id, { enabled: user?.isPrivileged() })
+
+  // Combine members and pending invitations into a single list
+  const combinedData = useMemo<MemberOrInvitation[]>(() => {
```
daryllimyt: Medium: members and pending invitations are merged client-side from separate queries. This should be a unified backend list endpoint so sorting/filtering/pagination happen on one authoritative dataset and one consistency snapshot.

frontend/src/components/organization/org-members-table.tsx
```diff
@@ -528,31 +771,122 @@ function ChangeUserRoleDialog({
...
+
+  const handleConfirm = useCallback(async () => {
+    if (!selectedUser?.user_id) return
+    setIsSaving(true)
+    try {
+      // Update org role
+      onConfirm(newRole)
+
+      // Update changed workspace roles
+      const updates = Object.entries(wsRoleChanges)
```
daryllimyt: High: this client-side fan-out (`PATCH` org role + N workspace membership updates) is non-atomic and can leave partial updates on mid-loop failure. This should be a single backend endpoint/mutation that applies all role changes in one transaction.

frontend/src/components/auth/sign-up.tsx
```diff
@@ -176,7 +287,48 @@ export function BasicRegistrationForm({
           password: values.password,
         },
       })
-      // Redirect is handled by the useEffect in SignUp component
+
+      // After successful login, check if this is an invitation flow
+      const invitation = extractInvitation(returnUrl ?? null)
```
daryllimyt: High: invitation acceptance is being orchestrated in the client after sign-up (parse return URL, preflight lookup, then accept). This policy/transaction logic should live server-side so sign-up + invitation acceptance are atomic and race-free for both org and workspace invitations.

frontend/src/components/workspaces/workspace-members-table.tsx
```diff
@@ -50,69 +48,140 @@ import {
...
+  first_name: string | null
+  last_name: string | null
+  workspace_role: WorkspaceRole
+  status: "active" | "pending"
+  // Original data for actions
+  _type: "member" | "invitation"
+  _original:
+    | WorkspaceMember
+    | { id: string; email: string; role: WorkspaceRole }
+}
```
daryllimyt: type should be derived from backend

tracecat/organization/router.py
```diff
@@ -374,6 +375,30 @@ async def update_org_member(
         ) from e
 
 
+@router.get(
+    "/members/{user_id}/workspace-memberships",
+    response_model=list[UserWorkspaceMembership],
+)
+async def list_member_workspace_memberships(
```
daryllimyt, line 399: Architecture: this route exists mainly to support client-side role orchestration (read memberships, diff in UI, then fan-out writes). If we move role updates server-side (recommended), consider consolidating to a single member-roles API instead: `GET /organization/members/{user_id}/roles` + atomic `PATCH /organization/members/{user_id}/roles` returning `{ org_role, workspace_roles[] }`. That would make this endpoint redundant, reduce API surface area, and eliminate client-side consistency/race issues.

tracecat/organization/service.py
```diff
@@ -483,9 +565,90 @@ async def create_invitation(
...
+        Validates that each workspace belongs to this organization.
+        Skips workspaces that already have a pending, unexpired invitation for this email.
+        Replaces expired/revoked invitations.
+
+        Must be called within an existing transaction (before commit).
+
+        Returns:
+            Number of workspace invitations created.
+        """
+        assert self.role is not None and self.role.user_id is not None
```
daryllimyt: nit: throw with friendly exception

tracecat/workspaces/router.py
```diff
@@ -474,3 +486,141 @@ async def revoke_workspace_invitation(
...
+    session: AsyncDBSession,
+    params: WorkspaceInvitationAccept,
+) -> dict[str, str]:
+    """Accept an invitation and join the workspace.
+
+    This endpoint doesn't require workspace context since the user
+    may not belong to the workspace yet. Uses AuthenticatedUserOnly
+    which only requires an authenticated user.
+    """
+    assert role.user_id is not None
```
daryllimyt: nit: prefer to raise friendly exception instead of  plain assert

tracecat/workspaces/service.py
```diff
@@ -39,6 +42,175 @@
...
+        type="user",
+        user_id=user_id,
+        organization_id=organization_id,
+        workspace_id=invitation.workspace_id,
+        access_level=AccessLevel.BASIC,
+        service_id="tracecat-api",
+    )
+
+    # Log audit attempt
+    async with AuditService.with_session(audit_role, session=session) as svc:
```
daryllimyt: style: can just be

```suggestion
   svc = AuditService(session, audit_role)
```

tracecat/organization/service.py
```diff
@@ -483,9 +565,90 @@ async def create_invitation(
...
+        if workspace_assignments:
+            ws_invite_count = await self._create_workspace_invitations(
+                email=email,
+                assignments=workspace_assignments,
+            )
+
         await self.session.commit()
         await self.session.refresh(invitation)
-        return invitation
+        return invitation, ws_invite_count
```
daryllimyt: is the invite count used? not sure if it's worth changing the method signature to return this if not

tracecat/organization/service.py
```diff
@@ -483,9 +565,90 @@ async def create_invitation(
...
+                f"Workspaces do not belong to this organization: {invalid_ws_ids}"
+            )
+
+        now = datetime.now(UTC)
+        expires_at = now + timedelta(days=7)
+        count = 0
+
+        for assignment in assignments:
+            # Check for existing workspace invitation
+            existing_stmt = select(Invitation).where(
```
daryllimyt: P1: this `select -> maybe delete -> insert` pattern is vulnerable to races under concurrency. Two requests can both see no active row and then contend on insert/unique constraint. Recommend making this atomic with Postgres upsert (`INSERT ... ON CONFLICT (workspace_id, email) DO UPDATE ... WHERE status != PENDING OR expires_at <= now`) and counting created/updated rows from `RETURNING`. Also consider normalizing/storing email consistently (or a unique index on `lower(email)`) to avoid casing bypasses.

daryllimyt: Suggested shape (atomic upsert):

```python
from sqlalchemy import or_
from sqlalchemy.dialects.postgresql import insert

stmt = (
    insert(Invitation)
    .values(
        workspace_id=assignment.workspace_id,
        email=email.lower(),
        role=assignment.role,
        status=InvitationStatus.PENDING,
        invited_by=self.role.user_id,
        token=secrets.token_urlsafe(32),
        expires_at=expires_at,
        accepted_at=None,
    )
    .on_conflict_do_update(
        index_elements=[Invitation.workspace_id, Invitation.email],
        set_={
            "role": assignment.role,
            "status": InvitationStatus.PENDING,
            "invited_by": self.role.user_id,
            "token": secrets.token_urlsafe(32),
            "expires_at": expires_at,
            "accepted_at": None,
        },
        where=or_(
            Invitation.status != InvitationStatus.PENDING,
            Invitation.expires_at <= now,
        ),
    )
    .returning(Invitation.id)
)

row = (await self.session.execute(stmt)).scalar_one_or_none()
if row is not None:
    count += 1
```

This removes the select/delete/insert race window and keeps stale-vs-active replacement explicit.

daryllimyt (changes requested):

tracecat/workspaces/service.py
jordan-umusu: no need, the validation error is only raised when the update no-op's and hasn't committed any changes.

tracecat/organization/service.py
jordan-umusu: Agreed though it was, but was low value in the first place.

tracecat/invitations/router.py
```diff
@@ -0,0 +1,217 @@
...
+    For workspace-scoped invitations, checks ``workspace:member:{scope_action}``.
+    For org-scoped invitations, checks ``org:member:invite``.
+
+    Raises:
+        HTTPException 403 if the user lacks the required scope.
+    """
+    if invitation.workspace_id is not None:
+        required = f"workspace:member:{scope_action}"
+    else:
+        required = "org:member:invite"
```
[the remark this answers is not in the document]
jordan-umusu: Intended, remove memberships use the remove member scope

PR #2126, opened 2026-02-18
refactor(cases): Split list/search API contracts

topher-lo:
### Motivation
- The backend API was changed to split case listing vs search semantics into `/cases` and `/cases/search`, so the frontend client must be regenerated to match the new OpenAPI contract. 
- Keep generated TypeScript types and service bindings consistent with server-side behavior to avoid mismatches in hooks and pagination logic.

### Description
- Ran the project-standard client generation (`just gen-client-ci`) and committed the updated generated client artifacts `frontend/src/client/services.gen.ts` and `frontend/src/client/types.gen.ts`.
- The generated types now make `CasesListCasesData` a simplified listing contract (`limit`, `orderBy`, `sort`, `workspaceId`) and `casesListCases` sends only those params to `GET /cases`.
- `CasesSearchCasesData` retains full filtering/pagination fields and `casesSearchCases` targets `GET /cases/search` with the complete set of query parameters.
- Auto-format and lint steps invoked during generation were run to keep generated code consistent with repo formatting rules.

### Testing
- Executed `just gen-client-ci` to regenerate the client and run associated auto-fix hooks, and it succeeded.
- Ran `pnpm -C frontend run typecheck` and it completed with no TypeScript errors.
- The generation workflow also ran `pnpm -C frontend check` (frontend Biome) and `uv run ruff check . --fix` / `uv run ruff format .` as part of the `just gen-client-ci` workflow, and those steps passed.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6994e840731c83339db941841dbf168b)

---
## Summary by cubic
Split the Cases API into simple list (/cases) and full search (/cases/search), added cursor pagination to both, and fixed registry UDFs to forward cursor/reverse only when paginate=true. Updated SDK, routers, services, frontend hooks/client, and tests to match the new contracts.

- **Refactors**
  - Backend: /cases now only accepts limit, cursor, reverse, order_by, sort; all filtering moved to /cases/search. Public/internal routers have separate handlers; search paths call CasesService.search_cases.
  - Services: search_cases performs filtering + pagination; list_cases wraps it with simplified defaults and builds CursorPaginationParams.
  - Registry UDFs: list_cases forwards only basic params; search_cases builds full filter params and conditionally forwards cursor/reverse when paginate=true.
  - SDK: list_cases hits /cases with basic params; search_cases hits /cases/search with full filters and cursor controls.
  - Frontend: Regenerated client. Hooks now call casesSearchCases for filtered views; types ensure casesListCases only takes basic params.
  - Tests: Updated unit/registry/API tests to target search for filtered queries and to assert conditional cursor forwarding.

- **Migration**
  - Use casesSearchCases for any filtered views or when using search_term, tags, dates, or assignee filters.
  - Use casesListCases for simple lists with optional cursor pagination; do not send filters to /cases.

<sup>Written for commit f2729b008b12cdf7c170498e6bae84c68de2398f. Summary will update on new commits.</sup>

### DIFF
--- a/packages/tracecat-registry/tracecat_registry/core/cases.py
+++ b/packages/tracecat-registry/tracecat_registry/core/cases.py
@@ -270,13 +270,13 @@ async def list_cases(
     cursor: Annotated[
         str | None,
         Doc(
-            "Pagination cursor used to fetch a specific page. Response metadata is returned when paginate=true."
+            "Pagination cursor used to fetch a specific page when paginate=true."
         ),
     ] = None,
     reverse: Annotated[
         bool,
         Doc(
-            "Reverse pagination direction. Response metadata is returned when paginate=true."
+            "Reverse pagination direction when paginate=true."
         ),
     ] = False,
     order_by: Annotated[
@@ -299,9 +299,9 @@ async def list_cases(
         )
 
     params: dict[str, Any] = {"limit": limit}
-    if cursor is not None:
+    if paginate and cursor is not None:
         params["cursor"] = cursor
-    if reverse:
+    if paginate and reverse:
         params["reverse"] = reverse
     if order_by is not None:
         params["order_by"] = order_by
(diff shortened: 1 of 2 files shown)


topher-lo:
@cursor review

PR #2127, opened 2026-02-18
fix: Isolate subprocess protocol from action output

topher-lo:
### Motivation
- The direct executor parses subprocess `stdout` as the JSON protocol payload, so any action or integration that writes to `stdout`/`stderr` can corrupt protocol framing. 
- A targeted change in one integration (e.g., Ansible) is partial and fragile, so the containment must live at the executor boundary where subprocess protocol is enforced. 

### Description
- In `main_minimal` (`tracecat/executor/minimal_runner.py`) capture action-level `stdout` and `stderr` using `contextlib.redirect_stdout`/`redirect_stderr` into `io.StringIO` while invoking `run_action_minimal`. 
- Emit any suppressed output as `RuntimeWarning` (truncated to 500 chars) so noisy output remains observable without corrupting the JSON protocol. 
- Add unit tests in `tests/unit/executor/test_minimal_runner_protocol.py` that mock `importlib.import_module` to verify noisy actions return a clean structured success payload and that raised exceptions return structured error payloads. 

### Testing
- Ran `uv run ruff check tracecat/executor/minimal_runner.py tests/unit/executor/test_minimal_runner_protocol.py` and it passed. 
- Ran `uv run ruff format --check tracecat/executor/minimal_runner.py tests/unit/executor/test_minimal_runner_protocol.py` and it passed after formatting. 
- Attempted `uv run pytest tests/unit/executor/test_minimal_runner_protocol.py` but the repository-level pytest fixtures require a PostgreSQL instance on `localhost:5432`, causing the run to fail in this environment; a focused manual smoke check (`uv run python - <<'PY' ...`) confirmed `main_minimal` returns a structured success even when the action prints to `stdout`/`stderr`.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_699514f089d0833390051b9d47d5ea5b)

---
## Summary by cubic
Isolated the executor’s JSON protocol from action output by capturing and capping action stdout/stderr in minimal_runner and defaulting Ansible runs to quiet. This prevents protocol corruption and limits memory growth while keeping results and errors structured.

- **Bug Fixes**
  - Capture action stdout/stderr with capped buffers (8 KB); emit a RuntimeWarning with up to 500 chars preview, and fall back to stderr if warnings fail.
  - Default ansible-runner to quiet=True (overridable via runner_kwargs).
  - Add tests for noisy actions, structured errors, quiet override, buffer cap, and warning fallback; satisfy basedpyright.

<sup>Written for commit 32105bf7defda890726aaac7288ce0466ca0f780. Summary will update on new commits.</sup>

### DIFF
--- a/packages/tracecat-registry/tracecat_registry/integrations/ansible.py
+++ b/packages/tracecat-registry/tracecat_registry/integrations/ansible.py
@@ -79,6 +79,10 @@ def run_playbook(
 
     extravars = extravars or {}
     runner_kwargs = runner_kwargs or {}
+    # Prevent ansible-runner from writing console output to stdout.
+    # Registry actions are executed in a subprocess where stdout must contain
+    # only the JSON protocol payload returned by minimal_runner.
+    runner_kwargs.setdefault("quiet", True)
 
     if "inventory" in runner_kwargs:
         raise ValueError(
--- a/tests/registry/test_ansible.py
+++ b/tests/registry/test_ansible.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+
+from typing import Any
+
+import tracecat_registry.integrations.ansible as ansible_integration
+
+
+class _FakeRunner:
+    def __init__(self) -> None:
+        self.stdout = ""
+        self.events = []
+
+
+def test_run_playbook_sets_quiet_by_default(monkeypatch) -> None:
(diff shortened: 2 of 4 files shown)


topher-lo:
@cursor review

daryllimyt (approved):
lgtm

PR #2130, opened 2026-02-18
fix: make action graph migration tolerant of legacy act IDs

daryllimyt:
## Summary
- make `9079ae19374b` tolerant of mixed action ID formats in legacy `workflow.object` graph payloads
- coerce node/edge action IDs from either UUID strings or legacy `act-<32hex>` format before updating `action` rows
- skip uncoercible/invalid IDs instead of crashing migration execution

## Why this is safe
- the migration only updates non-key fields on `action`: `position_x`, `position_y`, and `upstream_edges`
- it does **not** modify primary keys, foreign keys, or constraints
- if an ID is orphaned (no matching `action.id`), the `UPDATE ... WHERE id = :id` affects zero rows
- `workflow.object` is dropped at the end of this migration, so stale graph-only orphan IDs are naturally discarded

## Validation
- `uv run ruff check alembic/versions/9079ae19374b_add_action_graph_fields.py`
- `uv run basedpyright alembic/versions/9079ae19374b_add_action_graph_fields.py`

---
## Summary by cubic
Make the action graph migration tolerant of legacy action IDs (act-<32hex>) and UUIDs so it doesn’t crash on mixed formats. Ensures positions and upstream edges are backfilled correctly.

- **Bug Fixes**
  - Coerce node/edge IDs to UUID and skip invalid IDs instead of failing.
  - Normalize edge source IDs to trigger-* or UUID.
  - Only updates non-key fields (position_x, position_y, upstream_edges); orphan IDs are no-ops and workflow.object is dropped after migration.

<sup>Written for commit 982c0e47068e8fbb45974b6630441377538c50e2. Summary will update on new commits.</sup>

### DIFF
--- a/alembic/versions/9079ae19374b_add_action_graph_fields.py
+++ b/alembic/versions/9079ae19374b_add_action_graph_fields.py
@@ -7,6 +7,8 @@ Create Date: 2025-12-03 15:22:20.734638
 """
 
 import json
+import re
+import uuid
 from collections.abc import Sequence
 
 import sqlalchemy as sa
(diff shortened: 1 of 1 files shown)


jordan-umusu (approved):
LGTM

PR #2155, opened 2026-02-23
chore: drop rbac feature flag

jordan-umusu:
## Summary by cubic
Dropped the RBAC feature flag and replaced it with the rbac_addons entitlement. EE RBAC APIs now require this entitlement, and the UI only shows RBAC pages when it’s present; scope checks still apply.

- **Refactors**
  - Enforced rbac_addons on all EE RBAC routers via a router-level dependency; removed app-level feature-flag gating.
  - Removed FeatureFlag.RBAC; added rbac_addons across tiers (enums, schemas, types, service, defaults map legacy “rbac” flag) and in generated TS types.
  - Frontend now uses useEntitlements and the rbac_addons key in org/workspace pages and settings; the members toggle renders only when entitled; loading states handled.

- **Migration**
  - Remove any "rbac" feature flag from config.
  - Set entitlements.rbac_addons to enable RBAC endpoints and UI.

<sup>Written for commit 825d485e100d4fc6c45dbb835b09d356ef8b7b89. Summary will update on new commits.</sup>

### DIFF
--- a/tracecat/api/app.py
+++ b/tracecat/api/app.py
@@ -80,7 +80,7 @@ from tracecat.db.dependencies import AsyncDBSession
 from tracecat.db.engine import get_async_session_context_manager
 from tracecat.editor.router import router as editor_router
 from tracecat.exceptions import EntitlementRequired, ScopeDeniedError, TracecatException
-from tracecat.feature_flags import FeatureFlag, FlagLike, is_feature_enabled
+from tracecat.feature_flags import FlagLike, is_feature_enabled
 from tracecat.feature_flags.router import router as feature_flags_router
 from tracecat.inbox.router import router as inbox_router
 from tracecat.integrations.router import (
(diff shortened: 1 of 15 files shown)


tracecat/api/app.py
```diff
@@ -456,26 +456,11 @@ def create_app(**kwargs) -> FastAPI:
...
-    )
-    app.include_router(
-        rbac_assignments_router,
-        dependencies=[Depends(feature_flag_dep(FeatureFlag.RBAC))],
-    )
-    app.include_router(
-        rbac_user_assignments_router,
-        dependencies=[Depends(feature_flag_dep(FeatureFlag.RBAC))],
-    )
+    app.include_router(rbac_scopes_router)
```
daryllimyt, line 459: [P1] Gate RBAC management routers by entitlement

These routers are now mounted without any guard, so organizations with `rbac` entitlement set to `false` (the new default) can still call `/rbac/*` APIs directly if the caller has RBAC scopes. That bypasses the new entitlement gating added in the UI and makes RBAC management functionality available even when the plan says it is disabled.

tracecat/tiers/enums.py
```diff
@@ -10,3 +10,4 @@ class Entitlement(StrEnum):
     GIT_SYNC = "git_sync"
     AGENT_ADDONS = "agent_addons"
     CASE_ADDONS = "case_addons"
+    RBAC = "rbac"
```
daryllimyt: maybe RBAC_ADDONS/RBAC_ADVANCED? RBAC kind of implies it's either available or not

daryllimyt (changes requested):

daryllimyt (changes requested):
router guards + naming

daryllimyt (approved):

PR #2187, opened 2026-02-25
fix(rbac): dedup seeded scopes

jordan-umusu:
## Problem
  Registry action scopes were being seeded in two places: as platform scopes (organization_id = NULL) and again as per-organization custom scopes. Because both used the same scope
  names, this created duplicate scope records and redundant role_scope links, which inflated RBAC data and made scope assignment non-canonical.
## Solution
  This PR makes platform scopes the single source of truth for registry actions and removes duplicate data paths:

---
## Summary by cubic
Canonicalizes duplicate registry action scopes by remapping role-scope links to platform scopes and deleting redundant custom scopes. Seeding now inserts platform scopes only, and the migration is constrained to registry action scopes (execute) to avoid touching unrelated scopes.

- **Bug Fixes**
  - Constrained dedupe to registry action scopes (resource=action, action=execute) using name+source_ref; remap role_scope links to platform scopes, drop stale custom links, and delete duplicate custom scope rows.
  - Updated seed_registry_scopes to seed platform scopes only; custom seeding paused until org-specific keys differ.
  - Adjusted tests to expect only platform insertions and zero custom duplicates; seeding remains idempotent.

- **Migration**
  - Run database migrations (alembic upgrade head).
  - Downgrade is irreversible due to data cleanup.

<sup>Written for commit c098cd60854e95d8c06446a35635e1c0f6e77d9b. Summary will update on new commits.</sup>

### DIFF
--- a/alembic/versions/c9e4f54f0a2b_canonicalize_duplicate_custom_scope_links.py
+++ b/alembic/versions/c9e4f54f0a2b_canonicalize_duplicate_custom_scope_links.py
@@ -32,8 +32,16 @@ def upgrade() -> None:
             FROM scope AS custom_scope
             JOIN scope AS platform_scope
                 ON platform_scope.name = custom_scope.name
+               AND platform_scope.source_ref = custom_scope.source_ref
             WHERE custom_scope.organization_id IS NOT NULL
+              AND custom_scope.source = 'CUSTOM'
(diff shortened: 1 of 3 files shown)


daryllimyt (approved):
lgtm

PR #2228, opened 2026-03-02
fix(mcp): fail fast in missing creds

jordan-umusu:
## Summary by cubic
Fail fast on MCP startup when required config (e.g., OIDC creds) is missing by exiting immediately with code 0 to prevent retries. Transient errors still retry.

- **Bug Fixes**
  - Added MCPNonRetryableStartupError; map auth config ValueError to it during server init.
  - main catches it, logs with logger.error, and exits with code 0.
  - Tests updated: added immediate-exit test (expects code 0); retry test uses transient RuntimeError.

<sup>Written for commit 7c1fd83a0e1477d349ad9bf32ae9919cc3a35ca6. Summary will update on new commits.</sup>

### DIFF
--- a/tracecat/mcp/__main__.py
+++ b/tracecat/mcp/__main__.py
@@ -43,7 +43,7 @@ def main() -> None:
             logger.info("MCP server interrupted; shutting down")
             return
         except MCPNonRetryableStartupError:
-            logger.exception(
+            logger.error(
                 "MCP server startup failed due to non-retryable configuration error"
             )
             raise SystemExit(1) from None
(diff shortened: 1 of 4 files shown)


tracecat/mcp/__main__.py
```diff
@@ -41,6 +42,11 @@ def main() -> None:
         except KeyboardInterrupt:
             logger.info("MCP server interrupted; shutting down")
             return
+        except MCPNonRetryableStartupError:
+            logger.exception(
```
daryllimyt: nit: can we use error here? we probably don't care about the stack trace for this error

daryllimyt (approved):

PR #2229, opened 2026-03-02
fix(frontend): sanitize AI chat markdown before rendering

topher-lo:
Problem: AI chat markdown could render unsafe HTML/attributes; DOMPurify caused SSR blank output and hydration mismatches. CSP alone doesn’t prevent dangerous protocols or clickable phishing links.

Solution:
- Sanitize during markdown rendering using rehype-sanitize with a safe schema; apply sanitization before KaTeX to preserve math.
- Constrain URLs with SAFE_MARKDOWN_LINK_PREFIXES and SAFE_MARKDOWN_IMAGE_PREFIXES via Streamdown.
- In Response, spread ...props first, then set allowed*Prefixes and rehypePlugins so callers can’t override sanitization.
- Harden CSP (base-uri 'self', script-src-attr 'none').
- Remove DOMPurify usage to avoid SSR pitfalls; tests mock ESM rehype plugins in jest.setup.js.


topher-lo:
@cursor review

topher-lo:
@cursor push fc97b08677

PR #2235, opened 2026-03-03
feat(api): Unify case and table aggregation search contracts

topher-lo:
Summary
- replace the split case aggregation endpoints with a single POST search contract that returns scalar and grouped aggregation objects
- keep the internal table search path intact while extending schemas/services/SDKs to support the new aggregation response shape and functions
- normalize aggregation inputs (`avg` → `mean`), enforce validation (bucket limits, function/field allowlists), and update the frontend hooks/components to consume the shared response
Testing
- Not run (not requested)

---
## Summary by cubic
Unifies case and table search under a single POST contract with built-in aggregation. Replaces the legacy case aggregates endpoint and updates schemas, services, SDKs, and frontend to the new response shape.

- **New Features**
  - POST /cases/search returns items plus an aggregation result (scalar value and grouped buckets).
  - Shared search schema (tracecat/search) with standard agg functions (count, sum, min, max, mean, median, mode, n_unique) and bucket limits.
  - Tables now support the same aggregation options via POST /tables/{table}/search (group_by, agg, agg_field, bucket_limit).
  - Normalized agg inputs (avg → mean, value_counts → n_unique) and enforced validation (bucket limits, field/function allowlists, short_id format).
  - Added DB indexes on case_tag_link and case_dropdown_value for faster grouped queries.

- **Migration**
  - /cases/search is now POST with a JSON body (filters, pagination, group_by, agg, agg_field, bucket_limit). Update clients to send a body instead of query params.
  - Removed CasesSearchCaseAggregates. Use the aggregation in /cases/search.
  - SDK methods now post request bodies and accept aggregation fields; adopt normalized agg names.
  - Response includes aggregation under aggregation. Update frontend hooks/components to read this shape (stage counts via hooks, removed CaseSearchAggregateRead).

<sup>Written for commit 12b58aa136a76fe454370ef434e3a3e8c88a934e. Summary will update on new commits.</sup>

### DIFF
--- a/alembic/versions/d6e1f2a0a1ec_add_case_aggregation_indexes.py
+++ b/alembic/versions/d6e1f2a0a1ec_add_case_aggregation_indexes.py
@@ -1,0 +1,40 @@
+"""add case aggregation indexes
+
+Revision ID: d6e1f2a0a1ec
+Revises: 8f4f1bd13e9c
+Create Date: 2026-03-02 16:35:53.844817
+
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "d6e1f2a0a1ec"
+down_revision: str | None = "8f4f1bd13e9c"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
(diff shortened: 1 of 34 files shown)


topher-lo:
@cursor review

PR #2239, opened 2026-03-03
feat: Unify case and table search aggregation contract

topher-lo:
Summary
- consolidate case search APIs around the new aggregation-aware POST contract and remove the legacy GET aggregation endpoint
- extend the existing table search handler to return the stable aggregation response shape while keeping the same route
- add the new aggregation schemas/validation so scalar and grouped aggregates always include `value`, `buckets`, and limit/truncation metadata

Testing
- Not run (not requested)

---
## Summary by cubic
Unifies case and table search behind a single POST, body-based, aggregation-aware contract. Updates backend, registry SDK, and frontend to a shared request/response shape.

- **New Features**
  - POST /cases/search and /tables/{table}/search accept body filters, order_by/sort, and aggregation params (group_by, agg, agg_field, bucket_limit).
  - Responses include cursor pagination and aggregation.value + aggregation.buckets with truncation metadata.
  - Shared search schemas added; “avg” normalized to “mean,” with env limits for bucket sizes.
  - Frontend client/hooks switch to POST; cases hook derives totals and stage counts from aggregation; table link command updated.
  - Registry SDK forwards aggregation params and returns structured payloads when cursor/aggregation is used; keeps list-only shape otherwise. Alembic indexes added for tag/dropdown aggregation paths.

- **Migration**
  - Switch case searches to POST /cases/search with a JSON body; legacy GET search/aggregate endpoints are removed.
  - Read totals from aggregation.value and stage counts from aggregation.buckets.
  - Optional: configure TRACECAT__LIMIT_AGG_BUCKET_DEFAULT and TRACECAT__LIMIT_AGG_BUCKET_MAX.

<sup>Written for commit a12c9ad5bbbd2aedfb52d1a44102d2988a5f4e6e. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/components/ai-elements/response.tsx
+++ b/frontend/src/components/ai-elements/response.tsx
@@ -5,7 +5,6 @@ import { Streamdown } from "streamdown"
 import {
   ALLOWED_MARKDOWN_IMAGE_PREFIXES,
   ALLOWED_MARKDOWN_LINK_PREFIXES,
-  DEFAULT_MARKDOWN_ORIGIN,
   getStreamdownRehypePlugins,
 } from "@/lib/sanitize-markdown"
 import { cn } from "@/lib/utils"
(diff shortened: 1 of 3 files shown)


topher-lo:
@cursor review

PR #2247, opened 2026-03-03
fix(cases): use OR matching for multi-tag case filters

daryllimyt:
### Motivation
- Filtering cases by multiple tags previously added one filter per tag which produced an implicit AND (requiring a case to have all selected tags) instead of returning cases that match any of the selected tags.

### Description
- Change `CasesService._build_search_filters` to use a single `IN` predicate: `CaseTagLink.tag_id.in_(tag_ids)` so multiple tag IDs are matched with OR semantics.
- Add a regression unit test `test_search_cases_tag_filter_uses_or_logic` in `tests/unit/test_cases_service.py` that creates two tags, attaches them to separate cases, and verifies `search_cases(tag_ids=[...])` returns cases that have either tag and excludes unrelated cases.
- Update test imports to include `TagCreate` and `CaseTagsService` used by the new test and run formatting/linting (`ruff`) on the modified files.

### Testing
- Ran linter/formatter checks with `uv run ruff check tracecat/cases/service.py tests/unit/test_cases_service.py` which passed.
- Attempted to run the new unit test with `uv run pytest tests/unit/test_cases_service.py -k tag_filter_uses_or_logic -q`, but the test run failed in this environment due to a missing local PostgreSQL instance (connection refused to `localhost:5432`), not due to application logic; the test should pass when run in CI or a local environment with the test DB available.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69a74bec3d688320b74d1ffb9e71c3df)

---
## Summary by cubic
Fix case search with multiple tag filters to return cases that match any selected tag. Replaces per-tag AND filters with a single IN predicate and adds a regression test.

- **Bug Fixes**
  - Build one filter with Case.id.in_(select where CaseTagLink.tag_id.in_(tag_ids)) to apply OR logic.
  - Add regression test using CaseTagsService to create/link tags and verify cases with either tag are returned while unrelated cases are excluded.

<sup>Written for commit 1b6ea9370e09758f20ce359af4bbb7d29cf45872. Summary will update on new commits.</sup>

### DIFF
--- a/tests/unit/test_cases_service.py
+++ b/tests/unit/test_cases_service.py
@@ -25,7 +25,9 @@ from tracecat.cases.schemas import (
     CaseReadMinimal,
     CaseUpdate,
 )
+from tracecat.tags.schemas import TagCreate
 from tracecat.cases.service import CaseFieldsService, CasesService
+from tracecat.cases.tags.service import CaseTagsService
 from tracecat.db.models import Case
 from tracecat.exceptions import TracecatAuthorizationError
 from tracecat.pagination import CursorPaginationParams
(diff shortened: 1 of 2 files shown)


jordan-umusu:
LGTM, but this seems just a design choice. The implicit OR is probably the more common use case, but the implicit AND doesn't seem uncommon either.

jordan-umusu (approved):

PR #2265, opened 2026-03-05
fix(executor): handle Pydantic model return types in sandbox runner

jordan-umusu:
## Summary by cubic
Fixes JSON serialization in the sandbox runner by handling Pydantic model return types with an orjson default that calls model_dump(mode="json"). Prevents TypeError when tasks return models, including nested and rich types.

- **Bug Fixes**
  - Added _orjson_default to serialize via model_dump(mode="json"); raises TypeError with the type name when unavailable or fails.
  - Updated json_dumps to use the default handler; added regression tests for nested models, rich types, and failure cases; corrected Slack sink tests to patch the slack module paths.

<sup>Written for commit e5563d3d406e44ed4a8cf128bf092a53c2723b59. Summary will update on new commits.</sup>

### DIFF
--- a/tests/unit/executor/test_minimal_runner_protocol.py
+++ b/tests/unit/executor/test_minimal_runner_protocol.py
@@ -3,7 +3,14 @@ from __future__ import annotations
 import io
 import sys
 import types
+from datetime import datetime
+from decimal import Decimal
 from typing import Any
+from uuid import UUID
+
+import orjson
+import pytest
+from pydantic import BaseModel
 
 from tracecat.executor import minimal_runner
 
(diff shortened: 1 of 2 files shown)


tracecat/executor/minimal_runner.py
```diff
@@ -32,8 +32,13 @@ def json_loads(data: bytes | str) -> dict:
             data = data.encode()
         return orjson.loads(data)
 
+    def _orjson_default(obj: Any) -> Any:
+        if hasattr(obj, "model_dump"):
+            return obj.model_dump()
```
daryllimyt:
```suggestion
            return obj.model_dump(mode="json")
```

daryllimyt (changes requested):

daryllimyt (approved):

PR #2308, opened 2026-03-09
feat: Run workflow from case comments

topher-lo:
Summary
- add workflow selector support across frontend comments, hooks, and registry SDK schemas
- align case services, triggers, and migrations with workflow-backed comments and retain normalization per conventions
- regenerate generated client code to match updated schemas and types

Testing
- Not run (not requested)

---
## Summary by cubic
Run workflows from case comments with a selector and live status. Users can attach a workflow to a comment, auto-start it, see progress, and retries are idempotent to prevent duplicate runs.

- **New Features**
  - Case comments: select a workflow when posting; auto-start on submit; comment shows workflow title/alias, execution link, and status (running/succeeded/failed). Duplicate starts are rejected and safe to retry.
  - Frontend: workflow/folder selector; fixed compact execution fetch (removed URL-encoding); restored OIDC authorize redirect.
  - API/Backend: `CaseCommentCreate` accepts `workflow_id`; `CaseCommentRead` includes `workflow` metadata and status enum; generated clients updated; `tracecat-registry` schemas/types updated. DB adds workflow snapshot fields and drops the `case_comment.workflow_id` FK. Publisher supports extra fields; consumer starts workflows, updates comments with execution ID/status, and records audit events. Case triggers and workflow management accept short workflow IDs and backfill defaults; explicit workflow updates commit before dedupe; workflow dispatch uses `WorkflowIDReusePolicy.REJECT_DUPLICATE`.

- **Migration**
  - Run database migrations to add workflow columns and drop the FK on `case_comment`.
  - No data backfill required; existing comments remain unchanged.

<sup>Written for commit a360c0870db3ee5eb60416ed078c3851c7eb809b. Summary will update on new commits.</sup>


topher-lo:
@cursor review

PR #2319, opened 2026-03-10
feat(mcp): Add workflow folder tools

topher-lo:
## Summary
This change adds a small MCP-only workflow folder management surface without changing the existing REST API, frontend, or workflow service contracts. Agents can now inspect folder structures, create folders by path, and move workflows into or out of folders in batches through the MCP server. The tree listing also returns workflow tag metadata so nested workflows can be tagged directly with the existing workflow tag MCP tools.

## Issue and user impact
Before this change, the MCP server exposed general workflow CRUD but not folder-oriented operations. That made agents inefficient at organizing workflows because they had to rely on UI-only or REST-only folder capabilities that were not available through MCP. In practice, an agent could not accurately inspect nested folder structure, create missing folders, or perform a multi-workflow reorganization flow over MCP.

## Root cause
Folder management primitives already existed in the application services and REST routers, but they were not surfaced in `tracecat/mcp/server.py`. The MCP surface also did not return folder-aware workflow metadata that would help an agent continue into tagging operations after locating a nested workflow.

## Fix
The MCP server now exposes three thin wrappers over the existing workflow folder services:

- `list_workflow_tree` lists folders and workflows under a path with bounded or unlimited depth.
- `create_workflow_folder` creates a folder from an absolute path and supports parent creation when requested.
- `move_workflows` performs best-effort, non-atomic batch moves into a destination folder or back to root, with a `dry_run` mode for validation.

A small path normalization helper was added to keep MCP path handling explicit and predictable. The tree listing now includes each workflow's `workflow_id`, `alias`, and current `tags` so agents can tag foldered or nested workflows directly using the existing tag tools.

## Safety and compatibility
This change is intentionally scoped to the MCP server. It does not modify the frontend, REST routes, or underlying service implementations, which keeps regression risk low for existing UI and API behavior. Batch moves are described and implemented as best-effort and non-atomic because the underlying single-workflow move service commits per workflow today; the MCP contract does not claim stronger guarantees than the current service layer provides.

## Validation
I validated the change with the existing repository checks that apply to the touched file:

- `uv run ruff check tracecat/mcp/server.py`
- `uv run basedpyright --warnings tracecat/mcp/server.py`
- pre-commit hooks during `git commit`

---
## Summary by cubic
Add MCP-only workflow folder tools so agents can list folders, create folders, and batch-move workflows via the MCP server. Hardened path and validation edge cases; tree listings include tag metadata for tagging nested workflows.

- **New Features**
  - `list_workflow_tree(path, depth, include_workflows)` lists folders and workflows from a path. Returns normalized `root_path` and items. Each workflow includes `workflow_id`, `alias`, `title`, `status`, `tags`, and `folder_path`. Depth `0` is unlimited.
  - `create_workflow_folder(path, parents)` creates a folder by absolute path. Optional `parents` creates missing ancestors. Rejects root paths and "."/".." segments. Returns `folder_id`, `created_paths`, and whether it already existed.
  - `move_workflows(workflow_ids, destination_path, dry_run)` moves workflows into a folder or back to root. Best-effort, non-atomic. Validates workflow IDs and destination folder; `dry_run` reports counts and per-item errors.
  - MCP-only change. No updates to REST, frontend, or underlying service contracts.

<sup>Written for commit 5c00ecfe6af13e2d981332cd7f44c1f05d622704. Summary will update on new commits.</sup>

### DIFF
--- a/tracecat/mcp/server.py
+++ b/tracecat/mcp/server.py
@@ -120,6 +120,7 @@ from tracecat.workflow.case_triggers.schemas import (
 from tracecat.workflow.case_triggers.service import CaseTriggersService
 from tracecat.workflow.executions.service import WorkflowExecutionsService
 from tracecat.workflow.management.definitions import WorkflowDefinitionsService
+from tracecat.workflow.management.folders.service import WorkflowFolderService
 from tracecat.workflow.management.management import WorkflowsManagementService
 from tracecat.workflow.management.schemas import WorkflowCreate, WorkflowUpdate
 from tracecat.workflow.schedules.schemas import (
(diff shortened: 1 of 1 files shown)


topher-lo:
@cursor review

PR #2323, opened 2026-03-10
fix(mcp): provide oauth metadata

jordan-umusu:
## Summary by cubic
Expose OAuth/OIDC discovery and public-client registration for MCP, and relax edge WAF to allow OAuth bootstrapping without a User-Agent. This lets MCP clients auto-discover endpoints and complete authorization reliably.

- **Bug Fixes**
  - Advertise public-client auth by patching discovery to include token_endpoint_auth_methods_supported = ["none", "client_secret_post", "client_secret_basic"] and preserve all upstream fields; include registration_endpoint and support public-client registration at `/register`.
  - Serve protected-resource discovery at `/.well-known/oauth-protected-resource/mcp` with accurate `resource` and `authorization_servers`.
  - Merge platform OIDC scopes into requested scopes (order-preserving, no duplicates) and add/remove `offline_access` based on provider support.
  - Update AWS WAF to allow missing User-Agent on MCP public endpoints (discovery, transport, OAuth) while blocking elsewhere; add regex pattern set and a targeted rule, and adjust rule priorities.
  - Expand tests for metadata endpoints, protected-resource metadata, public-client registration, and scope merging.

<sup>Written for commit bbf5932f091bc16e4673a8b2b0103a14c8b9ec88. Summary will update on new commits.</sup>

### DIFF
--- a/tracecat/mcp/auth.py
+++ b/tracecat/mcp/auth.py
@@ -19,20 +19,19 @@ from fastmcp.server.dependencies import get_access_token
 from key_value.aio.stores.redis import RedisStore
 from key_value.aio.wrappers.encryption import FernetEncryptionWrapper
 from key_value.aio.wrappers.prefix_collections import PrefixCollectionsWrapper
-from mcp.server.auth.handlers.metadata import MetadataHandler
 from mcp.server.auth.provider import (
     AuthorizationParams,
     TokenError,
 )
-from mcp.server.auth.routes import build_metadata, cors_middleware
-from mcp.server.auth.settings import ClientRegistrationOptions, RevocationOptions
 from mcp.shared.auth import OAuthClientInformationFull
 from pydantic import BaseModel, Field
 from redis.asyncio import Redis as AsyncRedis
 from sqlalchemy import select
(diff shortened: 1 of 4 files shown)


tracecat/mcp/auth.py
```diff
@@ -689,6 +703,33 @@ async def _show_consent_page(
...
+
+                metadata = build_metadata(
+                    self.base_url,
+                    self.service_documentation_url,
+                    self.client_registration_options or ClientRegistrationOptions(),
+                    self.revocation_options or RevocationOptions(),
+                )
+                metadata.token_endpoint_auth_methods_supported = (
+                    _MCP_TOKEN_ENDPOINT_AUTH_METHODS
+                )
```
daryllimyt: This looks accidental rather than intentional. The change here rebuilds the discovery metadata so we can advertise `none` in `token_endpoint_auth_methods_supported`, but replacing the whole metadata object also drops fields the upstream handler was already publishing, including `client_id_metadata_document_supported`. So CIMD support disappeared from discovery as a side effect of rebuilding the metadata from scratch, not because this PR was clearly trying to turn CIMD off.

PR #2339, opened 2026-03-11
feat(mcp): Remove inline workflow YAML

topher-lo:
## Summary by cubic
Switch MCP workflows from inline YAML to file uploads stored as short‑lived, session‑scoped artifacts in `redis.asyncio`. Adds enforcement that remote uploads must be finalized before create/update, hardens POSIX path handling, and removes stdio-only MCP paths.

- **New Features**
  - File-based workflow definitions with upload and retrieval; uploads are Redis-backed with TTL and bound to the MCP session.
  - Removed inline `definition_yaml` from `create_workflow` and `update_workflow`.
  - New tools: `get_workflow_file`, `prepare_workflow_file_upload`, `create_workflow_from_uploaded_file`, `update_workflow_from_uploaded_file`.
  - Require finalized uploads before using `create_workflow_from_uploaded_file` or `update_workflow_from_uploaded_file`; in-progress uploads are rejected.

- **Migration**
  - Stop sending inline YAML to workflow tools.
  - Upload YAML via `prepare_workflow_file_upload(...)`, then pass the returned `artifact_id`.
  - Finalize the upload before calling `create_workflow_from_uploaded_file` or `update_workflow_from_uploaded_file`.
  - Use `create_workflow_from_uploaded_file` or `update_workflow_from_uploaded_file` for changes (optionally set `use_workflow_id=true` on create).
  - Use `get_workflow_file` to scaffold or update local files if needed.
  - Move to non-stdio transports; stdio-only MCP paths were removed.

<sup>Written for commit b3f47e01eda4177f658b0e56d0e4a632e024e810. Summary will update on new commits.</sup>

PR #2360, opened 2026-03-13
fix(mcp) workflow uploads and restore inline yaml

topher-lo:
## Summary
This fixes the MCP workflow editing regression introduced when inline workflow YAML support was removed in #2339. Remote MCP clients were forced through staged uploads, but those uploads reused the global blob presigned URL expiry, which defaults to 10 seconds. In practice that was too short for an LLM client to receive the presigned URL, reason about the next step, and then perform the PUT, so workflow updates could fail with expired upload URLs.

This change restores a hybrid model for workflow authoring. Small workflow definitions can be edited inline again through `create_workflow` and `update_workflow`, while larger workflow payloads continue to use the staged file transfer flow. `get_workflow` also regains an opt-in inline definition path so clients can fetch small workflow YAML directly without forcing a file round trip.

## What changed
The MCP server now has a dedicated `TRACECAT_MCP__FILE_TRANSFER_URL_EXPIRY_SECONDS` setting with a 5 minute default, and all staged MCP file flows use that instead of the generic 10 second blob presigned URL expiry. That applies to workflow downloads, workflow uploads, template uploads, and CSV exports.

Inline workflow editing was reintroduced on the existing CRUD tools with a 128 KB guardrail. Small workflow YAML payloads go through the same shared parsing and update helpers used by the staged finalize tools, so patch vs replace behavior, auto-layout generation, schedules, case triggers, and metadata updates stay aligned between both paths. For reads, `get_workflow(include_definition_yaml=true)` returns inline YAML when the serialized workflow stays under the limit and otherwise returns `definition_transport`, `definition_size_bytes`, `inline_limit_bytes`, and `suggested_relative_path` so the client can switch to staged retrieval.

The MCP docs and server instructions now describe the hybrid inline/staged contract, and the compose and Helm config wiring expose the new MCP-specific transfer TTL.

## User impact
Claude and other remote MCP clients can once again make small workflow edits without writing out a full local file and racing a 10 second upload URL. When workflows are too large for inline editing, the staged upload flow is still available, but it now has a practical TTL for real tool-call latency.

## Validation
I ran:
- `uv run ruff check --fix .`
- `uv run ruff format .`
- `TRACECAT__SERVICE_KEY=test-service-key uv run basedpyright --warnings --threads 4`
- `pnpm install --dir frontend`
- `pnpm -C frontend check`
- `pnpm -C frontend run typecheck`
- `TRACECAT__SERVICE_KEY=test-service-key uv run pytest tests/unit/test_mcp_server.py -q`

---
## Summary by cubic
Fixes MCP workflow upload timeouts and restores inline workflow YAML editing. Adds a 5‑minute MCP file transfer URL TTL and wires Helm to pass the blob storage presign endpoint for reliable remote uploads/downloads.

- **Bug Fixes**
  - Use `TRACECAT_MCP__FILE_TRANSFER_URL_EXPIRY_SECONDS` (default 300s) for all MCP staged transfers: workflow uploads/downloads, template uploads, and CSV exports.
  - Restore inline YAML editing for small workflows to avoid staged upload races.
  - Helm MCP env now includes blob storage config and sets `TRACECAT__BLOB_STORAGE_PRESIGNED_URL_ENDPOINT` when a public S3 URL is configured.

- **New Features**
  - `create_workflow` and `update_workflow` accept `definition_yaml` (≤128 KB). `update_workflow` adds `update_mode` ("patch" or "replace").
  - `get_workflow(include_definition_yaml=true)` returns inline YAML when small; otherwise returns `staged_required` with size, limit, and `suggested_relative_path`.
  - Docs and configs updated: new env var in `.env`, `docker-compose*`, Helm values; Helm merges `blobStorage` into MCP env.

<sup>Written for commit 04e7d130bbb2395d75c8e7de2c81b09bdb21ad09. Summary will update on new commits.</sup>


topher-lo:
@codex review

PR #2372, opened 2026-03-16
fix(mcp): preserve proxy identity and scopes across oidc token swap

jordan-umusu:
## Motivation
FastMCP's OIDC proxy correctly resolves user identity during the initial OAuth login (via         _extract_upstream_claims), embedding email and Tracecat-specific scopes into the proxy JWT. However, on subsequent MCP tool calls, load_access_token validates the proxy JWT by swapping it for the upstream provider's access token and the returned AccessToken.claims reflect the upstream token, not the proxy JWT. This silently drops the identity context (email, client_id, workspace/org scopes) that was established at login.

Until an upstream resolve_claims extension point exists, this PR overrides load_access_token to merge the proxy JWT's identity claims and scopes back into the validated AccessToken, ensuring the caller identity established at login survives the OIDC token swap on every tool call.

## Summary by cubic
Fixes MCP OIDC auth to preserve proxy identity and scopes when `fastmcp` swaps its JWT for the upstream token. Adds a reliable email resolution path using the OIDC `userinfo` endpoint to keep identities consistent.

- **Bug Fixes**
  - Preserve `client_id`, org/workspace scopes, and `upstream_claims` from the `fastmcp` JWT by merging them into the validated upstream token in `load_access_token`.
  - Resolve email from `id_token` if present; otherwise fetch from `userinfo` with `access_token` (accepts missing `id_token`), and require matching `sub` when both are available; trims whitespace.
  - Better error mapping: `userinfo` fetch failures -> `invalid_grant`; missing email after both sources -> `invalid_client`.

<sup>Written for commit 8cebdcefda164ff1897e409873dec12b82bd5bc9. Summary will update on new commits.</sup>

### DIFF
--- a/tests/unit/test_mcp_auth.py
+++ b/tests/unit/test_mcp_auth.py
@@ -2,10 +2,12 @@ from __future__ import annotations
 
 import uuid
 from types import SimpleNamespace
+from typing import cast
 from unittest.mock import MagicMock, patch
 
 import pytest
 from fastmcp import FastMCP
+from fastmcp.server.auth import AccessToken
 from mcp.server.auth.provider import AuthorizationParams
 from mcp.shared.auth import OAuthClientInformationFull
 from pydantic import AnyUrl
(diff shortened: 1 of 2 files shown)


jordan-umusu:
@codex review

PR #2403, opened 2026-03-20
feat(mcp+infra): Stand-in dex idp for mcp auth

jordan-umusu:
## Summary by cubic
Adds a Dex-backed stand-in IdP for MCP auth with local password provisioning and optional upstream OIDC under `/auth`, plus a dedicated Dex vhost and consent page. Adds a Tracecat-native SAML bridge for MCP and disables Dex routes when `MCP_AUTH_MODE="saml"`; restores Claude Code OAuth and only mounts platform OIDC login when `OIDC_ISSUER` is set.

- **New Features**
  - Dex service with dynamic modes (`basic`/`oidc`) auto-selected by `scripts/dex-entrypoint.sh`, gRPC password API, and a dedicated Caddy vhost (`${DEX_BASE_DOMAIN}`); in SAML mode, Caddy denies `/auth/*` and redirects `/auth/callback` to `/sign-in`.
  - MCP auth: OIDC uses Dex at `/auth` by default (requests `offline_access`, supports refresh tokens); SAML mode uses a built-in MCP SAML bridge that issues Tracecat tokens and serves a minimal consent page.
  - Env/infra: new `MCP_AUTH_MODE` (`basic`/`oidc`/`saml`) with Fargate/K8s support; Compose adds a `dex` service and `DEX_*`/`MCP_DEX_*` defaults; cluster script recomputes Dex/MCP URLs; new org APIs to list/delete MCP sessions with frontend client updates.

- **Bug Fixes**
  - Restored Claude Code OAuth by aligning OIDC proxying, refresh handling, and callbacks.
  - Guarded mixed-auth startup so platform OIDC login only mounts when configured; Dex local passwords follow SSO policy and are reconciled on user create/update and invitation acceptance.
  - SAML mode fixes: disable Dex endpoints, patch OIDC metadata for client compatibility, and have discovery respect the explicit `MCP_AUTH_MODE`.

<sup>Written for commit b8368953cc55de383106a59f6192a55da9c89b85. Summary will update on new commits.</sup>


tracecat/mcp/auth.py
```diff
@@ -645,31 +759,70 @@ def _create_oidc_auth() -> OIDCProxy:
...
 
-    oidc_config = get_platform_oidc_config()
+    oidc_config = get_mcp_oidc_config()
+    dex_mode = get_mcp_dex_mode()
     if not oidc_config.issuer:
-        raise ValueError("OIDC_ISSUER must be configured for the MCP server.")
+        raise ValueError("DEX_ISSUER must be configured for the MCP server.")
+    if not oidc_config.client_id or not oidc_config.client_secret:
+        raise ValueError(
+            "DEX_TRACECAT_CLIENT_ID and DEX_TRACECAT_CLIENT_SECRET must be configured for the MCP server."
```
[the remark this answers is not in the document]
jordan-umusu: Will address in a separate PR

docker-compose.yml
```diff
@@ -8,18 +8,55 @@ services:
...
     environment:
-      - BASE_DOMAIN=${BASE_DOMAIN}
+      - BASE_DOMAIN=${BASE_DOMAIN:-:${PUBLIC_APP_PORT:-80}}
+      - DEX_BASE_DOMAIN=${DEX_BASE_DOMAIN:-http://dex.localhost:${PUBLIC_APP_PORT:-80}}
       - ADDRESS=${ADDRESS}
     volumes:
       - ./Caddyfile:/etc/caddy/Caddyfile
 
+  dex:
+    image: ${DEX_IMAGE:-ghcr.io/dexidp/dex:v2.45.0}
```
daryllimyt, line 19: does this have to override the whole path

tracecat/mcp/auth.py
```diff
@@ -80,41 +92,161 @@ class MCPTokenIdentity(BaseModel):
...
-    """Append extra scopes while preserving order and uniqueness."""
-    merged = scopes
+def _merge_unique_scopes(
+    scopes: Sequence[str],
+    extra_scopes: Sequence[str],
+) -> list[str]:
+    merged = list(scopes)
     for scope in extra_scopes:
-        merged = append_scope_if_missing(merged, scope)
+        if scope not in merged:
```
daryllimyt, line 129: O(N^2), use set instead

tracecat/mcp/auth.py
```diff
@@ -125,8 +257,522 @@ def supports_refresh_scope(scopes_supported: Sequence[str] | None) -> bool:
...
+                "token_endpoint",
+                "userinfo_endpoint",
+                "jwks_uri",
+                "registration_endpoint",
+                "revocation_endpoint",
+                "introspection_endpoint",
+                "service_documentation",
+            )
+            if (
+                rewritten := _rewrite_issuer_url(
```
daryllimyt, line 312: nit: hard to read

tracecat/mcp/auth.py
```diff
@@ -125,8 +257,522 @@ def supports_refresh_scope(scopes_supported: Sequence[str] | None) -> bool:
...
+        requests_refresh_tokens=requests_refresh_tokens,
+        fallback_access_token_expiry_seconds=_MCP_ACCESS_TOKEN_FALLBACK_EXPIRY_SECONDS,
+    )
+
+
+class _TracecatOIDCProxy(OIDCProxy):
+    """OIDC proxy with user-existence validation and public-client registration."""
+
+    __slots__ = ("_settings",)
+
```
daryllimyt, line 281: nit: assign direct

tracecat/auth/users.py
```diff
@@ -81,8 +94,63 @@ def __init__(self, user_db: SQLAlchemyUserDatabase[User, uuid.UUID]) -> None:
...
+            and isinstance((password := getattr(user_update, "password", None)), str)
+            and password
+            else self._pending_state.dex_password_hash
+        )
+        previous_email = user.email if "email" in set_fields else None
+        with self._with_pending_state(
+            dex_password_hash=next_dex_password_hash,
+            previous_email=previous_email,
+        ):
+            return await super().update(user_update, user, safe=safe, request=request)
```
daryllimyt, line 154: nit: hard to read

tracecat/auth/users.py
```diff
@@ -81,8 +94,63 @@ def __init__(self, user_db: SQLAlchemyUserDatabase[User, uuid.UUID]) -> None:
...
+        previous_email: str | None = None,
+    ):
+        previous_state = self._pending_state
+        self._pending_state = PendingDexSyncState(
+            invitation_token=invitation_token,
+            dex_password_hash=dex_password_hash,
+            previous_email=previous_email,
+        )
+        try:
+            yield
```
daryllimyt, line 114: This works today because `UserManager` is request-scoped, and `_with_pending_state()` is nesting-safe because it restores the previous value on exit. The part that still makes me uneasy is that `_pending_state` is a mutable instance field, so it is only safe as long as the same `UserManager` instance is never used by concurrent coroutines. Since these flows cross `await` boundaries and depend on transient state surviving into `on_after_*` callbacks, I think this would be safer as task-local state via `contextvars`.

That change would remove the lifecycle assumption from the correctness story: instead of relying on `UserManager` never being shared concurrently, the pending Dex sync state would be isolated per task by construction. It would also match the existing pattern in `tracecat/contexts.py`.

Something like this would preserve the current push/pop behavior:

```py
from contextvars import ContextVar, Token

_pending_dex_sync_state: ContextVar[PendingDexSyncState | None] = ContextVar(
    "pending_dex_sync_state",
    default=None,
)

def _get_pending_dex_sync_state() -> PendingDexSyncState:
    return _pending_dex_sync_state.get() or PendingDexSyncState()

@contextlib.contextmanager
def _with_pending_state(
    self,
    *,
    invitation_token: str | None = None,
    dex_password_hash: str | None = None,
    previous_email: str | None = None,
):
    token: Token[PendingDexSyncState | None] = _pending_dex_sync_state.set(
        PendingDexSyncState(
            invitation_token=invitation_token,
            dex_password_hash=dex_password_hash,
            previous_email=previous_email,
        )
    )
    try:
        yield
    finally:
        _pending_dex_sync_state.reset(token)

jordan-umusu:
k8s deployment missing new conflig vars/flags and new dex service. also need clearer error messaging for when users attempt to /mcp without having joined the platform and organization

topher-lo:
Finally had a chance to look at this. Abort. @jordan-umusu  This code is too risky. Good attempt but looking at this code, I'm now convinced we should take @daryllimyt's original idea (i.e. the OIDC internal server) will be easier to reason and maintain.

Authlib supports building your own OIDC server with different hybrid / authcode grant types: https://github.com/authlib/example-oidc-server

3 files of self contained code. Will close and take this on. Lesson for next time: raise the alarm bells faster next time (if it's more than 10 files changed, there needs to be a very good reason why)

PR #2410, opened 2026-03-24
fix(agents): litellm hardening and logging

jordan-umusu:
## Summary by cubic
Hardens the `litellm` sidecar and gateway with robust health checks, shared load tracking, and request tracing. Adds benchmarks for sidecar throughput and time-to-first-token, switches the gateway to stdlib `logging`, and redacts secrets from error logs.

- **New Features**
  - Runtime: multi-worker `litellm` sidecar with readiness probes; fine-grained healthcheck timeouts (connect/read/write/pool); periodic status logs; worker exits on unhealthy sidecar.
  - Gateway/Proxy: switch to stdlib `logging`; shared process-local load tracker and per-request counters in logs; `x-request-id` added to requests/logs; provider credential cache with TTL; readiness endpoints allowed without auth; improved error mapping including 529 overload.
  - Benchmarks: `scripts/benchmark/litellm_sidecar_benchmark.py` and `scripts/benchmark/agent_session_first_token_benchmark.py` for latency and concurrency.
  - Config/Infra: new knobs `TRACECAT__LITELLM_NUM_WORKERS`, `TRACECAT__LITELLM_CREDENTIAL_CACHE_TTL_SECONDS`, `TRACECAT__LITELLM_HEALTHCHECK_INTERVAL_SECONDS`, `TRACECAT__LITELLM_HEALTHCHECK_TIMEOUT_SECONDS`, `TRACECAT__LITELLM_HEALTHCHECK_CONNECT_TIMEOUT_SECONDS`, `TRACECAT__LITELLM_HEALTHCHECK_READ_TIMEOUT_SECONDS`, `TRACECAT__LITELLM_HEALTHCHECK_WRITE_TIMEOUT_SECONDS`, `TRACECAT__LITELLM_HEALTHCHECK_POOL_TIMEOUT_SECONDS`, `TRACECAT__LITELLM_STATUS_LOG_INTERVAL_SECONDS`, `TRACECAT__LITELLM_HEALTHCHECK_FAILURE_THRESHOLD`; wired through Terraform and Docker Compose; k8s submodule bump.

- **Bug Fixes**
  - Executor loopback flags streamed assistant content to avoid missing results.
  - Gateway sanitizes exception messages to redact API keys and bearer tokens.

<sup>Written for commit 6d53be363c611aca71fde753b5a587d77bed694e. Summary will update on new commits.</sup>

### DIFF
--- a/tracecat/agent/gateway.py
+++ b/tracecat/agent/gateway.py
@@ -7,8 +7,10 @@ via JWT tokens minted by the agent executor.
 
 from __future__ import annotations
 
+import logging
 import re
 
+import orjson
 from fastapi import Request
 from litellm.caching.dual_cache import DualCache
 from litellm.integrations.custom_logger import CustomLogger
@@ -23,7 +25,6 @@ from tracecat.auth.types import Role
 from tracecat.authz.scopes import SERVICE_PRINCIPAL_SCOPES
 from tracecat.config import TRACECAT__LITELLM_CREDENTIAL_CACHE_TTL_SECONDS
 from tracecat.identifiers import OrganizationID, WorkspaceID
-from tracecat.logger import logger
 
 # Apply monkeypatches for LiteLLM adapter compatibility fixes
 apply_patch()
@@ -42,6 +43,7 @@ _hook_request_counters: dict[int, int] = {}
 _CREDENTIALS_CACHE_PREFIX = "tracecat:litellm:provider-creds"
 _SANITIZED_ERROR_MAX_LENGTH = 512
 _TRACE_REQUEST_ID_HEADER = "x-request-id"
+_gateway_logger = logging.getLogger(__name__)
 _SENSITIVE_ERROR_PATTERNS: tuple[tuple[re.Pattern[str], str], ...] = (
     (
         re.compile(r"(?i)\b(bearer)\s+[A-Za-z0-9._~+/=-]+"),
@@ -109,6 +111,16 @@ def _sanitize_exception_message(exc: Exception) -> str:
     if len(sanitized) > _SANITIZED_ERROR_MAX_LENGTH:
         return f"{sanitized[:_SANITIZED_ERROR_MAX_LENGTH]}...[truncated]"
     return sanitized
+
+
+def _gateway_log(level: int, message: str, **fields: object) -> None:
+    if not _gateway_logger.isEnabledFor(level):
+        return
+    if not fields:
+        _gateway_logger.log(level, message)
+        return
+    serialized_fields = orjson.dumps(fields, default=str).decode("utf-8")
+    _gateway_logger.log(level, "%s | %s", message, serialized_fields)
 
 
 async def get_provider_credentials(
@@ -184,7 +196,8 @@ async def user_api_key_auth(request: Request, api_key: str | None) -> UserAPIKey
     actually our JWT token minted by the agent executor.
     """
     if request.url.path in _UNAUTHENTICATED_HEALTH_ROUTES:
-        logger.debug(
+        _gateway_log(
+            logging.DEBUG,
             "Allowing unauthenticated LiteLLM health probe",
             path=request.url.path,
             **_load_fields(),
@@ -197,7 +210,7 @@ async def user_api_key_auth(request: Request, api_key: str | None) -> UserAPIKey
     try:
         claims = verify_llm_token(api_key or "")
     except ValueError as e:
-        logger.warning("LLM token validation failed", **_load_fields())
+        _gateway_log(logging.WARNING, "LLM token validation failed", **_load_fields())
         raise ProxyException(
             message="Invalid or expired token",
             type="auth_error",
@@ -248,7 +261,8 @@ class TracecatCallbackHandler(CustomLogger):
         model = user_api_key_dict.metadata.get("model")
         provider = user_api_key_dict.metadata.get("provider")
         if not model or not provider:
-            logger.warning(
+            _gateway_log(
+                logging.WARNING,
                 "Model or provider not found in token metadata",
                 workspace_id=workspace_id,
                 model=model,
@@ -272,7 +286,8 @@ class TracecatCallbackHandler(CustomLogger):
         )
 
         if not creds:
-            logger.warning(
+            _gateway_log(
+                logging.WARNING,
                 "No credentials configured for provider",
                 workspace_id=workspace_id,
                 provider=provider,
@@ -336,7 +351,8 @@ class TracecatCallbackHandler(CustomLogger):
         load_snapshot = _gateway_load_tracker.end_request()
         request_counter = _pop_hook_request_counter(request_data)
         sanitized_error = _sanitize_exception_message(original_exception)
-        logger.error(
+        _gateway_log(
+            logging.ERROR,
             "LiteLLM call failed",
             request_counter=request_counter,
             trace_request_id=user_api_key_dict.metadata.get("trace_request_id"),
@@ -372,8 +388,10 @@ def _inject_provider_credentials(
         case "openai":
             api_key = creds.get("OPENAI_API_KEY")
             if not api_key:
-                logger.warning(
-                    "Required credential key missing for provider", provider=provider
+                _gateway_log(
+                    logging.WARNING,
+                    "Required credential key missing for provider",
+                    provider=provider,
                 )
                 raise ProxyException(
                     message="Provider credentials incomplete",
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -359,6 +359,12 @@ services:
         TRACECAT__EXECUTOR_QUEUE: ${TRACECAT__EXECUTOR_QUEUE:-shared-action-queue}
         TRACECAT__AGENT_EXECUTOR_MAX_CONCURRENT_ACTIVITIES: ${TRACECAT__AGENT_EXECUTOR_MAX_CONCURRENT_ACTIVITIES:-1}
         TRACECAT__LLM_PROXY_READ_TIMEOUT: ${TRACECAT__LLM_PROXY_READ_TIMEOUT:-300}
+        TRACECAT__LITELLM_NUM_WORKERS: ${TRACECAT__LITELLM_NUM_WORKERS:-4}
```
daryllimyt: 4 feels slightly high as a default for docker deploys

tracecat/config.py
```diff
@@ -626,6 +626,36 @@ def _parse_auth_types() -> set[AuthType]:
...
+)
+"""Number of LiteLLM proxy workers to run per agent-executor sidecar."""
+
+TRACECAT__LITELLM_CREDENTIAL_CACHE_TTL_SECONDS = float(
+    os.environ.get("TRACECAT__LITELLM_CREDENTIAL_CACHE_TTL_SECONDS") or 60.0
+)
+"""TTL for process-local LiteLLM credential cache entries in seconds."""
+
+TRACECAT__LITELLM_HEALTHCHECK_INTERVAL_SECONDS = float(
+    os.environ.get("TRACECAT__LITELLM_HEALTHCHECK_INTERVAL_SECONDS") or 5.0
```
daryllimyt: is 5s too frequent?

tracecat/agent/runtime_services.py
```diff
@@ -24,10 +172,115 @@ async def _stream_litellm_stderr(process: asyncio.subprocess.Process) -> None:
...
+async def _monitor_litellm_health(
+    url: str = _LITELLM_READINESS_URL,
+) -> None:
+    """Continuously poll LiteLLM readiness and fail fast on repeated probe failures."""
+    timeout = httpx.Timeout(
+        connect=TRACECAT__LITELLM_HEALTHCHECK_TIMEOUT_SECONDS,
+        read=TRACECAT__LITELLM_HEALTHCHECK_TIMEOUT_SECONDS,
+        write=TRACECAT__LITELLM_HEALTHCHECK_TIMEOUT_SECONDS,
+        pool=TRACECAT__LITELLM_HEALTHCHECK_TIMEOUT_SECONDS,
+    )
```
daryllimyt: should these all use the same timeout?

daryllimyt: why use this instead of just passing the numeric value directly into the AsyncClient

tracecat/config.py
```diff
@@ -626,6 +626,36 @@ def _parse_auth_types() -> set[AuthType]:
 )
 """Read timeout for the LLM socket proxy in seconds (default: 5 minutes)."""
 
+TRACECAT__LITELLM_NUM_WORKERS = int(
+    os.environ.get("TRACECAT__LITELLM_NUM_WORKERS") or 4
+)
+"""Number of LiteLLM proxy workers to run per agent-executor sidecar."""
```
daryllimyt: validate its >0

tracecat/agent/gateway.py
```diff
@@ -58,6 +134,53 @@ async def get_provider_credentials(
...
 
+async def _get_cached_provider_credentials(
+    *,
+    cache: DualCache,
+    workspace_id: WorkspaceID,
+    organization_id: OrganizationID,
+    provider: str,
+    use_workspace_creds: bool,
+) -> tuple[dict[str, str] | None, bool]:
+    if not (hasattr(cache, "async_get_cache") and hasattr(cache, "async_set_cache")):
```
daryllimyt: when would this be true?

tracecat/agent/litellm_observability.py
```diff
@@ -0,0 +1,88 @@
...
+            peak_active_connections=self._peak_active_connections,
+            peak_active_requests=self._peak_active_requests,
+        )
+
+
+_TRACKERS: dict[str, LiteLLMLoadTracker] = {}
+_TRACKERS_LOCK = Lock()
+
+
+def get_load_tracker(name: str) -> LiteLLMLoadTracker:
```
daryllimyt, line 79: why do we need this in both the proxy and litellm side?

tracecat/agent/runtime_services.py
```diff
@@ -62,60 +328,104 @@ async def start_litellm_proxy() -> None:
...
-    async with httpx.AsyncClient(
-        timeout=httpx.Timeout(connect=2.0, read=2.0, write=2.0, pool=2.0)
-    ) as client:
+    timeout = httpx.Timeout(
+        connect=TRACECAT__LITELLM_HEALTHCHECK_TIMEOUT_SECONDS,
+        read=TRACECAT__LITELLM_HEALTHCHECK_TIMEOUT_SECONDS,
+        write=TRACECAT__LITELLM_HEALTHCHECK_TIMEOUT_SECONDS,
+        pool=TRACECAT__LITELLM_HEALTHCHECK_TIMEOUT_SECONDS,
+    )
+    async with httpx.AsyncClient(timeout=timeout) as client:
```
daryllimyt: should this all use the same value?

tracecat/agent/gateway.py
```diff
@@ -192,15 +329,72 @@ async def async_pre_call_hook(
             }
             data.update(safe_settings)
 
+        request_id, load_snapshot = _gateway_load_tracker.begin_request()
```
daryllimyt: not sure if this should be called request_id if its really the request counter value?

daryllimyt: what I had more in mind was a http header like x-request-id or something, that we would use for tracing

tracecat/agent/executor_worker.py
```diff
@@ -82,6 +101,8 @@ async def main() -> None:
 
     try:
         client = await _start_runtime_services()
+        if runtime_failure_reason is not None:
+            raise RuntimeError(runtime_failure_reason)
```
daryllimyt: if we only set runtime_failure_reason on the unhealthy callback, and set the interrupt event after that, why not just wait for the below interrupt event to fire? or is this for early return?

tracecat/agent/sandbox/llm_proxy.py
```diff
@@ -405,19 +462,104 @@ async def _forward_request(
                     return
 
         except httpx.ConnectError as e:
-            logger.error("Failed to connect to LiteLLM", error=str(e))
+            error_class = _classify_httpx_exception(e)
```
daryllimyt: why not just return the fully qualified class name + module? i think it would be clearer

daryllimyt:
[P2] Remove synchronous per-request INFO logging from LiteLLM path — agent/gateway.py:335-346
    This log, together with the new LLM token authenticated and LiteLLM call completed INFOs in the same file, now runs on every successful LiteLLM request. Our logger is configured with a direct sink=sys.stderr
    at INFO by default, so each completion now pays for multiple blocking stderr writes plus UUID/model serialization on the hot path; at any meaningful QPS this will show up directly in first-token latency and
    throughput. If we need this telemetry, it should be sampled/debug-level or exported as metrics instead of unconditional per-request INFO logs.
    
 I’d fix this by removing the unconditional INFOs at tracecat/agent/gateway.py:217, tracecat/agent/gateway.py:335, and tracecat/agent/gateway.py:358, while keeping the request tracker bookkeeping so failures
  still have request_id correlation. That matches the actual runtime cost here: INFO logs go to a synchronous sys.stderr sink by default in tracecat/logger/_logger.py:40, so steady-state auth/start/success logs
  should not live on this path.

  I’d also sweep the other request-path INFOs in the same module at tracecat/agent/gateway.py:517, tracecat/agent/gateway.py:546, and tracecat/agent/gateway.py:560, since they have the same hot-path problem for
  subsets of traffic. If telemetry still matters, make it explicit and off by default: sampled or opt-in logging guarded before field construction, or a real metric export later. I would not rely on the existing
  heartbeat in tracecat/agent/runtime_services.py:265, because that reports llm_socket_proxy, not litellm_gateway.

tracecat/agent/litellm_observability.py
jordan-umusu: The proxy load tells us how much traffic our runtimes are pushing into LiteLLM, while gateway load
tells us how much work LiteLLM is actually processing toward providers. The delta could be valuable to observe.

PR #2465, opened 2026-04-02
fix(agents): externalize Claude resume history from workflow boundaries

jordan-umusu:
## Summary by cubic
Externalized Claude SDK resume history out of workflow/activity payloads. The executor now stages a JSONL file and passes a `resume_session_path` to the runtime, cutting Temporal payload size and stabilizing resumes; added a bounded message history preview in results and the execution UI; resume is skipped when history is missing.

- **Refactors**
  - Pass `resume_source_session_id`; executor materializes history to `resume/{sdk_session_id}.jsonl` and sends `resume_session_path` (keeps legacy inline `sdk_session_data` for replay).
  - Updated `RuntimeInitPayload`, loopback, durable workflow, and Claude runtime to read staged files, accept `resume/{id}.jsonl` or `/work/resume/{id}.jsonl`, and validate paths/IDs. NSJail now mounts the executor job dir at `/work` and cleans it up.
  - Activity returns a truncated `messages` preview (byte/count bounded) for workflow results; on approval requests, it returns early without messages. Frontend labels the preview and documents it in `AgentOutput`.

- **Bug Fixes**
  - Skip resume when persisted history can’t be materialized.
  - Validate `sdk_session_id` and resume path to prevent bad IDs/path traversal.
  - Add progress heartbeats during history materialization (per batch) to avoid timeouts.

<sup>Written for commit c4274bbebb464cbb7c9dc8b3a378f8ac22edfc0b. Summary will update on new commits.</sup>


jordan-umusu:
@cubic re-review

tracecat/agent/session/service.py
```diff
@@ -610,13 +622,7 @@ async def load_session_history(
             line = orjson.dumps(entry.content).decode("utf-8")
```
daryllimyt: this for loop can block temporal's event loop because run_agent_activity is async, run this including the "\n".join() in separate thread

tracecat/agent/session/activities.py
```diff
@@ -66,7 +66,7 @@ class LoadSessionResult(BaseModel):
 
     found: bool
     sdk_session_id: str | None = None
-    sdk_session_data: str | None = None
+    resume_source_session_id: uuid.UUID | None = None
```
daryllimyt: since we don't have Temporal worker versioning, this would cause a breaking change for running DurableAgentWorkflows (workflows in progress or replay). We should support both paths instead of hard replacement

tracecat/agent/common/protocol.py
```diff
@@ -42,7 +38,7 @@ class RuntimeInitPayload:
     # Resolved tool definitions (orchestrator resolves action names → full definitions)
     allowed_actions: dict[str, MCPToolDefinition] | None = None
     sdk_session_id: str | None = None
-    sdk_session_data: str | None = None  # JSONL content for resume
+    resume_session_path: str | None = None  # Runtime-visible staged JSONL file path
```
daryllimyt, line 44: prefer to expand the schema instead of hard replacing (deprecate over time)

tracecat/agent/executor/activity.py
```diff
@@ -62,9 +91,9 @@ class AgentExecutorInput(BaseModel):
     )
     # Resolved tool definitions
     allowed_actions: dict[str, MCPToolDefinition] | None = None
-    # Session resume data (from previous run, includes tool_result for approval flow)
+    # Session resume descriptor (JSONL materialized locally by this activity)
     sdk_session_id: str | None = None
-    sdk_session_data: str | None = None
+    resume_source_session_id: uuid.UUID | None = None
```
daryllimyt, line 163: breaking

daryllimyt (changes requested):
some operational safety concerns due to hard schema replacement

jordan-umusu:
@cubic re-review

tracecat/agent/executor/activity.py
daryllimyt: @jordan-umusu maybe we should change the mental model - passing in a session id to a DurableAgentWorkflow should signal that this is what we will continue from. Then the workflow gets its session id through its temporal workflow id

tracecat/agent/executor/activity.py
```diff
@@ -486,6 +594,97 @@ async def _create_job_directory(self) -> Path:
...
+                sdk_session_id=self.input.sdk_session_id,
+            )
+            return None
+
+        try:
+            validated_session_id = _validate_sdk_session_id(self.input.sdk_session_id)
+        except AgentSandboxValidationError as exc:
+            raise AgentSandboxExecutionError(str(exc)) from exc
+
+        resume_dir = self._job_dir / "resume"
```
daryllimyt, line 663: what is this resume directory used for?

daryllimyt (commented):
-clarification on /resume
-we might want to shift the mental model to mean: "when i pass a session id into an agent workflow it signals that this passed session id is the starting point"

daryllimyt:
@jordan-umusu im going to take this on

daryllimyt:
Closing this in favor of a more targeted PR that prevents passing Claude session data to/from Temporal.

daryllimyt:
Closing this in favor of a more targeted PR that prevents passing Claude session data to/from Temporal.

PR #2518, opened 2026-04-11
feat(ui+api): Rename inbox to approvals

topher-lo:
## Summary
This changes the workspace sidebar so approval work is grouped under the existing `Actions` section instead of appearing as a separate top-level inbox entry. It also removes the standalone `Add case` button from the sidebar.

## User impact
Before this change, the sidebar exposed `Inbox` as its own top-level item and also showed `Add case` as a global action. That made the workspace navigation feel split between top-level shortcuts and accordion content, even though approvals belong with other action-oriented workspace tools.

After this change, users find approvals under `Workspace` -> `Actions` as `Approvals`, alongside the existing actions registry entry when that scope is available. The redundant top-level case creation entry is removed.

## Root cause
The sidebar defined inbox and case creation in a separate group above the workspace accordion, while the rest of the workspace navigation lived inside `navWorkspace`. That structure prevented approvals from being organized with related workspace tools.

## Fix
The sidebar now models `Actions` as a grouped navigation item with sub-items:
- `Registry` for the existing actions page when the registry scope is available.
- `Approvals` for the inbox route when inbox access is available.

The obsolete top-level `Add case` button and its unused dialog wiring were removed from the sidebar component.

## Validation
I validated the changed file with:
- `pnpm -C frontend exec biome check src/components/sidebar/app-sidebar.tsx`

I also installed frontend dependencies in this worktree so the packaged frontend tooling was available.

---
## Summary by cubic
Moved approvals into the Sidebar’s Actions group, replaced the old Inbox surface with Approvals across UI and API, and removed the Add case button. This tidies navigation and standardizes the approvals flow.

- **Refactors**
  - UI: Added nested Actions items “Registry” and “Approvals”; routes/pages renamed to `/approvals`; removed Inbox and the sidebar “Add case” wiring.
  - API/scopes: `inbox:read` -> `approval:read`; `GET /inbox/items` -> `GET /approvals`; submit approvals is now `PATCH /approvals/{session_id}`; `ContinueRunRequest.source` default is `"approval"`.

<sup>Written for commit 965103812cd1cd5003a256863ac066fa222ee10e. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/app/workspaces/[workspaceId]/approvals/layout.tsx
+++ b/frontend/src/app/workspaces/[workspaceId]/approvals/layout.tsx
@@ -3,39 +3,41 @@
 import { XIcon } from "lucide-react"
 import type React from "react"
 import { createContext, useCallback, useContext, useRef, useState } from "react"
-import { InboxChat } from "@/components/inbox/inbox-chat"
+import { ApprovalsChat } from "@/components/approvals/approvals-chat"
 import { ControlsHeader } from "@/components/nav/controls-header"
 import { AppSidebar } from "@/components/sidebar/app-sidebar"
 import { Button } from "@/components/ui/button"
 import { ResizableSidebar } from "@/components/ui/resizable-sidebar"
 import { SidebarInset, SidebarProvider } from "@/components/ui/sidebar"
-import type { InboxSessionItem } from "@/lib/agents"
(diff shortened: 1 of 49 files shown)


topher-lo:
@codex review

PR #2566, opened 2026-04-24
feat(agents): llm providers v2 migration

jordan-umusu:
## Summary by cubic
Backfills orgs into the LLM Providers v2 catalog without decrypting secrets, seeds the platform catalog, and ships a control plane for catalog/custom providers, access, and default model selection. Adds tenant RLS, capability flags, and stricter passthrough to prep broader model management.

- **New Features**
  - Control plane for catalog CRUD, custom provider CRUD, and model access (org/workspace), with cursor pagination and audit logs.
  - Default model selection via GET/PUT endpoints using `catalog_id`.
  - Platform catalog bumped with capability flags (`supports_native_streaming`, `supports_parallel_function_calling`) and stricter passthrough validation; tenant RLS enforced; dropped custom-provider discovery status.

- **Migration**
  - Copies ciphertext from `organization_secret` `agent-{provider}-credentials` to `encrypted_config` on catalog/custom-provider rows (no decryption).
  - Seeds platform catalog from `tracecat/agent/platform_catalog.json`.
  - Cloud providers (`bedrock`, `azure_openai`, `azure_ai`, `vertex_ai`): one org-scoped catalog row (`model_name` = provider) plus an org-wide access grant.
  - Direct providers (`openai`, `anthropic`, `gemini`): org gets access to all matching platform catalog rows; no org catalog rows.
  - Custom provider: creates an `agent_custom_provider` and linked catalog row (`custom-model-provider`/`custom`) with org-wide access.
  - Backfills `catalog_id` on presets and versions; prefers platform rows over org rows when both match.
  - Resolves `agent_default_model` to `agent_default_model_catalog_id`.
  - Idempotent, expansion-only; downgrade is a no-op.

<sup>Written for commit e66558622dde8feedd029a09d0f231cf1e72fce1. Summary will update on new commits. <a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2566?utm_source=github">Review in cubic</a></sup>


jordan-umusu:
@cubic re-review

daryllimyt (commented):
The data migration can fail or backfill incorrect data for supported existing database states, notably duplicate secret environments and pre-existing custom providers.

Full review comments:

- [P1] Filter provider secrets by environment — `alembic/versions/96470fdcc686_v2_backfill_catalog_and_access.py:307-310`
  When an organization has the same agent credential name in multiple secret environments, this lookup ignores `organization_secret.environment`, so `.one_or_none()` raises and blocks the whole migration; if only a non-default environment exists, the migration can also copy credentials/grant access that the agent service would not use because runtime lookups default to the `default` environment. Add the environment column and filter these provider-secret queries to the default environment.

- [P1] Do not reuse arbitrary custom providers — `alembic/versions/96470fdcc686_v2_backfill_catalog_and_access.py:436-440`
  For orgs that already have custom-provider rows when this migration runs, this unscoped lookup either raises `MultipleResultsFound` if there is more than one provider or attaches the legacy `custom-model-provider` catalog entry to an unrelated existing provider without copying the legacy ciphertext onto that provider. The backfill needs to find a migration-created/provider-specific row or create a dedicated one for the legacy secret.

alembic/versions/96470fdcc686_v2_backfill_catalog_and_access.py
```diff
@@ -0,0 +1,691 @@
...
+    stmt = stmt.on_conflict_do_update(
+        index_elements=[
+            "organization_id",
+            "custom_provider_id",
+            "model_provider",
+            "model_name",
+        ],
+        set_={
+            "model_metadata": stmt.excluded.model_metadata,
+            "last_refreshed_at": stmt.excluded.last_refreshed_at,
```
daryllimyt, line 310: [P1] Filter provider secrets by environment

When an organization has the same agent credential name in multiple secret environments, this lookup ignores `organization_secret.environment`, so `.one_or_none()` raises and blocks the whole migration; if only a non-default environment exists, the migration can also copy credentials/grant access that the agent service would not use because runtime lookups default to the `default` environment. Add the environment column and filter these provider-secret queries to the default environment.

alembic/versions/96470fdcc686_v2_backfill_catalog_and_access.py
```diff
@@ -0,0 +1,691 @@
...
+                agent_catalog_tbl.c.custom_provider_id.is_(None),
+                agent_catalog_tbl.c.model_provider == provider,
+                agent_catalog_tbl.c.model_name == provider,
+            )
+        ).one()
+        catalog_id = existing.id
+    _grant_org_access(conn, org_id=org_id, catalog_id=catalog_id)
+
+
+def _migrate_custom_provider(conn: sa.engine.Connection, *, org_id: uuid.UUID) -> None:
```
daryllimyt, line 440: [P1] Do not reuse arbitrary custom providers

For orgs that already have custom-provider rows when this migration runs, this unscoped lookup either raises `MultipleResultsFound` if there is more than one provider or attaches the legacy `custom-model-provider` catalog entry to an unrelated existing provider without copying the legacy ciphertext onto that provider. The backfill needs to find a migration-created/provider-specific row or create a dedicated one for the legacy secret.

alembic/versions/96470fdcc686_v2_backfill_catalog_and_access.py
jordan-umusu: This is technically possible, but not through the typical agent credential flow so not relevant to the one-time migration for this exact flow.

alembic/versions/96470fdcc686_v2_backfill_catalog_and_access.py
jordan-umusu: This isn't possible. Pre-migration users can only configure a singular setting for custom providers/

daryllimyt (approved):

PR #2589, opened 2026-04-30
fix: Preserve legacy case duration anchors

daryllimyt:
## Summary

- Restore required-filter validation for newly authored case duration anchors so empty required filters are rejected on create/update payloads.
- Preserve legacy stored definitions with empty filters by treating them as match-any anchors without relaxing public API validation.
- Keep SQL-compatible legacy empty-filter anchors on the SQL path, and reserve the event-scan fallback for custom timestamp paths or genuinely unsupported legacy filters.
- Add cooperative checkpoints to the legacy fallback scan so compatibility work yields back to the event loop during large in-memory scans.
- Tighten the opt-in case-duration burst benchmark so burst healthcheck errors fail the test.

## Review Map

Before the original perf change in `2dec4719e` / #2558, duration computation looked roughly like this:

```text
load definitions
load case events ordered by created_at
for each definition:
  start = scan events with anchor.event_type + anchor.field_filters
  end = scan events after start with anchor.event_type + anchor.field_filters
  timestamp = resolve anchor.timestamp_path from the event
  sort candidates by resolved timestamp
```

That old path was slower under bursts because every computation scanned ORM-loaded events in Python, but it also had two broad legacy behaviors:

```text
field_filters = {}            -> match any event of the configured type
timestamp_path = data.foo     -> use event.data.foo, not event.created_at
```

After #2558, the normal path became SQL-first and much faster:

```text
for each definition:
  query first/last start match in SQL
  query first/last end match in SQL with per-case start bound
```

The regression risk was that SQL only understood the newer typed filter model and `created_at` anchors. Stored legacy rows with `{}` filters or custom timestamp paths could therefore stop matching the way they did before.

This PR keeps the #2558 SQL fast path, and adds a compatibility split:

```text
typed filters + created_at
  -> SQL fast path

legacy {} filters + created_at
  -> SQL fast path as match-any event type

custom timestamp path or unsupported legacy filters
  -> event-scan compatibility fallback
```

## Compatibility Notes

This PR separates new payload validation from legacy stored-row reconstruction:

- New create/update payloads still reject empty required filters for `status_changed`, priority/severity changes, tag events, field changes, and dropdown value changes.
- Stored legacy rows with `{}` filters are reconstructed with an internal compatibility marker so they continue to match any event of the configured type.
- Stored rows with custom timestamp paths, such as `data.resolved_at`, continue to use that extracted timestamp for first/last selection instead of silently falling back to `created_at`.

## Performance Notes

The main SQL fast path remains unchanged for normal typed-filter definitions.

The fallback path is now narrower and cheaper:

- Empty legacy filters with `created_at` stay on SQL because SQL can already express “match any event of this type.”
- Python event scanning is reserved for custom timestamp paths or legacy filters that cannot be represented by the typed SQL filters.
- The remaining event scan selects the best match incrementally instead of collecting and sorting all candidate events.
- For `created_at`-ordered anchors, the scanner can return early for first-match scans and scan in reverse for last-match scans.
- The fallback stays on the event loop because it works with ORM-loaded models, but it uses `cooperative_every` checkpoints to avoid monopolizing the API loop during rare large legacy scans.

## Validation

- `uv run pytest tests/unit/test_case_duration_service.py tests/unit/test_concurrency.py`
- `TRACECAT__SERVICE_KEY=test-service-key uv run pytest tests/unit/test_case_duration_service.py`
- `uv run ruff check tracecat/cases/durations/service.py tracecat/cases/durations/schemas.py tests/unit/test_case_duration_service.py tests/integration/test_case_duration_benchmarks.py`
- `uv run basedpyright tracecat/cases/durations/service.py tracecat/cases/durations/schemas.py tests/unit/test_case_duration_service.py tests/integration/test_case_duration_benchmarks.py`
- `uv run ruff format --check tracecat/cases/durations/service.py tracecat/cases/durations/schemas.py tests/unit/test_case_duration_service.py tests/integration/test_case_duration_benchmarks.py`
- `TRACECAT_RUN_CASE_DURATION_BENCHMARKS=1 TRACECAT_CASE_DURATION_BENCHMARK_OUTPUT=/tmp/tracecat-case-duration-benchmark-after-fix.json uv run pytest tests/integration/test_case_duration_benchmarks.py -s`

Benchmark run: burst health errors `0`, burst health p95 `135.3ms`, max `219.2ms`.

## Additional QA - April 30, 2026

Last-mile QA was run against the current branch in an isolated local Tracecat stack:

- Started `just cluster up -d` for worktree `codex-case-duration-compat`, cluster `codex-case-duration-compat-2`.
- Used local API `http://localhost:180/api` and seeded workspace `299ec1da-4ed0-4418-b26e-51b119d8ad2f`.
- Created a temporary workspace service account with `case:read`, `case:create`, `case:update`, and `case:delete`, then used its bearer key for the API checks.
- Verified normal duration computation via API: `case_created` -> `status_changed` filtered to `resolved` computed and persisted.
- Verified status alias behavior via API:
  - `status_changed` filtered to `closed` matched a `case_closed` event.
  - `status_changed` filtered to `in_progress` matched a `case_reopened` event.
- Verified dropdown duration filters via API by creating a temporary dropdown, setting a case dropdown value, and confirming `dropdown_value_changed` with `dropdown_definition_id` + `dropdown_option_ids` computed.
- Verified public API validation rejects newly authored incompatible anchors:
  - Empty required `status_changed` filters returned `422`.
  - Custom `timestamp_path` returned `422`.
- Verified legacy DB compatibility through API-triggered sync:
  - Stored legacy `{}` status filters still matched any status change.
  - Stored legacy `data.resolved_at` timestamp path used the extracted timestamp and selected by custom timestamp order, not `created_at` order.
  - Stored arbitrary legacy field filter `data.items.name` used the event-scan fallback and matched the expected event.
- Cleanup completed: 7 temporary cases deleted, 7 temporary duration definitions deleted, 1 temporary dropdown definition deleted, and temporary service account `8964e7d6-2576-470d-9544-5e38e8d0752c` disabled.
- Post-cleanup DB check found 0 remaining QA cases, 0 duration definitions, 0 dropdown definitions, and 0 active QA service accounts for run `codex-qa-0f6e96ea`.
- Stopped the local cluster with `just cluster down` without removing volumes.

Additional verification from this QA pass:

- `uv run pytest tests/unit/test_case_duration_service.py` -> 36 passed.
- `TRACECAT_RUN_CASE_DURATION_BENCHMARKS=1 TRACECAT_CASE_DURATION_BENCHMARK_OUTPUT=/tmp/case-duration-benchmark-codex-qa.json uv run pytest tests/integration/test_case_duration_benchmarks.py -s` -> 3 passed.
- Benchmark run: burst health errors `0`, burst health p95 `62.8ms`, max `125.8ms`; update errors `0`; update p95/max `686.0ms`.
- `uv run ruff check tracecat/cases/durations/schemas.py tracecat/cases/durations/service.py tests/unit/test_case_duration_service.py tests/integration/test_case_duration_benchmarks.py`
- `uv run basedpyright tracecat/cases/durations/schemas.py tracecat/cases/durations/service.py tests/unit/test_case_duration_service.py tests/integration/test_case_duration_benchmarks.py`
- `git diff --check main...HEAD`

### DIFF
--- a/tracecat/cases/durations/service.py
+++ b/tracecat/cases/durations/service.py
@@ -50,6 +50,7 @@ from tracecat.exceptions import (
     TracecatValidationError,
 )
 from tracecat.service import BaseWorkspaceService, requires_entitlement
+from tracecat.tables.common import coerce_to_utc_datetime
 from tracecat.tiers.enums import Entitlement
 
 type _DurationEventMatch = tuple[uuid.UUID, datetime]
@@ -198,6 +199,7 @@ class CaseDurationDefinitionService(BaseWorkspaceService):
         raw_filters = dict(getattr(entity, f"{prefix}_field_filters") or {})
         event_type = getattr(entity, f"{prefix}_event_type")
         selection = getattr(entity, f"{prefix}_selection")
+        timestamp_path = getattr(entity, f"{prefix}_timestamp_path") or "created_at"
         filters, has_unsupported_filters = self._filters_from_storage(
             event_type, raw_filters
         )
@@ -222,6 +224,8 @@ class CaseDurationDefinitionService(BaseWorkspaceService):
                     selection=selection,
                 )
         anchor._has_unsupported_filters = has_unsupported_filters
+        anchor._timestamp_path = timestamp_path
+        anchor._legacy_field_filters = raw_filters
         return anchor
 
     def _anchor_attributes(
@@ -230,7 +234,7 @@ class CaseDurationDefinitionService(BaseWorkspaceService):
         filters = self._filters_to_storage(anchor.filters)
         return {
             f"{prefix}_event_type": anchor.event_type,
-            f"{prefix}_timestamp_path": "created_at",
+            f"{prefix}_timestamp_path": anchor._timestamp_path,
             f"{prefix}_field_filters": filters,
             f"{prefix}_selection": anchor.selection,
         }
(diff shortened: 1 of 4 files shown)


tracecat/cases/durations/service.py
```diff
@@ -660,6 +714,231 @@ async def _compute_durations_from_db_for_cases(
...
+                if actual_normalized is None:
+                    return False
+                if isinstance(actual_normalized, list):
+                    if not any(
+                        item in expected_normalized for item in actual_normalized
+                    ):
+                        return False
+                elif actual_normalized not in expected_normalized:
+                    return False
+            elif actual_normalized != expected_normalized:
```
jordan-umusu, line 902: resolve_field for lists returns a scalar and then compares the raw value against the scalar and will reject a correctly matching event

jordan-umusu (approved):
LGTM once this issue is fixed

PR #2644, opened 2026-05-07
fix(cases): idempotent link rows

jordan-umusu:
## Summary by cubic
Make linking case rows idempotent by returning the existing link when a duplicate insert happens. This prevents duplicate link rows and avoids extra events.

- **Bug Fixes**
  - Catch `sa_exc.IntegrityError` on `flush`; rollback, fetch via `dedupe_stmt`, and return the existing link.
  - Skip event creation when the link already exists.

<sup>Written for commit 4141116097aa04efc3382889ef6bbe117b152d05. Summary will update on new commits.</sup>

### DIFF
--- a/tracecat/cases/rows/service.py
+++ b/tracecat/cases/rows/service.py
@@ -200,6 +200,14 @@ class CaseTableRowsService(BaseWorkspaceService):
             row_id=params.row_id,
         )
         self.session.add(link)
+        try:
+            await self.session.flush()
+        except sa_exc.IntegrityError:
+            await self.session.rollback()
+            existing = (await self.session.execute(dedupe_stmt)).scalars().first()
+            if existing is None:
+                raise
+            return existing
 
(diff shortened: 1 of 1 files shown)


tracecat/cases/rows/service.py
```diff
@@ -200,6 +200,14 @@ async def link_row(
             row_id=params.row_id,
         )
         self.session.add(link)
+        try:
+            await self.session.flush()
+        except sa_exc.IntegrityError:
+            await self.session.rollback()
```
daryllimyt, line 206: because this session is shared over the whole service, would this rollback more than just the link addition?

daryllimyt (changes requested):

tracecat/cases/rows/service.py
jordan-umusu: no, its doing read only queries up until this point, this flush is the first write attempt.

PR #2645, opened 2026-05-07
fix(cases): make row links idempotent

daryllimyt:
## Summary
- make `CaseTableRowsService.link_row` return an existing link for duplicate case/table/row input
- use `ON CONFLICT DO NOTHING` against `uq_case_table_row_link` to avoid duplicate-key 500s under retry/race conditions
- add a regression test that duplicate linking creates one row and returns the original link

## Verification
- `uv run pytest tests/unit/test_case_table_rows_service.py tests/unit/api/test_api_case_rows.py`
- `uv run ruff check tracecat/cases/rows/service.py tests/unit/test_case_table_rows_service.py`
- `uv run basedpyright tracecat/cases/rows/service.py tests/unit/test_case_table_rows_service.py`

---
## Summary by cubic
Make case row linking idempotent to prevent duplicates and 500s under retries or races, including when link limits are reached. Duplicate case/table/row requests now return the existing link instead of erroring.

- **Bug Fixes**
  - Updated `CaseTableRowsService.link_row` to use Postgres `insert(...).on_conflict_do_nothing` on `uq_case_table_row_link` and to return the existing link even when `MAX_LINKED_ROWS_PER_CASE` or `MAX_TABLES_PER_CASE` are hit.
  - Added regression tests for duplicate linking and limit-reached race scenarios to ensure only one `CaseTableRow` is created and the original link is returned.

<sup>Written for commit 11b1c04827653c7bd4d8313d87fa1c566b20538a. Summary will update on new commits.</sup>

### DIFF
--- a/tests/unit/test_case_table_rows_service.py
+++ b/tests/unit/test_case_table_rows_service.py
@@ -9,6 +9,7 @@ from tracecat.auth.types import Role
 from tracecat.cases.enums import CasePriority, CaseSeverity, CaseStatus
 from tracecat.cases.rows.schemas import CaseTableRowLinkCreate
 from tracecat.cases.rows.service import (
+    MAX_LINKED_ROWS_PER_CASE,
     MAX_TABLES_PER_CASE,
     CaseTableRowsService,
 )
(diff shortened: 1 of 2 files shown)


jordan-umusu:
LGTM

jordan-umusu (approved):

PR #2646, opened 2026-05-07
fix(api): map known database errors

daryllimyt:
## Summary
- map the duplicate case-row link database error to a structured 409 response
- map the observed DB value type mismatch to a structured 422 response
- keep the existing generic 500 handler for all other unexpected exceptions

## Verification
- `uv run pytest tests/unit/api/test_api_common.py`
- `uv run ruff check tracecat/api/common.py tests/unit/api/test_api_common.py`
- `uv run basedpyright tracecat/api/common.py tests/unit/api/test_api_common.py`

---
## Summary by cubic
Map known database errors to structured API responses so clients get 409/422 instead of a generic 500. The 409 case–row link conflict is now handled via a small helper in both link routes; unexpected errors still return 500.

- **Bug Fixes**
  - Map `IntegrityError` for duplicate case–row link (constraint `uq_case_table_row_link`) to 409 with code CASE_ROW_ALREADY_LINKED and roll back the session; applied in both public and internal routes.
  - Map database value type mismatch to 422 with code DATABASE_VALUE_TYPE_MISMATCH via the generic exception handler.
  - Add tests for public and internal link endpoints to cover the 409 response and rollback.

<sup>Written for commit 6a87a2031c853cf7228bfda5a36f7a5312b9e399. Summary will update on new commits.</sup>


tracecat/api/common.py
```diff
@@ -26,7 +27,47 @@
...
 
 
+def _exception_text(exc: Exception) -> str:
+    orig = getattr(exc, "orig", None)
+    if orig is None:
+        return str(exc)
+    return f"{exc} {orig}"
+
+
+def _known_database_error_response(request: Request, exc: Exception) -> Response | None:
```
jordan-umusu: move this change to the link case row router (external/internal) instead of catching at the generic

jordan-umusu (changes requested):

jordan-umusu (approved):

PR #2657, opened 2026-05-08
feat(agent): expose system prompt overrides on custom providers

### LINKED ISSUE #2654: feat: expose system prompt overrides on `ai.action` and custom sources
### Update — scope clarification (post-investigation)

While prototyping this against the codebase, we found that the action-level schema for `ai.action` (`AgentActionArgs`) lives in `packages/tracecat-ee/tracecat_ee/agent/schemas.py`, not in the OSS tree. The custom-source-level part is fully OSS, but exposing `system_prompt_replace` / `system_prompt_append` on the `ai.action` form requires extending that EE schema.

We don't intend to ship code into the EE module ourselves. Two viable paths:

1. We deliver the OSS portion (DB column on `agent_custom_provider`, schemas, runtime cascade with defensive `getattr` for the action-level fields). Maintainers add the two fields to `AgentActionArgs` in EE at merge time. The OSS code is forward-compatible — when EE exposes the fields, the cascade picks them up automatically.
2. Maintainers tell us where they'd prefer the schema to live, and we adjust before opening the PR.

Happy to follow whichever path you prefer.

---

### TL;DR

`ai.action` always prepends a hardcoded line to the system prompt:

```
"If asked about your identity, you are a Tracecat automation assistant."
```

The user-facing `Instructions` field can only **append** more text — never replace this baseline. There is no way for a workflow author to take full control of the system prompt, neither at the action level nor at the custom-source level.

This issue proposes to expose two new optional fields (`system_prompt_replace`, `system_prompt_append`) at both levels, mirroring capabilities **already present** in `claude-agent-sdk==0.1.51`. Default `None` everywhere → zero behaviour change for existing workflows.

### Why it matters

Self-hosted operators routinely want their agent to identify as their own product, or to ground it for a domain-specific role. Tested against `qwen3.5:9b` via Ollama, same user prompt *"Who are you? Reply in one sentence."*:

| Setup | Model reply |
|---|---|
| Default (Tracecat baseline) | "I'm a Tracecat automation assistant built on Anthropic's Claude Agent SDK." |
| `--system-prompt "You are Qwen, by Alibaba."` | "I'm Qwen, an AI model developed by Alibaba..." |

The baseline injection is a deliberate Tracecat choice — fine — but it should be overridable.

### Capability already in the SDK

`ClaudeAgentOptions` natively supports:

- `system_prompt: str` → `--system-prompt <s>` on the bundled CLI 2.1.85 (full replace).
- `system_prompt={"type":"preset","preset":"claude_code","append":"..."}` → `--append-system-prompt`.

The proposal is to surface these in the Tracecat schema, not to invent new logic.

### Proposed change

Two new optional fields, exposed at two levels.

#### Custom source (defaults)

| Field | Type | Default |
|---|---|---|
| `system_prompt_replace` | `str \| None` | `None` |
| `system_prompt_append` | `str \| None` | `None` |

#### `ai.action` schema (per-call overrides)

Same two fields, with the same semantics. When set, they take precedence over the source-level defaults.

#### Cascade rules (resolved in `dsl/action.py:build_agent_args_activity`)

```
final_replace = action.replace or source.replace or None
final_append  = "\n\n".join(filter(None, [
                    source.append,
                    action.append,
                    action.instructions,   # legacy field, kept for backward compat
                ]))

if final_replace is not None:
    system_prompt = final_replace
else:
    system_prompt = base_tracecat_default
if final_append:
    system_prompt = f"{system_prompt}\n\n{final_append}"
```

### Backward compatibility

- All new fields default to `None`. **Zero behaviour change** for existing workflows or custom sources.
- The legacy `Instructions` field on `ai.action` is preserved and treated as an append-mode contributor; documented as a soft alias of `system_prompt_append`.
- The default Tracecat baseline string is unchanged.

### Webhook usage (bonus)

Because all string fields accept Tracecat template expressions, this also enables passing the system prompt at trigger time:

```bash
curl -X POST .../webhooks/wf_xxx/secret \
  -d '{
        "user_prompt": "Hello",
        "system_prompt": "You are a domain-specific assistant for our team."
      }'
```

…with the action wired as `system_prompt_replace: ${{ TRIGGER.system_prompt }}`.

### Scope of implementation

- `tracecat/db/models.py`: two `Text NULL` columns on `agent_custom_provider`.
- Alembic migration.
- `tracecat/agent/provider/schemas.py`: extend Create/Update/Read.
- Registry action schema for `ai.action`: add the two fields.
- `tracecat/dsl/action.py:build_agent_args_activity`: implement cascade.
- `tracecat/agent/runtime/claude_code/runtime.py`: revise `_build_system_prompt`.
- `tracecat/agent/runtime/pydantic_ai/runtime.py`: honor `system_prompt_replace`/`append` on the `instructions` kwarg.
- Frontend: types regenerated via `just gen-client-ci`. Custom source modal: add the two fields under a collapsible "Agent settings" group, next to Passthrough mode.
- Tests: cascade matrix, runtime mapping, regression on existing `Instructions` behaviour.
- Docs: update `docs/agents/ai-action.mdx` parameters section.

### Open question for maintainers

Naming: `system_prompt_replace` / `system_prompt_append` was preferred over a single `system_prompt_mode: Literal["append","replace","bare"]` enum because it maps 1:1 to what the SDK already does. Open to a different naming if there's a precedent in the codebase.

Happy to send the PR. Could also split source-level and action-level into two PRs if reviewers prefer.

eviltik:
Fixes #2654

# Expose system prompt overrides on custom providers

## Why

Today every `ai.action` invocation goes through `claude_code/runtime.py:_build_system_prompt`, which prepends a hardcoded baseline (`"If asked about your identity, you are a Tracecat automation assistant."`) to the system prompt sent to the model. The user-facing `Instructions` field on `ai.action` only **appends** to this baseline — there is no way to:

- replace the baseline (e.g. so the agent identifies as the operator's own product, or for a domain-specific role);
- drop the baseline (e.g. when the upstream model already has a strong identity);
- configure the override **once per backend** (custom source) so every workflow that uses the source inherits it.

The capability is already in `claude-agent-sdk==0.1.51`: `ClaudeAgentOptions.system_prompt: str` translates to `--system-prompt <s>` on the bundled Claude Code CLI 2.1.85 (full replace), and the `{"type":"preset","preset":"claude_code","append":"..."}` shape maps to `--append-system-prompt`. This PR surfaces those capabilities at the Tracecat custom-source level.

**No behaviour change for existing workflows or sources.** All new fields default to `None`; the existing baseline + per-action `Instructions` flow is preserved unchanged.

## What's in this PR

Five commits, individually reviewable:

1. **`feat(agent): expose system prompt overrides on custom providers`**
   - `tracecat/db/models.py`: two `Text NULL` columns on `agent_custom_provider`.
   - `alembic/versions/c969b5f63428_…`: nullable migration, no backfill needed.
   - `tracecat/agent/provider/schemas.py`: `system_prompt_replace` and `system_prompt_append` exposed on Create / Update / Read.
   - `tracecat/agent/provider/service.py`: propagated into the create / update CRUD path.
   - `tracecat/agent/provider/activities.py` (new): OSS Temporal activity `resolve_custom_provider_overrides_activity` that maps a `catalog_id` to the source-level overrides via the `agent_catalog`/`agent_custom_provider` join. Returns an empty result when the catalog row is not backed by a custom provider, so the cascade gracefully falls back to defaults.
   - `tracecat/dsl/workflow.py` (`AI_ACTION` case): calls the new activity when `catalog_id` is set, then resolves the cascade. Action-level fields are read defensively via `getattr` because `AgentActionArgs` (defined in the EE schema module) does not yet expose them — the cascade is forward-compatible the day EE adds the matching fields.
   - `tracecat/agent/types.py` and `tracecat/agent/common/types.py`: `AgentConfig` and `SandboxAgentConfig` carry the resolved values to the runtime layer (added to `from_dict`, `from_agent_config`, `to_dict` for orjson serialization through the sandbox boundary).
   - `tracecat/agent/runtime/claude_code/runtime.py`: `_build_system_prompt` now honours `system_prompt_replace` / `system_prompt_append`. The default baseline is preserved for callers that don't set `replace`. An empty-string `replace` is honoured as a deliberate "no baseline" override (it's not silently mapped to the default).
   - `tracecat/agent/runtime/pydantic_ai/runtime.py`: pydantic-ai exposes a single `instructions=` kwarg with no implicit baseline. The runtime collapses `replace` + legacy `instructions` + `append` into one string in cascade order.
   - `tracecat/agent/internal_router.py`: `/internal/agent/run` propagates the new fields to `runtime_run_agent`.
   - `tracecat/agent/worker.py`: registers the new activity on the agent worker.
   - `tracecat/agent/schemas.py`: `AgentConfigSchema` (the request shape for `/internal/agent/run`) extended.

2. **`test(agent): unit cover system-prompt cascade and runtime resolvers`**
   - Pulls the cascade out of `dsl/workflow.py` into a pure helper `tracecat.agent.provider.cascade.resolve_system_prompt_overrides`, returning a frozen `ResolvedSystemPromptOverrides` dataclass that also exposes diagnostic metadata (`replace_source`, `append_count`).
   - Pulls pydantic-ai's collapse logic into a module-level `_collapse_instructions(replace, instructions, append)`.
   - 31 unit tests across 3 modules:
     - `test_agent_provider_cascade.py` — action > source > default precedence, append cumulation order, empty-string replace honoured, immutability of the result.
     - `test_agent_runtime_system_prompt.py` — `_build_system_prompt` matrix covering legacy `instructions`, `output_type`, `system_prompt_replace`, `system_prompt_append`, and the legacy field appended last for backward compatibility.
     - `test_pydantic_ai_runtime_collapse.py` — collapse behaviour with empty replace, append after replace, append-only, etc.

3. **`docs(agents): add system prompt overrides guide`**
   - New page `docs/agents/system-prompt-overrides.mdx` walking through when to use replace vs append, how to configure on a custom source, the resolution order, webhook trigger usage with `${{ TRIGGER.system_prompt }}`, and backward compatibility.
   - Cross-link wired into `scripts/generate_ai_docs.py` so the next regen of `ai-action.mdx` picks it up. The regenerated `ai-action.mdx` is intentionally **not** in this PR — running the generator locally also surfaces unrelated diff churn (the EE registry layout influences the output ordering), which should not ship in this PR.
   - `docs/docs.json` entry added.

4. **`feat(frontend): expose system prompt overrides on the custom source modal`**
   - `frontend/src/components/organization/org-settings-agent.tsx`: two new `Textarea` fields (`System prompt replace`, `System prompt append`) added to the custom-source dialog under the existing `Passthrough mode` toggle. Schema, defaults, and Create / Update payload builders extended to carry the new values.
   - The dialog now uses a flex column layout with `max-h-[90vh]` and an inner scroll container so the form stays usable on smaller viewports now that the body is taller.
   - `frontend/src/client/types.gen.ts`: the manual edit mirrors what `just gen-client-ci` will produce once the backend changes land — kept here so the branch is testable end-to-end without a separate regen step.

5. **`fix(agent): register provider overrides activity on the DSL worker`**
   - `tracecat/dsl/worker.py`: registers `resolve_custom_provider_overrides_activity` on the DSL worker. The activity is invoked from `dsl/workflow.py:AI_ACTION` (DSL worker), not the agent worker, so without this every `ai.action` errored with `Activity function ... is not registered on this worker` until the registration was added on both workers.

## Resolution order

The cascade lives in `dsl/workflow.py` (and is unit-tested as a pure function in `cascade.py`):

```
final_replace = action.replace or source.replace or None
final_append  = "\n\n".join(filter(None, [
    source.append,
    action.append,
])) or None

# Runtime then composes:
if final_replace is not None:
    system_prompt = final_replace
else:
    system_prompt = base_tracecat_default
if output_type is not None:
    system_prompt += structured_output_instruction
if final_append:
    system_prompt = f"{system_prompt}\n\n{final_append}"
if instructions:  # legacy ai.action field
    system_prompt = f"{system_prompt}\n\n{instructions}"
```

`replace`: action wins over source. `append`: contributions cumulate in source-then-action order. The legacy per-action `instructions` field is appended last, untouched.

## Webhook usage

Because all string fields accept Tracecat template expressions, this also enables passing the system prompt at trigger time:

```bash
curl -X POST .../webhooks/wf_xxx/secret \
  -d '{
        "user_prompt": "Hello",
        "system_prompt": "You are a domain-specific assistant for our team."
      }'
```

…with the action wired as `system_prompt_replace: ${{ TRIGGER.system_prompt }}`. The trigger value flows through the cascade like a statically configured override.

## Iterations and verification procedure

The PR was developed incrementally and verified end-to-end against a remote Ollama backend running `qwen3.5:9b`:

1. Started with the DB migration + schemas, confirmed Pydantic round-trips cleanly.
2. Wired the cascade into `dsl/workflow.py:AI_ACTION` first, hit `Activity function ... not registered` on the first run because the activity was only registered on the agent worker. Fixed by adding it to the DSL worker too (commit 5).
3. Wired `claude_code/runtime.py:_build_system_prompt` and verified locally that `qwen3.5:9b` switches identity:
   - **Default**: replies as a "Tracecat automation assistant"
   - **`system_prompt_replace="You are Qwen, an AI model by Alibaba."`**: replies as Qwen
4. Refactored the cascade into a pure helper for unit testing. 31 tests added.
5. Frontend: extended schema + form, ran `pnpm exec biome check --write` + `pnpm exec tsc --noEmit` clean.
6. Confirmed end-to-end via the UI on a sub-path deployment: edit custom source → save → run workflow → verify model identity in the run output panel.

### Tests

| Check | Result |
| --- | --- |
| `uv run ruff check --no-cache <changed files>` | All checks passed |
| `uv run ruff format --no-cache --check <changed files>` | All formatted |
| `pnpm -C frontend exec biome check src/components/organization/org-settings-agent.tsx src/client/types.gen.ts` | No issues |
| `pnpm -C frontend exec tsc --noEmit` | No errors |
| `uv run pytest --collect-only tests/unit/test_agent_provider_cascade.py tests/unit/test_agent_runtime_system_prompt.py tests/unit/test_pydantic_ai_runtime_collapse.py` | 31 tests collected |

The repo's full pytest suite needs `just cluster up` to provide PostgreSQL/Redis — the new tests are designed to run inside that environment via CI. The cascade and collapse helpers are also exercised manually with a smoke script (results in the commit body).

End-to-end smoke: live workflow run on a Tracecat deployment with a custom Ollama source. Set `system_prompt_replace="You are Qwen, an AI model by Alibaba. Reply concisely."` on the source, ran `ai.action` with `Who are you ?`. Model replied as Qwen, no Tracecat self-identification, baseline tokens removed from the system prompt.

## What this PR does NOT do

- **Does not change any default behaviour.** All new fields default to `None`. Existing workflows and existing custom sources behave exactly as they do today.
- **Does not touch the EE schema module.** Action-level overrides on `ai.action` (i.e. `system_prompt_replace` / `system_prompt_append` as inputs of the action itself) require two fields to be added to `AgentActionArgs` in `packages/tracecat-ee/tracecat_ee/agent/schemas.py`. The cascade in this PR reads them defensively via `getattr` so no error is raised, and the source-level path is fully functional without that EE change. Happy to follow your guidance — apply the EE one-liner at merge time, or land this PR as source-level only and fold the action-level addition into a follow-up.
- **Does not regenerate `ai-action.mdx`.** The doc generator depends on the EE registry layout and re-running it locally surfaces unrelated diff churn. The cross-link is wired into `scripts/generate_ai_docs.py` so the next regen picks it up cleanly.
- **Does not modify any default Tracecat baseline string.** The string is preserved as-is and is still the default when neither the source nor the action sets `replace`.

## Documentation

A new page `docs/agents/system-prompt-overrides.mdx` walks through:

- when to use replace vs append (with two short cards),
- how to configure on a custom source via the UI,
- the resolution order across the Tracecat default, the source, and the (forthcoming) action levels,
- webhook trigger usage with `${{ TRIGGER.system_prompt }}`,
- backward compatibility guarantees.

A cross-link is wired into the auto-doc generator so the existing `ai-action.mdx` page picks it up on the next regen.

## Disclosure

Written with Claude Code, supervised by a developer with deep backend experience. Every iteration above was driven by a concrete bug or behaviour observed on a running deployment — not by guesswork. Even so, only the maintainers can judge whether each change fits the codebase's existing conventions; please flag anything off and I'll rework it.

---
## Summary by cubic
Expose system prompt overrides on custom providers so you can replace or append the default agent baseline across all workflows that use that source. Defaults keep current behavior; existing sources and workflows are unchanged. Addresses #2654.

- **New Features**
  - Backend/runtime: add `system_prompt_replace` and `system_prompt_append` (nullable) to `agent_custom_provider`; resolve cascade (action > source > default) via a new Temporal activity; `claude_code` and `pydantic-ai` runtimes honor overrides and append legacy `instructions` last.
  - Frontend: add “System prompt replace” and “System prompt append” fields to the custom source modal; make the dialog scrollable on small viewports.
  - Docs & tests: new “System prompt overrides” guide; unit tests cover cascade, `claude_code` builder, and `pydantic-ai` collapse.

- **Bug Fixes**
  - Graceful fallback on provider override lookup: catch DB errors and Temporal activity failures in the DSL; return empty overrides so runs keep the default baseline.
  - Register the provider overrides activity on the DSL worker to prevent “Activity function … is not registered” errors in `ai.action`.
  - Preserve empty-string replace in the UI payload so an explicit “no baseline” override is kept end-to-end.

<sup>Written for commit ec75042cdd1d273ae32f7f11c8530688e3d84cb6. Summary will update on new commits.</sup>

### DIFF
--- a/alembic/versions/c969b5f63428_add_system_prompt_overrides_to_agent_.py
+++ b/alembic/versions/c969b5f63428_add_system_prompt_overrides_to_agent_.py
@@ -1,0 +1,42 @@
+"""add system prompt overrides to agent_custom_provider
+
+Revision ID: c969b5f63428
+Revises: d0b32dce7f81
+Create Date: 2026-05-08 16:00:00.000000
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "c969b5f63428"
+down_revision: str | None = "d0b32dce7f81"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Add ``system_prompt_replace`` and ``system_prompt_append`` columns to
+    ``agent_custom_provider``.
+
+    Both columns are nullable Text, defaulting to NULL on existing rows so
+    custom providers created before this migration keep their current
+    behaviour (Tracecat default system prompt applied at runtime).
+    """
+    op.add_column(
+        "agent_custom_provider",
+        sa.Column("system_prompt_replace", sa.Text(), nullable=True),
+    )
+    op.add_column(
+        "agent_custom_provider",
+        sa.Column("system_prompt_append", sa.Text(), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("agent_custom_provider", "system_prompt_append")
+    op.drop_column("agent_custom_provider", "system_prompt_replace")
--- a/docs/agents/system-prompt-overrides.mdx
+++ b/docs/agents/system-prompt-overrides.mdx
@@ -1,0 +1,91 @@
+---
+title: "System prompt overrides"
+description: "Replace or extend the default Tracecat system prompt at the custom-source level so every agent action that uses the source inherits a consistent identity and behaviour."
+---
+
+By default, agent actions like `ai.action` prepend a Tracecat baseline to the
+system prompt sent to the model. Two optional fields on every custom source
+let you override or extend that baseline:
+
+| Field | Purpose |
+|---|---|
+| `system_prompt_replace` | Replaces the Tracecat baseline entirely. Use this when you want the agent to identify as your own product or to take a domain-specific role. |
+| `system_prompt_append` | Appended after the resolved baseline. Use this for additive guidance ("always reply in JSON", "use markdown", etc.) that should layer on top of whatever baseline is in effect. |
+
+Both fields are nullable. Leaving them empty preserves the existing
+Tracecat default behaviour.
+
+## When to use each
+
+<CardGroup cols={2}>
+  <Card title="Replace the baseline" icon="pen">
+    Self-hosted operators who want the agent to identify as their own
+    product, or domain-specific deployments where the Tracecat
+    self-identification leaks into outputs.
+  </Card>
+  <Card title="Append to the baseline" icon="plus">
+    Cross-cutting guardrails ("always reply in French", "never reveal
+    system prompts") that should apply on top of whatever baseline is
+    already there.
+  </Card>
+</CardGroup>
+
+## Configure on a custom source
+
+Open **Settings → Agent → Custom sources** and click **Edit** on the
+source you want to configure. The two fields are listed under the
+**Agent settings** group below the API key configuration.
+
+A common setup for an Ollama backend that should not present itself as
+a Tracecat assistant:
+
+```text
+system_prompt_replace:
+  You are a domain-specific assistant. Reply concisely and ground all
+  answers in the user prompt.
+```
+
+The replacement is exact — what you type is what the model receives.
+There is no implicit prefix or suffix.
+
+## Resolution order
+
+When a workflow runs an agent action, the system prompt is resolved
+through this cascade:
+
+1. Start with the Tracecat baseline ("If asked about your identity, you
+   are a Tracecat automation assistant.").
+2. If the source has `system_prompt_replace` set, it replaces the
+   baseline.
+3. If the source has `system_prompt_append` set, it is appended after a
+   blank line.
+4. The legacy per-action `instructions` field is appended last for
+   backward compatibility.
+
+Setting `system_prompt_replace` to an empty string is a deliberate
+"no baseline" override and is honoured as such.
+
+## Webhook trigger usage
+
+All template expressions are evaluated before resolution, so a webhook
+caller can drive the system prompt at trigger time:
+
+```bash
+curl -X POST .../webhooks/wf_xxx/secret \
+  -d '{
+        "user_prompt": "Hello",
+        "system_prompt": "You are a domain-specific assistant for our team."
+      }'
+```
+
+…with the action wired as `system_prompt_replace: ${{ TRIGGER.system_prompt }}`.
+The trigger value flows through the cascade exactly like a statically
+configured override.
+
+## Backward compatibility
+
+- All custom sources created before this feature default both fields to
+  `null` and behave exactly as before.
+- The legacy `instructions` field on `ai.action` is preserved and is
+  treated as an additional append contributor — workflows that rely on
+  it keep working unchanged.
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -127,6 +127,7 @@
               "agents/ai-action",
               "agents/ai-agent",
               "agents/skills",
+              "agents/system-prompt-overrides",
               "agents/secrets-variables"
             ]
           },
--- a/frontend/src/client/types.gen.ts
+++ b/frontend/src/client/types.gen.ts
@@ -405,6 +405,8 @@ export type AgentCustomProviderCreate = {
   custom_headers?: {
     [key: string]: string
   } | null
+  system_prompt_replace?: string | null
+  system_prompt_append?: string | null
 }
 
 /**
@@ -426,6 +428,8 @@ export type AgentCustomProviderRead = {
   passthrough: boolean
   api_key_header: string | null
   last_refreshed_at: string | null
+  system_prompt_replace?: string | null
+  system_prompt_append?: string | null
 }
 
 /**
@@ -440,6 +444,8 @@ export type AgentCustomProviderUpdate = {
   custom_headers?: {
     [key: string]: string
   } | null
+  system_prompt_replace?: string | null
+  system_prompt_append?: string | null
 }
 
 export type AgentModel = {
--- a/frontend/src/components/organization/org-settings-agent.tsx
+++ b/frontend/src/components/organization/org-settings-agent.tsx
@@ -114,6 +114,8 @@ const customProviderSchema = z
     apiKey: z.string().optional(),
     customHeadersJson: z.string().optional(),
     passthrough: z.boolean(),
+    systemPromptReplace: z.string().optional(),
+    systemPromptAppend: z.string().optional(),
   })
   .superRefine((value, ctx) => {
     const raw = value.customHeadersJson?.trim()
@@ -159,6 +161,8 @@ const DEFAULT_CUSTOM_PROVIDER_VALUES: CustomProviderFormValues = {
   apiKey: "",
   customHeadersJson: "",
   passthrough: false,
+  systemPromptReplace: "",
+  systemPromptAppend: "",
 }
 
 const CLOUD_CATALOG_PROVIDERS = [
@@ -711,6 +715,8 @@ function getProviderDialogDefaults(
     apiKey: "",
     customHeadersJson: "",
     passthrough: provider.passthrough,
+    systemPromptReplace: provider.system_prompt_replace ?? "",
+    systemPromptAppend: provider.system_prompt_append ?? "",
   }
 }
 
@@ -724,6 +730,8 @@ function buildProviderCreatePayload(
     api_key: normalizeOptional(values.apiKey),
     custom_headers: parseCustomHeaders(values.customHeadersJson),
     passthrough: values.passthrough,
+    system_prompt_replace: normalizeOptional(values.systemPromptReplace),
+    system_prompt_append: normalizeOptional(values.systemPromptAppend),
   }
 }
 
@@ -735,6 +743,8 @@ function buildProviderUpdatePayload(
     base_url: normalizeOptional(values.baseUrl),
     api_key_header: normalizeOptional(values.apiKeyHeader),
     passthrough: values.passthrough,
+    system_prompt_replace: normalizeOptional(values.systemPromptReplace),
+    system_prompt_append: normalizeOptional(values.systemPromptAppend),
   }
 
   const apiKey = normalizeOptional(values.apiKey)
@@ -889,7 +899,7 @@ function CustomProviderDialog({
 
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
-      <DialogContent className="sm:max-w-[560px]">
+      <DialogContent className="flex max-h-[90vh] flex-col sm:max-w-[560px]">
         <DialogHeader>
           <DialogTitle>
             {provider ? "Edit custom source" : "Add custom source"}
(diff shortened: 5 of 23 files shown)


eviltik:
Thanks @cubic-dev-ai — all three issues are valid and pushed in `ec75042cd`:

1. **`provider/activities.py:75`** (P1) — wrapped the DB query in `try/except` and added a warning log. The docstring promised graceful fallback; the implementation now matches.

2. **`dsl/workflow.py:1042`** (P2) — wrapped the `execute_activity` call in `try/except (ActivityError, ApplicationError)` so Temporal-level failures (timeout, worker death) also fall back to the default cascade instead of aborting the run.

3. **`org-settings-agent.tsx:733`** (P2) — added a dedicated `preserveEmptyOptional` helper that only collapses `null`/`undefined`. Used for the two new override fields so a deliberate empty-string `system_prompt_replace` reaches the backend untouched. The runtime already honours an explicit empty string as a "no baseline" override; the frontend was the missing piece.

@cubic-dev-ai please re-run.

topher-lo:
@eviltik appreciate you putting the time into this but it's too risky for us to merge this (we take our code quality and surface area very seriously, especially if it's related to user experience)

My suggestion: if you really need this, can you reimplement this has an environment variable override. This should be a 5-10 line update with less than 4 files changed (`config.py`, the ai action and ai agents file, etc)

eviltik:
@topher-lo HA ! thank you for your feedback, let me take a look at this

eviltik:
Thanks @topher-lo for the quick read and the steer.

Totally fair on the surface-area concern. The env-var path you describe is straightforward and we'll likely use it as a fallback shim. That said, I want to be transparent that it doesn't fully cover what we hit in practice, so future readers landing here understand the tradeoff:

- **Per-source identity** is the actual need. We run multiple custom sources side by side (different Ollama models, an OpenAI fallback, etc.). A single global env var means every source picks up the same identity baseline; the moment two backends should present themselves differently, the env var has to flip per-deployment, which negates the "self-service custom source" UX that already exists for passthrough mode and credentials.
- **Webhook-driven prompts** were a knock-on benefit: with the cascade in place, a trigger can pass `system_prompt` in the body and the action wires it through expression templating. An env var can't do that since it's resolved once at process start.
- **Surface-area framing**: this PR doesn't actually add any new user-facing concept — it reuses the existing custom-source modal (next to Passthrough mode) and the existing template expression system. The DB column is the only really new surface, and it's nullable + opt-in.

I respect that you're the one weighing the maintenance cost against the breadth of users who'd benefit, so I'll close this PR and ship the env-var version as you suggested. The richer cascade lives on my fork ([eviltik/tracecat](https://github.com/eviltik/tracecat) on `feat/ai-action-system-prompt-overrides`) for anyone bumping into the same multi-source need — happy to revisit upstream if/when other contributors raise the same use case.

Issue #2654 stays open as a tracking signal for that.

PR #2658, opened 2026-05-08
feat(agent): expose per-source allowed_tools control on custom providers

### LINKED ISSUE #2655: feat: allow disabling/filtering Claude SDK built-in tools on `ai.action`
### Update — scope clarification (post-investigation)

Same as for TracecatHQ/tracecat#2654 (cross-link): exposing `allowed_tools` on the `ai.action` form requires extending `AgentActionArgs` in `packages/tracecat-ee/tracecat_ee/agent/schemas.py`, which is in the EE module. The custom-source-level part is fully OSS.

We don't intend to ship code into the EE module ourselves. The OSS PR will include a defensive `getattr` so the cascade is forward-compatible — when EE exposes the field, the action-level override starts taking effect automatically with no further OSS change. Maintainers can apply the EE one-liner at merge time, or guide us towards a different layout before the PR opens.

---

### TL;DR

`ai.action` always inlines the JSON schemas of the full Claude Code built-in toolset (Bash, Read, Write, Edit, Grep, Glob, MultiEdit, etc.) in the system prompt — even though the action is documented as a "simple LLM call, no tools" and runs with `max_tool_calls=0`.

That's **\~3 940 input tokens of pure overhead per call** (96 % of the system context), measured below.

This issue proposes exposing one optional field (`allowed_tools: list[str] | None`) at both the custom-source and action levels, mapping directly onto `ClaudeAgentOptions.tools` which **already supports** the full range (`None` = SDK default, `[]` = disable all built-ins, `["Bash","Read"]` = whitelist).

Default `None` everywhere → zero behaviour change for existing workflows.

### Measured impact

Tested directly against the bundled CLI 2.1.85 inside `sysdio2_tracecat_agent_executor`, `ANTHROPIC_BASE_URL` pointed at a remote Ollama running `qwen3.5:9b`. User prompt: `"Reply: hi"` (3 tokens). The CLI's `--output-format json` returns a `usage.input_tokens` field that captures system prompt + tool definitions + user prompt cumulatively.

| \# | Configuration | `input_tokens` | Delta vs A |
| -- | -- | -- | -- |
| A | CLI defaults (no flags) | 4 096 | reference |
| B | `--system-prompt "You are Qwen, by Alibaba."` | 4 096 | 0 |
| C | `--system-prompt "..." --tools ""` | 158 | **−96 %** |
| D | `--bare --system-prompt "..." --tools ""` | 157 | **−96 %** |

So roughly **96 % of the per-call overhead** comes from the inlined JSON tool definitions. For an `ai.action` that's just doing a one-shot LLM completion (classification, extraction, summarization), this is pure waste — and on top of that, smaller models often hallucinate around tools they "see" but cannot actually call.

### Capability already in the SDK

`ClaudeAgentOptions.tools` natively supports:

* `tools=None` → no `--tools` flag → CLI applies its default (full toolset).
* `tools=[]` → `--tools ""` → all built-ins disabled.
* `tools=["Bash","Read"]` → `--tools "Bash,Read"` → whitelist.

The proposal is to surface this in the Tracecat schema.

#### Custom source (default)

| Field | Type | Default |
| -- | -- | -- |
| `allowed_tools` | `list[str] \| None` | `None` |

#### `ai.action` schema (per-call override)

Same field. When set, takes precedence over the source-level value.

#### Cascade rule

```python
if action.allowed_tools is not None:
    resolved = action.allowed_tools          # action overrides source (even if [])
elif source.allowed_tools is not None:
    resolved = source.allowed_tools          # source default if no action value
else:
    resolved = None                           # SDK default (full toolset)
```

The runtime maps `resolved` directly onto `ClaudeAgentOptions.tools`.

### Backward compatibility

* New field defaults to `None` → **zero behaviour change** for existing workflows or sources.
* `disallowed_tools` and existing built-in filtering logic in Tracecat are untouched.
* MCP servers (`tracecat-registry`) are not affected by this PR.

### Note on `pydantic_ai` runtime

The pydantic-ai harness doesn't have the same notion of "implicit built-in tools" — it only sees the tools you give it. So `allowed_tools` would be a no-op there. We'd log a `WARNING` if it's set on a workflow that resolves to the pydantic-ai runtime, and document this clearly.

### Scope of implementation

* `tracecat/db/models.py`: `allowed_tools JSONB NULL` on `agent_custom_provider`.
* Alembic migration.
* `tracecat/agent/provider/schemas.py`: extend Create/Update/Read.
* Registry action schema for `ai.action`: add `allowed_tools`.
* `tracecat/dsl/action.py:build_agent_args_activity`: cascade.
* `tracecat/agent/runtime/claude_code/runtime.py`: pass `tools=resolved` to `ClaudeAgentOptions`.
* `tracecat/agent/runtime/pydantic_ai/runtime.py`: warn if set, ignore.
* Frontend: types regenerated via `just gen-client-ci`. Custom source modal: add the field under "Agent settings". Action form picks it up automatically through schema generation.
* Tests:
  * Unit: verify the `cmd` array produced by `subprocess_cli._build_command` includes `--tools ""` when `tools=[]`, `--tools "Bash,Read"` when whitelist, and no `--tools` flag when `None`.
  * Cascade: matrix of source × action × None.
  * Token-overhead regression: integration test asserting `input_tokens` for `tools=[]` is well below the default.
* Docs: update `docs/agents/ai-action.mdx` with a "Reducing token overhead" section.

### Open question for maintainers

Two design alternatives we considered but rejected:

1. **Auto-disable tools when** `model_provider == "custom-model-provider"`. Pro: zero UX. Con: silent behaviour change for existing custom-source users — some might rely on tools being there.
2. **Auto-disable tools when** `max_tool_calls == 0`. Same trade-off, with the bonus of fitting the `ai.action` semantics. Still a silent behaviour change though.

The opt-in design proposed here is more conservative and matches the spirit of `Passthrough mode` (you choose when to skip default behaviour). Happy to discuss if maintainers prefer one of the auto-disable variants.

Happy to send the PR. Splitting source-level and action-level fields into two PRs is doable.

eviltik:
Closes #2655

# Per-source `allowed_tools` control on custom providers

## Why

`ai.action` runs through `claude_code/runtime.py` with no `tools=`
argument set on `ClaudeAgentOptions`, so the bundled CLI activates
its **full built-in toolset** and inlines every tool's JSON schema
in the system prompt sent to the model — even when
`max_tool_calls=0`, the documented mode for `ai.action`.

Measured against `qwen3.5:9b` via Ollama with a one-line user prompt
(`"Reply: hi"`):

| Configuration | `input_tokens` | Delta |
|---|---:|---:|
| Defaults (current behaviour) | 4 096 | reference |
| `tools=""` on the CLI (this PR's wire effect) | 158 | **−96 %** |

For workflows that fire `ai.action` thousands of times a day, ~3.9 M
tokens of pure overhead per 1k calls — meaningful on hosted models,
meaningful in latency and context budget on local models.

The capability already exists in `claude-agent-sdk==0.1.51`:
`ClaudeAgentOptions.tools` accepts `None` (SDK default), `[]`
(disable everything), or a whitelist; all three map directly to
`--tools "..."` on the bundled CLI 2.1.85. This PR surfaces those
three states at the custom-source level (and forward-compatibly at
the action level pending the EE schema discussion in #2654).

## Design choice

This PR uses the same shape as #2657 (per-source DB column +
cascade + modal UI), deliberately. We considered the env-var
fallback you suggested on #2657, and it doesn't fit:

- a global env var resolved at process start can't differentiate
  between two custom sources running side by side;
- it can't be overridden at trigger time via `${{ TRIGGER.* }}`;
- it doesn't survive the "self-service custom source" UX promise
  that already exists for passthrough mode and credentials.

Reusing the same design as #2657 isn't a way to relitigate that PR
— it's a consistency point. If the per-source / cascade pattern is
the right shape for one piece of agent configuration, it's the
right shape for another. Happy to be arbitrated on the functional
case rather than ship two different shoehorned alternatives.

The maintenance cost is bounded: the cascade lives in a pure
function (`tracecat/agent/provider/tool_cascade.py`), the activity
is dependency-free, and the runtime change is a one-line forward
to `ClaudeAgentOptions.tools`. None of the existing routing or
auth surfaces are touched.

## What this PR does NOT do

- No behaviour change at the default. `allowed_tools = NULL`
  preserves today's full SDK toolset for every existing source.
- No EE module touched. Action-level overrides are read defensively
  via `getattr` so the cascade is forward-compatible the day
  `AgentActionArgs` exposes the field.

## Tests

Lint / format / typecheck pass on the touched files. Cascade
resolver covered by unit tests; runtime mapping (`tools=None` /
`[]` / whitelist → `--tools` flag) verified against the live SDK
end-to-end on a Tracecat dev deployment, with the token math above
reproducible from inside the agent_executor container.

---
## Summary by cubic
Adds per-source `allowed_tools` on custom providers to control which Claude SDK built-in tools are exposed to `ai.action`, reducing prompt tokens and latency. Defaults unchanged; sources with null keep the full toolset.

- **New Features**
  - DB/API: `agent_custom_provider.allowed_tools` (nullable JSONB) with create/read/update support.
  - UI: custom source modal adds “Built-in tools (Claude SDK)” with Default / Disable all / Whitelist (CSV); form-to-API mapping preserves the tristate; modal made scrollable.
  - Orchestration: new activity resolves source-level tools by `catalog_id`; pure helper applies cascade (action > source > default).
  - Runtime: resolved list passed to `ClaudeAgentOptions.tools`; `pydantic_ai` accepts, warns, and ignores.
  - Tests: unit tests for cascade precedence and empty-list semantics.

<sup>Written for commit 43902d69b4c755f1333e1ce27d6a735795004313. Summary will update on new commits.</sup>

### DIFF
--- a/alembic/versions/73d6e838c5aa_add_allowed_tools_to_agent_custom_provider.py
+++ b/alembic/versions/73d6e838c5aa_add_allowed_tools_to_agent_custom_provider.py
@@ -1,0 +1,41 @@
+"""add allowed_tools to agent_custom_provider
+
+Revision ID: 73d6e838c5aa
+Revises: d0b32dce7f81
+Create Date: 2026-05-08 17:00:00.000000
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "73d6e838c5aa"
+down_revision: str | None = "d0b32dce7f81"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Add ``allowed_tools`` to ``agent_custom_provider``.
+
+    Nullable JSONB list. ``NULL`` means "no override at the source
+    level"; the runtime keeps its default (full SDK toolset). An empty
+    list ``[]`` is a deliberate "disable all built-in tools" override.
+    """
+    op.add_column(
+        "agent_custom_provider",
+        sa.Column(
+            "allowed_tools",
+            postgresql.JSONB(astext_type=sa.Text()),
+            nullable=True,
+        ),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("agent_custom_provider", "allowed_tools")
--- a/frontend/src/client/types.gen.ts
+++ b/frontend/src/client/types.gen.ts
@@ -405,6 +405,7 @@ export type AgentCustomProviderCreate = {
   custom_headers?: {
     [key: string]: string
   } | null
+  allowed_tools?: Array<string> | null
 }
 
 /**
@@ -426,6 +427,7 @@ export type AgentCustomProviderRead = {
   passthrough: boolean
   api_key_header: string | null
   last_refreshed_at: string | null
+  allowed_tools?: Array<string> | null
 }
 
 /**
@@ -440,6 +442,7 @@ export type AgentCustomProviderUpdate = {
   custom_headers?: {
     [key: string]: string
   } | null
+  allowed_tools?: Array<string> | null
 }
 
 export type AgentModel = {
--- a/frontend/src/components/organization/org-settings-agent.tsx
+++ b/frontend/src/components/organization/org-settings-agent.tsx
@@ -114,6 +114,8 @@ const customProviderSchema = z
     apiKey: z.string().optional(),
     customHeadersJson: z.string().optional(),
     passthrough: z.boolean(),
+    allowedToolsMode: z.enum(["default", "disable_all", "whitelist"]),
+    allowedToolsList: z.string().optional(),
   })
   .superRefine((value, ctx) => {
     const raw = value.customHeadersJson?.trim()
@@ -159,6 +161,8 @@ const DEFAULT_CUSTOM_PROVIDER_VALUES: CustomProviderFormValues = {
   apiKey: "",
   customHeadersJson: "",
   passthrough: false,
+  allowedToolsMode: "default",
+  allowedToolsList: "",
 }
 
 const CLOUD_CATALOG_PROVIDERS = [
@@ -698,12 +702,45 @@ function ProviderMetaPill({
   )
 }
 
+function allowedToolsToFormState(
+  allowedTools: readonly string[] | null | undefined
+): {
+  mode: "default" | "disable_all" | "whitelist"
+  list: string
+} {
+  if (allowedTools == null) {
+    return { mode: "default", list: "" }
+  }
+  if (allowedTools.length === 0) {
+    return { mode: "disable_all", list: "" }
+  }
+  return { mode: "whitelist", list: allowedTools.join(", ") }
+}
+
+function allowedToolsFromFormState(
+  mode: "default" | "disable_all" | "whitelist",
+  list: string | undefined
+): string[] | null {
+  if (mode === "default") {
+    return null
+  }
+  if (mode === "disable_all") {
+    return []
+  }
+  // whitelist: split CSV, trim, drop empties
+  return (list ?? "")
+    .split(",")
+    .map((s) => s.trim())
+    .filter((s) => s.length > 0)
+}
+
 function getProviderDialogDefaults(
   provider: AgentCustomProviderRead | null
 ): CustomProviderFormValues {
   if (!provider) {
     return DEFAULT_CUSTOM_PROVIDER_VALUES
   }
+  const tools = allowedToolsToFormState(provider.allowed_tools)
   return {
     displayName: provider.display_name,
     baseUrl: provider.base_url ?? "",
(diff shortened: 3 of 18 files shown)


eviltik:
image

image

image

topher-lo:
@eviltik this actually looks pretty okay. Question: can you share a screenshot of the dialog in relation to the rest of the page? Is this in org settings. Not very clear from the screen

topher-lo:
~Also what is "Test connection" for?~ Okay I see what this dialog is now

eviltik:
Yes — it's in the existing custom-source modal under Settings → Agent → Custom sources → Edit (same dialog as Passthrough mode, Base URL, Custom headers, API key). The new "Built-in tools (Claude SDK)" block sits right under Passthrough mode and reuses the same card styling. Screenshot with surrounding context coming in a moment.

"Test connection" is the existing button — unchanged by this PR. It calls the source's `/models` endpoint to verify credentials and base URL before saving. We didn't touch its behaviour.

image

topher-lo:
@eviltik more context: the code quality and blast radius is okay it's just unclear if this is the correct solution

topher-lo:
Going to close this. It's a legit issue but we'd like to take it because it so core to the user experience.

We're going to add tool selection at the action level i.e. `ai.agent` directly AND we're going to fix the bug where `ai.action` see Claude default tools.

These changes should address your core issues.

PR #2637, opened 2026-05-06
feat(agents): chat stream resumption

jordan-umusu:
## Summary by cubic
Enable reliable chat stream resumption across reconnects and reloads. The server now honors `Last-Event-ID` and handles turn startup failures to avoid stuck input, stale cursors, and missing history.

- **New Features**
  - Enable resume mode in the chat hook and add `prepareReconnectToStreamRequest` for `/api/agent/sessions/{id}/stream?workspace_id=...` with credentials.
  - Persist the pending user message in `sessionStorage` (`chat-pending-msg:<id>`) until echoed, then clear.
  - On new turns, immediately set the saved stream cursor to "0-0" and flush SDK session history in the background.

- **Bug Fixes**
  - If a new turn fails to start, clear the stream buffer and reset the saved cursor via `abort_new_turn`; skip for `ContinueRunRequest`.
  - Stream start logic: return 204 only when there’s no saved `last_stream_id` and no `Last-Event-ID`. Otherwise attach starting from `Last-Event-ID`, then `last_stream_id`, else "0-0".
  - Improve reliability: persist session lines before `done`, handle partial JSONL lines, serialize concurrent socket sends, and keep the approval continuation prompt internal even if results arrive early.

<sup>Written for commit 581eab49b29ddda8c77bb85bf48ae1d5aa0c8d7a. Summary will update on new commits.</sup>

PR #2659, opened 2026-05-09
perf(agent): flush session history without stream latency

daryllimyt:
## Summary
- Persist Claude Code JSONL session lines during turns by tailing appended bytes instead of rereading the full file on every stream event.
- Coalesce eager persistence with an `asyncio.Event`, keeping stream events on the fast path and preserving a final flush before `done`.
- Serialize socket writes so foreground stream frames and background session-line frames cannot interleave.
- Tighten session-line persistence typing and add regression coverage for byte tailing, partial lines, resume/fork offsets, final-only turns, stream latency, and socket serialization.

## Review Notes
The runtime now treats SDK session persistence as a side channel: foreground stream events are sent first, then the flusher tails new JSONL rows by byte offset under its own lock. The final flush still runs before `done`, so result-only turns and races with the background task persist their session rows before completion.

```mermaid
sequenceDiagram
    participant SDK as Claude SDK
    participant Runtime as Runtime loop
    participant Flusher as Session flusher
    participant Socket as Event writer
    participant Parent as Orchestrator

    SDK-->>Runtime: stream / system / result message
    Runtime->>Socket: send foreground runtime event
    Runtime->>Flusher: signal flush event
    Flusher->>Flusher: tail JSONL from last byte offset
    Flusher->>Flusher: classify line visibility
    Flusher->>Socket: send_session_line(..., internal=...)
    Socket->>Parent: serialized socket write
    Runtime->>Flusher: final flush before done
    Runtime->>Socket: send done
```

Approval continuation prompts are classified at emit time instead of using a mutable "hide next prompt" flag. Both the background flush and final flush receive the same immutable approval-continuation context for the current SDK query, so whichever wins the race applies the same classification.

```mermaid
sequenceDiagram
    participant Runtime as Runtime loop
    participant SDK as Claude SDK
    participant File as SDK JSONL file
    participant Flusher as Session flusher
    participant Socket as Event writer

    Runtime->>SDK: query(..., is_approval_continuation=True)
    SDK->>File: append continuation prompt row
    par Background flush wins
        Runtime->>Flusher: signal flush with approval context
    and Result arrives first
        SDK-->>Runtime: ResultMessage
        Runtime->>Flusher: final flush with approval context
    end
    Flusher->>File: read new rows by byte offset
    Flusher->>Flusher: internal = existing_internal || approval_prompt_match
    Flusher->>Socket: persist continuation prompt as internal
```

## Session Interrupt Safety
This change does not make the streamed UI state the source of truth. The Claude SDK JSONL file remains canonical; the runtime only copies complete, parseable rows out of that file sooner.

For a graceful interrupt, such as the existing approval path, the runtime emits the approval request, calls `client.interrupt()`, and lets the SDK append its normal interrupt/control rows. Those rows then flow through the same downstream session-line path: the flusher tails them, the socket writer sends serialized `session_line` envelopes, and loopback persists them in `AgentSessionHistory` with UUID dedupe.

That means future user interrupts should use the same shape: send a graceful interrupt to the runtime, let the SDK write the interruption rows, then let the existing flush/final-drain path persist them. The safeguards are byte-offset tailing, complete-line parsing, offset advancement only after successful send, a flush lock, socket write serialization, loopback UUID dedupe, and a final flush before `done`.

The caveat is hard cancellation. Killing the runtime task/process or closing the socket would not corrupt persisted history, because partial lines are ignored, but it can still drop the final rows the SDK wrote but we never drained. User interrupt should therefore be implemented as a graceful `client.interrupt()`-style control action, not as process cancellation.

## Testing
- `uv run ruff check tests/unit/test_agent_runtime.py tests/unit/test_agent_session_messages.py tests/unit/test_agent_socket_io.py tracecat/agent/common/socket_io.py tracecat/agent/executor/loopback.py tracecat/agent/runtime/claude_code/runtime.py`
- `uv run ruff format --check tests/unit/test_agent_runtime.py tests/unit/test_agent_session_messages.py tests/unit/test_agent_socket_io.py tracecat/agent/common/socket_io.py tracecat/agent/executor/loopback.py tracecat/agent/runtime/claude_code/runtime.py`
- `uv run basedpyright tests/unit/test_agent_runtime.py tests/unit/test_agent_session_messages.py tests/unit/test_agent_socket_io.py tracecat/agent/common/socket_io.py tracecat/agent/executor/loopback.py tracecat/agent/runtime/claude_code/runtime.py`
- `PG_PORT=5532 MINIO_PORT=9100 REDIS_PORT=6479 TEMPORAL_PORT=7333 uv run pytest tests/unit/test_agent_runtime.py tests/unit/test_agent_socket_io.py tests/unit/test_agent_session_messages.py tests/unit/test_agent_session_list_sessions.py`

---
## Summary by cubic
Flush Claude Code session history in the background without delaying stream frames by tailing SDK JSONL files by byte offset. Also serialize socket sends and keep approval-continuation prompts internal even if the result arrives first.

- **Refactors**
  - Tail session files by byte offset and retry partial lines; seed offset from canonicalized on-disk data on resume; on fork, reset to zero unless a child file exists (then use its size).
  - Background flusher runs as a task gated by an `asyncio.Event` and a lock; stream/system/tool-result events signal flush; final flush on CLI exit; capture SDK session ID from any message; send stream frames before scheduling flush.
  - Add typed parsers for runtime envelopes and Claude session lines with explicit compaction phase parsing.

- **Bug Fixes**
  - Serialize socket writes with a lock to prevent overlapping drains and interleaved frames.
  - Keep stream events on the fast path even when session-line flush is slow; persist result-only turns before `done` with correct ordering.
  - Always mark approval-continuation prompts as internal across races; reset forked session offsets correctly.

<sup>Written for commit 9412f04213f1f4ac9e9580bd40440b641c254c75. Summary will update on new commits.</sup>

### DIFF
--- a/tests/unit/test_agent_runtime.py
+++ b/tests/unit/test_agent_runtime.py
@@ -5,6 +5,7 @@ Tests the runtime execution with mocked Claude SDK.
 
 from __future__ import annotations
 
+import asyncio
 import os
 import tempfile
 import uuid
@@ -123,6 +124,7 @@ def mock_socket_writer() -> MagicMock:
     writer.send_stream_event = AsyncMock()
     writer.send_message = AsyncMock()
     writer.send_session_line = AsyncMock()
+    writer.send_result = AsyncMock()
     writer.send_error = AsyncMock()
     writer.send_done = AsyncMock()
     writer.close = AsyncMock()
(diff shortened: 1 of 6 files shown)


daryllimyt:
@codex review

daryllimyt:
@codex review

daryllimyt:
@cubic-dev-ai review

jordan-umusu:
* **#2637** <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/2637?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2659** <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/2659?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/2659?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jordan-umusu (approved):

jordan-umusu:
LGTM

PR #2663, opened 2026-05-11
fix(agents): converse knob for bedrock

jordan-umusu:
## Problem

When litellm cannot pattern recognize against a bedrock model that needs converse support, our current hardcoded invoke path (/bedrock) will fail for said models.

---
## Summary by cubic
Add a “Use Converse” toggle for Bedrock models to route requests via the Bedrock Converse API. Also allow Bedrock updates that only change this flag, while keeping model reference rules strict.

- **New Features**
  - UI: Add “Use Converse API” switch for Bedrock in Org Settings; included in create/update payloads; defaults to false and pre-fills from metadata.
  - Catalog schemas: Add `use_converse` to Bedrock create/update models.
  - Credentials: Map `use_converse` to `AWS_BEDROCK_USE_CONVERSE` as "true"/"false".
  - Gateway: Use `bedrock/converse/...` prefix when `AWS_BEDROCK_USE_CONVERSE` is "true"; otherwise `bedrock/...`.
  - Tests: Add unit tests for prefix selection, credential serialization, and Bedrock update validator.

- **Bug Fixes**
  - Validation: Bedrock updates can change only `use_converse` (no ref fields). If refs are sent, require at most one; reject both explicitly null.

<sup>Written for commit bd7a195d8896c693e71258e5f26a8fdf71acd3a3. Summary will update on new commits.</sup>

image

### DIFF
--- a/frontend/src/client/schemas.gen.ts
+++ b/frontend/src/client/schemas.gen.ts
@@ -4874,6 +4874,11 @@ export const $BedrockCatalogCreate = {
       ],
       title: "Model Id",
     },
+    use_converse: {
+      type: "boolean",
+      title: "Use Converse",
+      default: false,
+    },
   },
   additionalProperties: false,
   type: "object",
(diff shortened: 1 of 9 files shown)


daryllimyt (approved):
lgtm

PR #2694, opened 2026-05-14
fix(saml): make accept_invitation_for_user idempotent with upserts

jordan-umusu:
## Summary by cubic
Make org invitation acceptance idempotent and run it during SAML login so invited roles apply even if SSO auto-provisioning already created membership.

- **Bug Fixes**
  - Use PostgreSQL upsert for `OrganizationMembership` and fetch the existing row on conflict to avoid duplicates.
  - Upsert `UserRoleAssignment` for the org-wide role (`workspace_id` is null); on conflict, update `role_id` within the same organization to upgrade existing assignments.
  - In SAML ACS, call `accept_invitation_for_user` to auto-accept pending invites on login; commit on success, rollback and log on error.

<sup>Written for commit 787b65926d23ed5dfad553266200b341376e3248. Summary will update on new commits.</sup>

### DIFF
--- a/tracecat/auth/saml.py
+++ b/tracecat/auth/saml.py
@@ -96,6 +96,7 @@ from tracecat.identifiers import OrganizationID
 from tracecat.invitations.enums import InvitationStatus
 from tracecat.logger import logger
 from tracecat.organization.domains import normalize_domain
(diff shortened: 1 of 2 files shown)


daryllimyt (approved):
lgtm

PR #2707, opened 2026-05-16
feat(tables): support dropping unique column index

topher-lo:
## Summary

`TablesService.create_unique_index()` had no inverse — once a column was chosen as a unique index, it was permanent for the lifetime of the table. This adds drop-index semantics across all four surfaces (service, REST, UI, MCP) with the minimum diff.

`TableColumnUpdate.is_index` becomes tri-state:

| `is_index` | Action |
|------------|--------|
| `None` (unset) | No-op |
| `true` | Create unique index (existing) |
| `false` | Drop unique index (new) |

## What's changed

- **Service** (`tracecat/tables/service.py`) — new `drop_unique_index()` that uses the SQLAlchemy inspector to look up the real PG index name, then issues a schema-qualified `DROP INDEX`. Errors with `ValueError` if no index exists on the column (mapped to 400 by the existing router). `update_column()` now dispatches on tri-state `is_index`.
- **Schema** (`tracecat/tables/schemas.py`) — documents the new tri-state in `TableColumnUpdate.is_index` description.
- **REST router** — no code change. The existing PATCH `/tables/{id}/columns/{id}` already maps `ValueError → 400`.
- **MCP** (`tracecat/mcp/server.py`) — two new dedicated tools: `tables_create_column_index` and `tables_drop_column_index`. Closes the column-level gap in the MCP surface (previously column operations were REST-only).
- **UI** (`frontend/src/components/tables/table-view-column-menu.tsx`) — dropdown menu now toggles between "Create unique index" / "Remove unique index" based on `column.is_index`. New `TableColumnDropIndexDialog` warns about upsert impact. `useUpdateColumn` error handling now uses `is_index !== undefined` to categorize both create and drop errors.
- **Tests** — three new unit tests covering drop lifecycle, no-index error, and update_column dispatch.

## Backward-compat note

Today `is_index=false` is a silent no-op (pop default was `False` and the check was truthy). After this change it attempts a drop and returns 400 if no index exists. The frontend only ever sent `true`, so the only observable change is for direct API clients that explicitly sent `false`.

## Test plan

- [x] `uv run pytest tests/unit/test_tables_service.py` — 117 passed (incl. 3 new index tests)
- [x] `uv run ruff check . && uv run ruff format --check .`
- [x] `uv run basedpyright --warnings --threads 4 tracecat/tables/ tracecat/mcp/server.py`
- [x] `pnpm -C frontend check && pnpm -C frontend run typecheck`
- [ ] Manual REST: PATCH with `{"is_index": true}` → 204; `{"is_index": false}` → 204; second drop → 400 with "No unique index exists..."
- [ ] MCP smoke: `tables_create_column_index` → `tables_drop_column_index` → confirm `tables_get_table` reflects toggling
- [ ] UI smoke: create index via menu → green "Index" badge appears; "Remove unique index" menu item → badge disappears

### DIFF
--- a/docs/snippets/mcp-tools.mdx
+++ b/docs/snippets/mcp-tools.mdx
@@ -275,6 +275,12 @@
     <ResponseField name="tables_export_csv" type="tool">
   Export table data as a staged download URL.
     </ResponseField>
+    <ResponseField name="tables_create_column_index" type="tool">
+  Create a unique index on a table column. Only one unique index per table is allowed.
+    </ResponseField>
+    <ResponseField name="tables_drop_column_index" type="tool">
+  Drop the unique index on a table column.
+    </ResponseField>
   </Expandable>
 </ResponseField>
 
--- a/frontend/src/client/schemas.gen.ts
+++ b/frontend/src/client/schemas.gen.ts
@@ -7780,7 +7780,8 @@ export const $CaseFieldUpdate = {
         },
       ],
       title: "Is Index",
-      description: "Whether the column is an index",
+      description:
+        "True creates a unique index, False drops it, None leaves unchanged.",
     },
     options: {
       anyOf: [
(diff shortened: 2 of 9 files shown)


topher-lo:
@codex review

topher-lo:
@cubic review

PR #2710, opened 2026-05-17
fix(graph): keep success+error edges between same nodes

### LINKED ISSUE #2619: Cannot connect both success and error edges from the same source action in the canvas
## Description

When connecting edges in the workflow canvas, it is not possible to have both a success edge and an error edge from the same source action to the same target action. Drawing the second edge silently removes the first.

## Steps to Reproduce

1. Open a workflow in the canvas editor.
2. Draw a success edge from action A to action B.
3. Draw an error edge from action A to action B.
4. Observe: only the most recently drawn edge remains. The first is removed.

The same happens in reverse order (error first, then success).

## Expected Behavior

Both edges should coexist, reflecting the following `depends_on` configuration:

```yaml
depends_on:
  - action_a
  - action_a.error
```

## Actual Behavior

Only one edge exists after the second is drawn. The canvas and the underlying data both reflect only the latest connection.

## Root Cause

In `tracecat/workflow/graph/service.py`, the `_add_edge` method deduplicates edges by `source_id + source_type` before appending the new edge:

```python
filtered_edges = [
    e
    for e in edges
    if not (
        e.get("source_id") == payload.source_id
        and e.get("source_type") == payload.source_type
    )
]
filtered_edges.append(new_edge)
```

Because `source_handle` (`"success"` or `"error"`) is not included in the filter, adding a success edge removes any existing error edge from the same source, and vice versa.

The same issue exists in `_delete_edge`, which matches on `source_id + source_type` only — meaning deleting one edge (e.g. error) would remove both edges from that source.

## Suggested Fix

Include `source_handle` in the deduplication filter in both methods.

**`_add_edge`:**
```python
filtered_edges = [
    e
    for e in edges
    if not (
        e.get("source_id") == payload.source_id
        and e.get("source_type") == payload.source_type
        and e.get("source_handle") == (payload.source_handle or "success")
    )
]
```

**`_delete_edge`:**
```python
target_action.upstream_edges = [
    e
    for e in edges
    if not (
        e.get("source_id") == payload.source_id
        and e.get("source_type") == payload.source_type
        and e.get("source_handle") == (payload.source_handle or "success")
    )
]
```

## Additional Notes

- The underlying data model supports multiple edges per source/target pair with different handles — this is purely a logic issue in the graph operations layer.
- The `POST /api/actions/{id}` endpoint is unaffected and can be used as a workaround to set both edges directly until a fix is released.
- The `join_strategy: any` use case with both success and error dependencies is a documented workflow pattern, making this a meaningful gap.

wtfashwin:
## Description

Fixes the canvas bug where drawing both a `success` and an `error` edge between the same source and target action silently dropped one of them. The edge dedup filter in `_add_edge` and `_delete_edge` matched on `source_id + source_type` only, ignoring `source_handle` — so the second edge always evicted the first.

The fix introduces `WorkflowGraphService._edge_dedup_key`, a small static helper that returns a canonical key including `source_handle`. For `udf` sources it defaults missing handles to `"success"` so legacy data dedupes against newly-written edges; trigger edges keep their handle-less semantics.

## Related Issues

Fixes #2619

## Steps to QA

Automated coverage in `tests/unit/test_workflow_graph_edges.py`:

- success → error and error → success leave both edges intact
- re-adding the same handle is idempotent (no duplicates)
- deleting one handle leaves the other intact (both directions)
- trigger edges still dedupe by `(source_id, source_type)` only
- a legacy udf edge without `source_handle` is treated as `"success"` and dedupes correctly

Manual:

1. Open a workflow in the canvas editor.
2. Draw a success edge A → B, then an error edge A → B. Both should remain.
3. Draw them in the opposite order. Both should remain.
4. Delete the error edge. The success edge should remain.

```
uv run pytest tests/unit/test_workflow_graph_edges.py tests/unit/test_workflow_graph_reconciliation.py tests/unit/api/test_api_workflow_graph.py
```

## Checklist

- [x] PR title is short and non-generic.
- [x] PR only implements a single fix.
- [x] Tests passing (`uv run pytest tests/unit/test_workflow_graph_edges.py`).
- [x] Lint / typecheck passing (`uv run ruff check`, `uv run basedpyright --warnings`).

---
## Summary by cubic
Fixes a canvas bug where adding both success and error edges between the same nodes would drop one. Edges now dedupe by handle so both can coexist. Fixes #2619.

- **Bug Fixes**
  - Dedupe includes `source_handle` via `WorkflowGraphService._edge_dedup_key`, allowing success and error edges between the same pair.
  - `udf` edges default missing `source_handle` to "success"; trigger edges ignore handles.
  - Deleting an edge matches on handle, so removing one does not remove the other.
  - Introduced `EdgeDedupKey` and `_stored_edge_dedup_key` to type and normalize dedup keys and to preserve malformed stored edges instead of matching them.

<sup>Written for commit e72c4725a55c2a2c6b4b26af18d9f2090634bf5c. Summary will update on new commits. <a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2710?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/tests/unit/test_workflow_graph_edges.py
+++ b/tests/unit/test_workflow_graph_edges.py
@@ -1,0 +1,358 @@
+"""Tests for workflow graph edge operations.
+
+Covers the dedup contract for `_add_edge` and `_delete_edge`: success and
+error edges between the same source/target pair must coexist independently,
+and deleting one handle must not affect the other.
+
+Regression for https://github.com/TracecatHQ/tracecat/issues/2619.
+"""
+
+import uuid
+from collections.abc import AsyncGenerator
+
+import pytest
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from tracecat.auth.types import Role
+from tracecat.db.models import Action, Workflow, Workspace
+from tracecat.identifiers.workflow import WorkflowUUID
+from tracecat.workflow.graph.service import WorkflowGraphService
+from tracecat.workflow.management.schemas import (
+    AddEdgePayload,
+    DeleteEdgePayload,
+    GraphOperation,
+    GraphOperationType,
+)
+
+pytestmark = pytest.mark.usefixtures("db")
+
+
+@pytest.fixture
+async def workflow_pair(
+    session: AsyncSession,
+    svc_workspace: Workspace,
+) -> AsyncGenerator[tuple[Workflow, Action, Action], None]:
+    """Workflow with a source action A and a target action B, no edges yet."""
+    workflow_id = uuid.uuid4()
+    workflow = Workflow(
+        id=workflow_id,
+        title="Edge Handle Workflow",
+        description="Edge handle regression fixture",
+        status="offline",
+        workspace_id=svc_workspace.id,
+        config={},
+    )
+    session.add(workflow)
+    await session.flush()
+
+    action_a = Action(
+        id=uuid.uuid4(),
+        workspace_id=svc_workspace.id,
+        workflow_id=workflow_id,
+        type="core.http_request",
+        title="Action A",
+        description="Source",
+        inputs="",
+        control_flow={},
+        position_x=0,
+        position_y=0,
+        upstream_edges=[],
+    )
+    action_b = Action(
+        id=uuid.uuid4(),
+        workspace_id=svc_workspace.id,
+        workflow_id=workflow_id,
+        type="core.http_request",
+        title="Action B",
+        description="Target",
+        inputs="",
+        control_flow={},
+        position_x=100,
+        position_y=0,
+        upstream_edges=[],
+    )
+    session.add_all([action_a, action_b])
+    await session.commit()
+    await session.refresh(workflow, ["actions"])
+
+    try:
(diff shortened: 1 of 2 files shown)


daryllimyt:
Thanks for this! This is superseded by #2711, which carries the fix for #2619 with the same behavior plus tighter edge-key typing/normalization.

Issue: #2619
Replacement PR: #2711

wtfashwin:
Thanks for the heads up @daryllimyt - I've just pushed `e72c4725a` to this branch, which adopts the same tighter approach as #2711: an `EdgeDedupKey` dataclass with typed `EdgeSourceType`/`EdgeHandle` aliases, a dedicated `_stored_edge_dedup_key` classmethod that preserves malformed stored edges instead of matching them via an empty-string sentinel, and the same handle-aware add/delete dedup behavior.

Happy to defer to #2711 if you'd prefer to land that one - just wanted to offer this as an alternative now that the implementation is at parity.

daryllimyt:
Superseded https://github.com/TracecatHQ/tracecat/pull/2711

PR #2759, opened 2026-05-23
fix(auth): allow API keys to manage workspace variables and case attachments

mcm:
## Summary

The `/variables/*` and `/cases/{id}/attachments/*` routes are declared with `WorkspaceUserRouteRole`, which sets `allow_api_key=False`. The auth dispatcher in `tracecat/auth/credentials.py:892-907` raises 403 Forbidden when a bearer token comes in as an API key but the route doesn't accept API keys — *before* any scope check runs. Service accounts can never reach these routes, regardless of granted scopes.

This blocks downstream integrations from:
- writing per-workspace connection variables via API key (the original symptom that surfaced this)
- listing / uploading / downloading / deleting case attachments via API key

The `case:read`/`case:update` scopes the attachments router requires were already in `WORKSPACE_SERVICE_ACCOUNT_ASSIGNABLE_SCOPES`, so attachments was an obvious oversight. The `variable:*` scopes were absent from the allowlist, so this PR adds them too.

## Changes

- `tracecat/variables/router.py` — swap `WorkspaceUserRouteRole` → `WorkspaceActorRouteRole` on all 6 endpoints.
- `tracecat/cases/attachments/router.py` — swap `WorkspaceUserRouteRole` → `WorkspaceActorRouteRole` on all 4 endpoints.
- `tracecat/service_accounts/constants.py` — add `variable:{read,create,update,delete}` to `WORKSPACE_SERVICE_ACCOUNT_ASSIGNABLE_SCOPES` (the org allowlist inherits via union).
- `tests/unit/test_service_accounts_validation.py` — move `variable:*` scopes from the "rejects user-only scopes" parametrize lists into the "allows supported API-key scopes" lists for both workspace and org variants.

## Compatibility

- User sessions are unaffected — `WorkspaceActorRouteRole` is a superset of `WorkspaceUserRouteRole` (both have `allow_user=True`, this one just also has `allow_api_key=True`).
- Existing service accounts won't auto-gain the new `variable:*` scopes; admins need to grant them per service account.
- Workspace membership is still required (unchanged `require_workspace="yes"`).

## Test plan

- [x] `ruff check` clean on touched files
- [x] `ruff format --diff` clean on touched files
- [x] `basedpyright --warnings` clean on touched files
- [ ] CI unit/registry/temporal test groups (no local docker stack to run them against)
- [ ] Manual verification: create + read + update a workspace variable via API key with the new `variable:*` scopes
- [ ] Manual verification: list + upload + download + delete a case attachment via API key

## Related

[...]

PR #2762, opened 2026-05-23
fix(auth): allow API keys to manage workspace tables

mcm:
## Problem

Every endpoint in `tracecat/tables/router.py` was declared with `WorkspaceUserRouteRole`, which sets `allow_api_key=False`. Service-account API keys were therefore rejected with 403 before the per-endpoint `require_scope(...)` check ever ran, regardless of which `table:*` scopes the key actually held.

On top of that — and what makes this PR broader than the recent siblings — the `table:*` scopes were never added to the service-account assignable allowlist (`WORKSPACE_SERVICE_ACCOUNT_ASSIGNABLE_SCOPES`). So even after flipping the dependency, admins couldn't actually grant table access to an API key. This PR opens both halves.

## Changes

- **`tracecat/tables/router.py`** — swap `WorkspaceUserRouteRole` → `WorkspaceActorRouteRole` on all 18 endpoints (and the import). `Actor` is a superset of `User`, so user-session callers see no behavior change.
- **`tracecat/service_accounts/constants.py`** — add `table:read`, `table:create`, `table:update`, `table:delete` to `WORKSPACE_SERVICE_ACCOUNT_ASSIGNABLE_SCOPES`. `ORG_SERVICE_ACCOUNT_ASSIGNABLE_SCOPES` picks them up via union, no separate edit. The scopes themselves are already defined in `tracecat/authz/scopes.py` (present in the VIEWER / EDITOR / ADMIN preset sets).
- **`tests/unit/test_service_accounts_validation.py`** — `table:read` and `table:create` move from the workspace "rejects user-only" parametrize list into the "allows supported API-key" list, alongside the new `table:update` / `table:delete`. Same shuffle on the org-variant lists.

## Compatibility

- **User sessions:** unchanged. `Actor` accepts `User` callers the same as `User` does.
- **Existing service accounts:** unchanged. The new scopes are merely *assignable* — no key is automatically granted `table:*`. Admins must explicitly grant them, same as any other scope.
- **Migrations:** none required.

## Precedent

Follows the same route-role-swap pattern as #2759 (workspace variables and case attachments). This one is the broadest of the recent series because it adds entirely new entries to the service-account assignable allowlist, rather than just flipping a dependency on routes whose scopes were already assignable.

## Test plan

- [x] `ruff check` clean on changed files
- [x] `ruff format --diff` clean on changed files
- [x] `basedpyright --warnings` reports 0 errors / 0 warnings on changed files
- [ ] CI test suite (pytest requires the full docker stack; deferred to CI)

---
## Summary by cubic
Allow service-account API keys to manage workspace tables by enabling API-key access on tables routes and making `table:*` scopes assignable. User sessions are unchanged; admins can now grant `table:read|create|update|delete` to service accounts.

- **Bug Fixes**
  - Swapped `WorkspaceUserRouteRole` → `WorkspaceActorRouteRole` on all `/tables` endpoints so API keys aren’t rejected before scope checks.
  - Added `table:read`, `table:create`, `table:update`, `table:delete` to `WORKSPACE_SERVICE_ACCOUNT_ASSIGNABLE_SCOPES` (org allowlist picks them up via union).
  - Updated scope-assignability tests to reflect the new table scopes.

<sup>Written for commit 33a2653788374c8a6112729298daae4664dad267. Summary will update on new commits. <a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2762?utm_source=github">Review in cubic</a></sup>


daryllimyt:
@codex review

daryllimyt (approved):

PR #2801, opened 2026-06-01
feat: support disabling TLS verification for blob storage

### LINKED ISSUE #2539: Support s3 without ssl verification
Tracecat is now compatible with self hosted s3 compatible storage
Sometime SSL is not available on the endpoint

**Describe the solution you'd like**
I would expect to be able to disable ssl verification when connect to s3

**Describe alternatives you've considered**
For example we could have TRACECAT__S3_SSL_VERIFY=false
and use it to handle the communication with s3 bucket

bmorros94:
## Summary

Adds a `TRACECAT__BLOB_STORAGE_SSL_VERIFY` setting (default `true`) and threads it into both S3 client paths in `get_storage_client()`, so self-hosted, S3-compatible blob storage that terminates TLS with a self-signed or otherwise unverifiable certificate can be used.

Closes #2539.

## Why

Self-hosted S3-compatible storage is sometimes served over an endpoint with a self-signed or absent TLS certificate. There was previously no way to disable certificate verification for the blob-storage client, so those deployments could not connect.

## Changes

- **`tracecat/config.py`** — new `TRACECAT__BLOB_STORAGE_SSL_VERIFY` (bool, default `true`), following the existing `TRACECAT__BLOB_STORAGE_*` convention.
- **`tracecat/storage/blob.py`** — pass `verify=` to both `session.client("s3", ...)` calls (the custom-endpoint/MinIO path and the default AWS path).
- **`.env.example`** — document the new variable.
- **`tests/unit/test_storage_blob.py`** — update the two existing client tests for the new keyword argument, and add tests covering the disabled path on both branches.

## Notes

- Default is `true`, so existing behavior is unchanged.
- I used `TRACECAT__BLOB_STORAGE_SSL_VERIFY` to match the existing `TRACECAT__BLOB_STORAGE_*` naming rather than the `TRACECAT__S3_SSL_VERIFY` suggested in the issue — happy to rename if you would prefer.

---
## Summary by cubic
Adds a TLS verification toggle for blob storage via `TRACECAT__BLOB_STORAGE_SSL_VERIFY` (default true) and only overrides verification when disabled, so self-hosted S3/MinIO endpoints with self-signed certs can connect.

- **New Features**
  - Pass `verify=` to both S3 client paths in `get_storage_client()` (custom endpoint and AWS).
  - Document `TRACECAT__BLOB_STORAGE_SSL_VERIFY` in `.env.example`; add tests for on/off.
  - Default behavior unchanged; set `TRACECAT__BLOB_STORAGE_SSL_VERIFY=false` to disable verification.

<sup>Written for commit a1b18f54f639aab99977f24f295aabac231908c7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2801?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -121,6 +121,9 @@ TRACECAT__BLOB_STORAGE_BUCKET_REGISTRY=tracecat-registry
 TRACECAT__BLOB_STORAGE_BUCKET_AGENT=tracecat-agent
 # Total request attempts (initial try + retries) for transient S3/MinIO failures.
 TRACECAT__BLOB_STORAGE_MAX_ATTEMPTS=5
+# Verify TLS certificates when connecting to blob storage (S3/MinIO).
+# Set to false for self-hosted S3-compatible storage with self-signed/absent certs.
+TRACECAT__BLOB_STORAGE_SSL_VERIFY=true
 
 # --- MinIO ---
 MINIO_ROOT_USER=minio
--- a/tests/unit/test_storage_blob.py
+++ b/tests/unit/test_storage_blob.py
@@ -91,7 +91,7 @@ class TestS3Operations:
                 "s3",
                 endpoint_url="http://localhost:9002",
                 config=blob_module._STORAGE_CLIENT_CONFIG,
-                verify=True,
+                verify=None,
                 aws_access_key_id="minioadmin",
                 aws_secret_access_key="minioadmin",
             )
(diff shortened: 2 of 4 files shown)


frederick-baldo-ext (approved):
Sounds good

PR #2807, opened 2026-06-03
feat(integrations): mcp catalog overhaul

jordan-umusu:
## Summary by cubic
Overhauled MCP integrations with a platform catalog and a guided connect flow. Adds catalog APIs, typed specs/models, stricter OAuth and catalog validation, centralized icons, and a faster connect UI; integrations are now tracked by `catalog_slug`.

- **New Features**
  - Platform MCP catalog: merges OSS entries with `tracecat-ee` private overlay; statuses (available/coming_soon/deprecated/hidden); strict slug lookup and bind validation; resilient loader cache; typed connection specs/credentials/options and typed raw models (MCP and agent catalogs).
  - API: list catalog and connect/disconnect; connect returns `connected` or `oauth_redirect` with `auth_url` and `provider_id`; `catalog_slug` stored on integrations; `server_type` can be changed on update (clears incompatible fields); reads include integration `state`.
  - OAuth: prefers stored OAuth endpoints; dynamic client registration with extra trusted hosts when registration uses a different host; strict URL/host/IP validation and cross-host checks; clearer discovery errors with guidance to use a manual OAuth integration; dropped unbounded catalog `allowed_hosts`; clears stale `token_endpoint_auth_method` on update; hardened connect validation against mismatched catalog binds.
  - UI/Client: new catalog page and revamped integration dialog with quick connect, spec-driven field templates, and OAuth redirect; centralized provider icons via `mcp-catalog-icons` and `getMcpProviderIconId`; generated types/services and a `useConnectMcpIntegration` hook; improved error copy for OAuth discovery failures.

- **Migration**
  - Run DB migrations (adds `catalog_slug` + index to `mcp_integration`; adds `token_endpoint_auth_method` to `oauth_integration`).
  - Existing MCP integrations keep working.
  - For private catalog entries, ensure `tracecat-ee` is installed.

<sup>Written for commit cb214cf0d618c7f49dd2577fc4b1bf7b6c605078. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2807?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tracecat/integrations/catalog/loader.py
```diff
@@ -0,0 +1,465 @@
...
+    raw_default = row.get("default_connection_option", row.get("default_option"))
+    if isinstance(raw_default, str) and raw_default.strip():
+        for option in connection_options:
+            if option["id"] == raw_default.strip():
+                return option["connection_spec"]
+
+    return connection_options[0]["connection_spec"]
+
+
+def _load_catalog_document(
```
daryllimyt: might be worth caching here

tracecat/integrations/service.py
```diff
@@ -333,6 +398,795 @@ async def list_integrations(
...
+        if parsed.fragment:
+            raise ValueError("MCP OAuth resource URI cannot include a fragment")
+
+        host = parsed.hostname.lower()
+        # ``urlparse`` strips the brackets from IPv6 literals; restore them so
+        # the rebuilt netloc stays a valid authority (e.g. ``[::1]:443``).
+        if ":" in host:
+            host = f"[{host}]"
+        netloc = f"{host}:{parsed.port}" if parsed.port else host
+        path = parsed.path if parsed.path and parsed.path != "/" else ""
```
daryllimyt: From my local Codex review:

**[P2] Preserve trailing slash in MCP resource URIs**

For generic/catalog OAuth rows whose server URI is the root path with an explicit trailing slash, such as the private `wiz-mcp` catalog row (`https://mcp.app.wiz.io/`), this normalizes the OAuth `resource` to `https://mcp.app.wiz.io`. The existing `WizMCPProvider` overrides this specifically because Wiz requires the slash, so direct catalog OAuth redirects/token exchanges for that row will fail.

tracecat/integrations/service.py
```diff
@@ -333,6 +398,795 @@ async def list_integrations(
...
+
+    @staticmethod
+    def _oauth_authorization_server_metadata_urls(issuer: str) -> list[str]:
+        issuer = issuer.rstrip("/")
+        parsed = urlparse(issuer)
+        if parsed.scheme.lower() != "https" or not parsed.netloc:
+            return []
+        if "/.well-known/oauth-authorization-server" in parsed.path:
+            return [issuer]
+        return [f"{issuer}/.well-known/oauth-authorization-server"]
```
daryllimyt: From my local Codex review:

**[P2] Build RFC 8414 metadata URLs before issuer paths**

When protected-resource metadata advertises an authorization server issuer with a path, for example `https://auth.example.com/tenant`, the RFC 8414 well-known URL is `https://auth.example.com/.well-known/oauth-authorization-server/tenant`, not `https://auth.example.com/tenant/.well-known/oauth-authorization-server`. This will make discovery fail for path-scoped issuers used by multi-tenant auth servers.

frontend/src/app/workspaces/[workspaceId]/mcp-servers/page.tsx
```diff
@@ -1,91 +1,338 @@
...
+        : integration.server_uri || "Custom MCP server"),
+    category: CUSTOM_CATEGORY,
+    status: "available",
+    icon_url: null,
+    docs_url: null,
+    provider_id: "custom",
+    connection_spec: connectionSpecFromIntegration(integration),
+    connection_options: [],
+    locked: false,
+    state: "connected",
```
Comment (daryllimyt) [inline frontend/src/app/workspaces/[workspaceId]/mcp-servers/page.tsx]: From my local Codex review:

**[P2] Don't mark all workspace OAuth MCP rows connected**

For workspace-authored HTTP OAuth2 MCP integrations whose backing OAuth integration is only configured or has been disconnected, the new page no longer loads OAuth integration statuses but this adapter always sets `state: \"connected\"`. Those rows now render with a Disconnect action that deletes the MCP row instead of showing them as configured/reconnectable, regressing the previous status check against the OAuth integration.

frontend/src/app/workspaces/[workspaceId]/mcp-servers/page.tsx
```diff
@@ -207,247 +410,382 @@ export default function McpServersPage() {
...
+      return
+    }
+
+    if (!canCreateMcp) {
+      return
+    }
+
+    if (
+      entry.connection_spec?.requires_config &&
+      !isProviderBackedOAuth(entry)
```
Comment (daryllimyt) [inline frontend/src/app/workspaces/[workspaceId]/mcp-servers/page.tsx]: From my local Codex review:

**[P2] Open configuration for provider-backed required specs**

For provider-backed catalog rows that still require catalog configuration, such as `github-mcp` with required OAuth client credentials, this condition skips the catalog config dialog and calls the direct connect endpoint. The backend rejects those rows with \"requires configuration before connect\", while the Configure action opens the legacy provider dialog instead of the catalog form, leaving the Connect CTA broken for these entries.

tracecat/integrations/service.py
```diff
@@ -1379,10 +2589,16 @@ def _is_platform_managed_mcp_integration(
...
-        can create their own MCP rows backed by the same OAuth integration.
+        ``_auto_create_mcp_integration_if_needed`` or created from catalog
+        recipes using an exact catalog slug.
         """
+        catalog_slugs = {
+            entry["slug"]
+            for entry in get_platform_mcp_catalog_entries(include_private=True)
+        }
+        if mcp_integration.slug in catalog_slugs:
+            return True
```
daryllimyt: From my local Codex review:

**[P2] Don't classify platform rows by slug alone**

If a workspace already had a custom MCP integration whose slug matches one of the newly added catalog slugs before this release, this early return marks it platform-managed even though it was not created from the catalog. The row then disappears from `source=workspace` lists and can make the catalog report the wrong custom server as connected; avoiding collisions for newly created rows does not protect existing data.

tracecat/integrations/catalog/loader.py
```diff
@@ -0,0 +1,465 @@
...
+            any(package["command"] in ALLOWED_MCP_COMMANDS for package in packages)
+            or raw_spec.get("stdio_command") in ALLOWED_MCP_COMMANDS
+        )
+        if not has_launchable_command:
+            base["requires_config"] = True
+        if auth_type == MCPAuthType.OAUTH2:
+            base["scopes"] = raw_spec.get("scopes") or []
+
+    try:
+        return _CONNECTION_SPEC_ADAPTER.validate_python(base).model_dump(mode="json")
```
daryllimyt: **Refactor (non-blocking): keep the validated `MCPConnectionSpec` instead of dumping it back to a dict**

`_normalize_connection_spec` validates `base` into the union, then immediately `.model_dump(mode="json")`s it back to `dict[str, Any]` — discarding the type we just earned and forcing `.get()`/`cast()` on every consumer.

```python
# before
def _normalize_connection_spec(row: dict[str, Any]) -> dict[str, Any] | None:
    ...
    return _CONNECTION_SPEC_ADAPTER.validate_python(base).model_dump(mode="json")

# after
def _normalize_connection_spec(row: dict[str, Any]) -> MCPConnectionSpec | None:
    ...
    return _CONNECTION_SPEC_ADAPTER.validate_python(base)   # no .model_dump()
```

Pair with typing the TypedDict fields (`loader.py:48-49`):

```python
# before
connection_spec: NotRequired[dict[str, Any] | None]
connection_options: NotRequired[list[dict[str, Any]] | None]
# after
connection_spec: NotRequired[MCPConnectionSpec | None]
connection_options: NotRequired[list[MCPConnectionOption] | None]
```

Then the casts in `catalog/service.py:235-241` delete themselves:

```python
# before
connection_spec = cast(MCPConnectionSpec | None, entry.get("connection_spec") if agent_addons_entitled else None)
connection_options = cast(list[MCPConnectionOption], (entry.get("connection_options") or []) if agent_addons_entitled else [])
# after
connection_spec = entry["connection_spec"] if agent_addons_entitled else None
connection_options = entry["connection_options"] if agent_addons_entitled else []
```

Downstream `spec.get("server_uri")` becomes `spec.server_uri`, and basedpyright narrows on `spec.kind`. These three land together (the cast removal won't typecheck without the field retype).

tracecat/integrations/catalog/loader.py
```diff
@@ -0,0 +1,465 @@
...
+            "description": credential["description"],
+            "target": credential["target"],
+            "required": credential["required"],
+            "secret": credential["secret"],
+        }
+        for credential in credentials
+    ]
+
+
+def _normalize_connection_spec(row: dict[str, Any]) -> dict[str, Any] | None:
```
daryllimyt: **Refactor (non-blocking): let a raw input model absorb the `.get()`/`isinstance` guards**

The two-level `not isinstance(...) or ... not in {...}` checks plus per-line `raw_spec.get(...)` collapse into one input model:

```python
class RawConnectionSpec(BaseModel):
    """Untrusted catalog JSON. Coerces + drops unknowns."""
    model_config = ConfigDict(extra="ignore")

    server_type: Literal["http", "stdio"]
    auth_type: MCPAuthType
    server_uri: str | None = None
    credentials: list[RawCredential] = Field(default_factory=list)
    scopes: list[str] = Field(default_factory=list)
    oauth_authorization_endpoint: str | None = None
    oauth_token_endpoint: str | None = None
    stdio_command: str | None = None
    stdio_args: list[str] = Field(default_factory=list)
    stdio_env: list[str] = Field(default_factory=list)
    packages: list[MCPPackageOption] = Field(default_factory=list)
```

Body then reads `raw.server_uri` / `raw.credentials` instead of guarded `.get()` per line; construct in try/`ValidationError` → `None` to keep "skip invalid row".

Same theme for the sibling normalizers (`_normalize_credentials` :98, `_normalize_packages` :131, `_normalize_connection_options` :242) — they rebuild shapes already modeled by `MCPConnectionCredential`, `MCPPackageOption`, `MCPConnectionOption`. Parse into those (per-item try/validate) rather than returning `list[dict[str, Any]]`.

tracecat/integrations/service.py
```diff
@@ -333,6 +398,795 @@ async def list_integrations(
...
+        await validate_oauth_endpoint_resolves_public_async(url)
+        async with httpx.AsyncClient() as client:
+            response = await client.get(url, timeout=10.0)
+        if response.status_code == 404:
+            return None
+        response.raise_for_status()
+        data = response.json()
+        return data if isinstance(data, dict) else None
+
+    async def _discover_mcp_oauth_endpoints(
```
daryllimyt, line 736: **Refactor (non-blocking): type the OAuth wire formats with a shared model**

`_fetch_oauth_json` returns `dict[str, object] | None`, then this function does ~20 lines of guarded `.get()` (`:677-697`); `providers/base.py:894`/`:993` even uses raw `discovery_doc["..."]` indexing (KeyError on bad input). One shared model:

```python
class OAuthServerMetadata(BaseModel):
    """Authorization-server / protected-resource metadata (untrusted, RFC 8414/9728)."""
    model_config = ConfigDict(extra="ignore")

    resource: str | None = None
    authorization_servers: list[str] = Field(default_factory=list)
    authorization_endpoint: str | None = None
    token_endpoint: str | None = None
    registration_endpoint: str | None = None
    token_endpoint_auth_methods_supported: list[str] = Field(default_factory=list)

    @field_validator("authorization_servers", "token_endpoint_auth_methods_supported", mode="before")
    @classmethod
    def _only_strings(cls, v: object) -> list[str]:
        """Drop malformed (non-string) list items rather than reject the doc."""
        return [i for i in v if isinstance(i, str)] if isinstance(v, list) else []

    @property
    def is_complete(self) -> bool:
        return bool(self.authorization_endpoint and self.token_endpoint)
```

Call sites simplify:

```python
# before
if "authorization_endpoint" in metadata and "token_endpoint" in metadata:
    direct_metadata = metadata
token_methods_raw = direct_metadata.get("token_endpoint_auth_methods_supported", [])
token_methods = [m for m in token_methods_raw if isinstance(m, str)] if isinstance(token_methods_raw, list) else []
# after
if meta.is_complete:
    direct_metadata = meta
token_methods = meta.token_endpoint_auth_methods_supported
```

**Keep out of the model:** the "endpoints required" check (context-dependent — resource metadata legitimately has only `authorization_servers`) stays at the call site; and `_validate_mcp_oauth_endpoint(..., base_domain=...)` stays explicit — that's the SSRF hardening from fb5f1ed3e, don't bury it in a validator/context.

Also reuse the existing `TokenResponse` model at `service.py:1108`/`:1172` instead of `cast(dict, ...).get("access_token")`, and a small `DCRResponse` for `:752` + `providers/base.py:414`.

tracecat/integrations/catalog/loader.py
```diff
@@ -0,0 +1,465 @@
...
+    catalog_data = _catalog_data(include_private=include_private)
+    if not catalog_data:
+        return []
+
+    servers = catalog_data.get("servers")
+    if not isinstance(servers, list):
+        return []
+
+    entries: list[PlatformMCPCatalogEntry] = []
+    for index, row in enumerate(servers):
```
daryllimyt: **Refactor (non-blocking): type the catalog row envelope too**

Companion to the `RawConnectionSpec` note above. The row *envelope* here is still parsed with `row.get()` + an 8-clause required-field guard. A `RawCatalogRow` model finishes the job:

```python
class RawCatalogRow(BaseModel):
    """One catalog server entry from the bundled JSON (untrusted)."""
    model_config = ConfigDict(extra="ignore")

    slug: str = Field(min_length=1)
    name: str = Field(min_length=1)
    description: str = Field(min_length=1)
    category: str = Field(min_length=1)

    status: PlatformMCPCatalogStatus | None = None
    icon: str | None = None
    docs: str | None = None
    provider_id: str | None = None

    # private overlay (consumed when include_private); fed to the normalizers
    connection_spec: dict[str, Any] | None = None
    metadata: dict[str, Any] | None = None
    connection_options: list[dict[str, Any]] | None = None
    options: list[dict[str, Any]] | None = None

    @field_validator("status", mode="before")
    @classmethod
    def _drop_unknown_status(cls, v: object) -> object:
        """Unknown status falls back to coming_soon downstream, not a hard reject."""
        return v if v in {"available", "coming_soon", "deprecated", "hidden"} else None

    @property
    def has_connection_metadata(self) -> bool:
        return bool(self.connection_spec or self.metadata
                    or self.connection_options or self.options)
```

The loop body then drops the guard wall:

```python
# before
for index, row in enumerate(servers):
    if not isinstance(row, dict):
        continue
    raw_slug = row.get("slug")
    raw_name = row.get("name")
    raw_description = row.get("description")
    raw_category = row.get("category")
    if not (
        isinstance(raw_slug, str) and raw_slug
        and isinstance(raw_name, str) and raw_name
        and isinstance(raw_description, str) and raw_description
        and isinstance(raw_category, str) and raw_category
    ):
        continue
    ...
    raw_provider_id = row.get("provider_id")
    provider_id = raw_provider_id if isinstance(raw_provider_id, str) else None
    raw_docs_url = row.get("docs")
    docs_url = raw_docs_url if isinstance(raw_docs_url, str) else None
    raw_status = row.get("status")
    if include_private and connection_spec is not None:
        status = "available"
    elif raw_status in {"available", "coming_soon", "deprecated", "hidden"}:
        status = raw_status
    else:
        status = "coming_soon"
    icon_url = row.get("icon")
    entry = {... "icon_url": icon_url if isinstance(icon_url, str) else None, ...}

# after
for index, raw in enumerate(servers):
    try:
        row = RawCatalogRow.model_validate(raw)   # absorbs the 8-clause guard
    except ValidationError:
        continue   # missing/blank slug/name/description/category => skip

    connection_options: list[MCPConnectionOption] = []
    connection_spec: MCPConnectionSpec | None = None
    provider_id = docs_url = None
    if include_private:
        connection_options = _normalize_connection_options(row)
        connection_spec = _default_connection_spec(row, connection_options)
        if row.has_connection_metadata and connection_spec is None:
            logger.warning("Skipping invalid MCP catalog connection metadata",
                           extra={"slug": row.slug})
        provider_id = row.provider_id
        docs_url = row.docs

    status = "available" if (include_private and connection_spec is not None) \
        else (row.status or "coming_soon")

    entries.append(PlatformMCPCatalogEntry(
        id=catalog_id_for_slug(row.slug),
        slug=row.slug, name=row.name, description=row.description,
        category=row.category, status=status,
        icon_url=row.icon, docs_url=docs_url, provider_id=provider_id,
        connection_spec=connection_spec,
        connection_options=connection_options or None,
        sort_key=f"{index:04d}:{row.name.lower()}",
    ))
```

The `_normalize_*` helpers take `RawCatalogRow` instead of `dict` (reading `row.connection_spec`/`row.metadata`). The overlay merge in `_catalog_data:355-364` stays a dict-level `{**row, **overlay}` — validation happens once per merged row here. Dovetails with the Group A retype: the local `connection_spec`/`connection_options` carry their types straight into the entry, no cast.

daryllimyt (changes requested):

tracecat/integrations/service.py
jordan-umusu: Resolved by adding a new col on mcp_integrations catalog_slug as a reference to platform entries.

tracecat/integrations/catalog/types.py
```diff
@@ -0,0 +1,161 @@
...
+class RawConnectionOption(BaseModel):
+    """Untrusted connect option row from catalog JSON."""
+
+    model_config = ConfigDict(extra="ignore")
+
+    id: Any = None
+    label: Any = None
+    description: Any = None
+    docs: Any = None
+    docs_url: Any = None
```
daryllimyt: why do these still have to be `Any`?

tracecat/integrations/catalog/loader.py
```diff
@@ -0,0 +1,461 @@
...
+        OSError,
+    ):
+        if required:
+            return {}
+        return None
+
+    return catalog_data if isinstance(catalog_data, dict) else {}
+
+
+def _catalog_data(*, include_private: bool) -> dict[str, Any]:
```
daryllimyt, line 235: make this typed?

tracecat/integrations/catalog/loader.py
```diff
@@ -0,0 +1,461 @@
...
+
+def _load_catalog_document(
+    package: str,
+    resource: str,
+    *,
+    required: bool,
+) -> dict[str, Any] | None:
+    try:
+        raw = resources.files(package).joinpath(resource).read_bytes()
+        catalog_data = orjson.loads(raw)
```
daryllimyt, line 223: use a basemodel?

tracecat/integrations/catalog/loader.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+def _default_connection_spec(
+    row: RawCatalogRow, connection_options: list[MCPConnectionOption]
+) -> MCPConnectionSpec | None:
+    if not connection_options:
+        return _normalize_connection_spec(row)
+
+    raw_default = row.default_connection_option or row.default_option
+    if isinstance(raw_default, str) and raw_default.strip():
```
daryllimyt: should early return with null check instead of typechecking for str

tracecat/integrations/catalog/types.py
```diff
@@ -0,0 +1,161 @@
...
+from tracecat.integrations.enums import MCPAuthType
+from tracecat.integrations.schemas import (
+    MCPConnectionOption,
+    MCPConnectionSpec,
+    PlatformMCPCatalogStatus,
+)
+from tracecat.integrations.types import MCPServerType
+
+
+class PlatformMCPCatalogEntry(TypedDict):
```
daryllimyt: is there a need for this to be a TypedDict over BaseModel?

tracecat/integrations/catalog/service.py
```diff
@@ -0,0 +1,265 @@
...
+                .outerjoin(
+                    OAuthIntegration,
+                    OAuthIntegration.id == MCPIntegration.oauth_integration_id,
+                )
+                .where(MCPIntegration.workspace_id == workspace_id)
+            )
+        ).all()
+
+        state_by_catalog_id: dict[uuid.UUID, tuple[MCPIntegration, bytes | None]] = {}
+        for mcp_integration, encrypted_access_token, provider_id, catalog_slug in rows:
```
daryllimyt, line 195: This unpacking comes out as `Any` for all four targets — not a pyrefly quirk, but how SQLAlchemy types `Row`.

The element types *are* tracked: `.all()` returns `Sequence[Row[tuple[MCPIntegration, bytes, str, str | None]]]`. But tuple-unpacking in the `for` loop goes through `Row.__iter__`, which the stubs annotate as `Iterator[Any]`, so every target collapses to `Any`.

Fix: add `.tuples()` to the result so iteration yields real `tuple`s instead of `Row`s:

```python
rows = (
    (await self.session.execute(
        sa.select(
            MCPIntegration,
            OAuthIntegration.encrypted_access_token.label("encrypted_access_token"),
            OAuthIntegration.provider_id.label("provider_id"),
            MCPIntegration.catalog_slug.label("catalog_slug"),
        )
        .outerjoin(
            OAuthIntegration,
            OAuthIntegration.id == MCPIntegration.oauth_integration_id,
        )
        .where(MCPIntegration.workspace_id == workspace_id)
    ))
    .tuples()
    .all()
)
```

Then the loop unpacks as `mcp_integration: MCPIntegration`, `encrypted_access_token: bytes`, `provider_id: str`, `catalog_slug: str | None` (verified via `reveal_type`).

Note: `encrypted_access_token` types as `bytes`, not `bytes | None`, even though it's an OUTER JOIN and can be NULL — the type system can't tell the join is outer. The `bytes | None` annotation on the dict value (already present) is the right call.

tracecat/integrations/service.py
```diff
@@ -1445,12 +2843,40 @@ async def update_mcp_integration(
...
+            "oauth_integration_id" in params.model_fields_set
+        )
+
+        if target_server_type == "http":
+            target_server_uri = params.server_uri or mcp_integration.server_uri
+            if not target_server_uri:
+                raise ValueError("server_uri is required for http-type servers")
+
+            target_auth_type = (
+                params.auth_type
```
daryllimyt: hard to read

tracecat/integrations/service.py
```diff
@@ -1445,12 +2843,40 @@ async def update_mcp_integration(
...
+                if params.auth_type is not None
+                else MCPAuthType.NONE
+                if server_type_changed
+                else mcp_integration.auth_type
+            )
+            target_oauth_integration_id = (
```
daryllimyt: hard to read

daryllimyt (changes requested):

tracecat/integrations/catalog/loader.py
```diff
@@ -0,0 +1,381 @@
...
+        catalog_data = orjson.loads(raw)
+    except (
+        FileNotFoundError,
+        IsADirectoryError,
+        ModuleNotFoundError,
+        orjson.JSONDecodeError,
+        OSError,
+    ):
+        return None
+    return catalog_data if isinstance(catalog_data, dict) else None
```
daryllimyt, line 232: what's the reason why we can't just validate this using a pydantic model, and have to pass it around below as a dict?

daryllimyt (approved):

tracecat/integrations/catalog/loader.py
jordan-umusu: thanks for spotting this, made it a pydantic model.

tracecat/integrations/service.py
```diff
@@ -1264,12 +2304,69 @@ async def _mcp_integration_slug_taken(self, slug: str) -> bool:
...
+        specs: list[MCPConnectionSpec] = []
+        if catalog_entry.connection_spec is not None:
+            specs.append(catalog_entry.connection_spec)
+        specs.extend(
+            option.connection_spec for option in catalog_entry.connection_options or []
+        )
+        for spec in specs:
+            if spec.server_type != params.server_type:
+                continue
+            if params.server_type == "stdio" or spec.auth_type == params.auth_type:
```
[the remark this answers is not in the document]
jordan-umusu: Valid concern but many providers expect tenant-specific URIs. Can be addressed in a followup.

PR #2837, opened 2026-06-09
feat(webhooks): add include_headers toggle to webhook trigger

topher-lo:
## Summary

Adds a per-webhook `include_headers` toggle. When enabled, the workflow `TRIGGER` receives the full request envelope — mirroring the shape of `core.http.request` — instead of just the parsed body:

```json
{
  "status_code": 200,
  "headers": { "content-type": "application/json", "x-custom-source": "my-siem" },
  "data": { "...the request body..." }
}
```

When the toggle is off (the default), behavior is unchanged: `TRIGGER` is the parsed body only. This lets workflows read signature headers, custom auth headers, delivery IDs, etc.

The webhook's own `x-tracecat-api-key` header is stripped from the exposed `headers` so workflow authors can't read the webhook's auth secret. NDJSON batches wrap each execution's payload in the same envelope.

## Changes

- DB: add `webhook.include_headers boolean not null default false` + Alembic migration `c9d2e7a4f1b3`.
- API schemas: add `include_headers` to `WebhookRead`/`WebhookCreate`/`WebhookUpdate`.
- Webhook handling: `validate_incoming_webhook` stashes the flag on `request.state`; `_incoming_webhook` wraps the payload via a small `_wrap_with_headers` helper (explicit param, defaults off).
- UI: add an "Include request headers" switch to the webhook trigger panel, mirroring the existing status toggle.
- Tests: cover on/off behavior, `x-tracecat-api-key` stripping, and NDJSON batch wrapping.

The existing `get_webhook`/`update_webhook` handlers surface and persist the field automatically.

## Verification

- `uv run pytest tests/unit/test_webhook_execution_path.py tests/unit/test_webhook.py` — 56 passed.
- `ruff check` / `ruff format --check` / `basedpyright --warnings` clean; `pnpm -C frontend check` + `typecheck` clean.
- Migration applied to a local stack; column confirmed.
- Live curl end-to-end (trigger read back from Temporal):
  - ON -> `{status_code, headers, data}` with the API-key header stripped and a custom header preserved.
  - OFF -> raw body, unchanged.
- UI toggle persisted `include_headers = true` via the real `PATCH` (frontend -> API -> DB).

---
## Summary by cubic
Adds a per-webhook `include_headers` toggle to send the full request envelope {status_code, headers, data} to TRIGGER instead of only the body. Also preserves this flag through workflow store export/import and stops logging wrapped headers in `/wait` and `/draft`.

- **New Features**
  - Backend: `include_headers` boolean; when on, TRIGGER receives `{status_code, headers, data}` (mirrors `core.http.request`) and filters sensitive headers (strips `x-tracecat-api-key`).
  - API/UI: added `include_headers` to Webhook Read/Create/Update; added “Include request headers” switch in the webhook trigger panel.
  - DB: adds `webhook.include_headers boolean not null default false`.

- **Bug Fixes**
  - Endpoints: apply the header envelope to `/wait` and `/draft` via a shared payload dependency; stop logging wrapped payloads to avoid leaking headers; typed the trigger envelope as `WebhookTriggerEnvelope` (no behavior change).
  - API: `WebhookRead` coerces NULL `include_headers` to false; `create_webhook` persists `include_headers`.
  - Workflow Store: round-trip `include_headers` via `RemoteWebhook` (export/import); defaults to false for older exports.
  - UI: toggle rolls back to the previous value on update error.

<sup>Written for commit bb1cfebf03cdd6e35c1c1b1e370dc2c488a3c321. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2837?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/alembic/versions/c9d2e7a4f1b3_add_webhook_include_headers.py
+++ b/alembic/versions/c9d2e7a4f1b3_add_webhook_include_headers.py
@@ -1,0 +1,35 @@
+"""add webhook include_headers
+
+Revision ID: c9d2e7a4f1b3
+Revises: 0f18bea0c115
+Create Date: 2026-06-09 00:00:00.000000
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "c9d2e7a4f1b3"
+down_revision: str | None = "0f18bea0c115"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "webhook",
+        sa.Column(
+            "include_headers",
+            sa.Boolean(),
+            nullable=False,
+            server_default=sa.text("false"),
+        ),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("webhook", "include_headers")
--- a/frontend/src/client/schemas.gen.ts
+++ b/frontend/src/client/schemas.gen.ts
@@ -27229,6 +27229,11 @@ export const $WebhookCreate = {
       type: "array",
       title: "Allowlisted Cidrs",
     },
+    include_headers: {
+      type: "boolean",
+      title: "Include Headers",
+      default: false,
+    },
   },
   type: "object",
   title: "WebhookCreate",
@@ -27283,6 +27288,11 @@ export const $WebhookRead = {
       type: "array",
       title: "Methods",
       description: "Methods to allow",
+    },
+    include_headers: {
+      type: "boolean",
+      title: "Include Headers",
+      default: false,
     },
     workflow_id: {
       type: "string",
@@ -27414,6 +27424,17 @@ export const $WebhookUpdate = {
         },
       ],
       title: "Allowlisted Cidrs",
+    },
+    include_headers: {
+      anyOf: [
+        {
+          type: "boolean",
+        },
+        {
+          type: "null",
+        },
+      ],
+      title: "Include Headers",
     },
   },
   type: "object",
--- a/frontend/src/client/types.gen.ts
+++ b/frontend/src/client/types.gen.ts
@@ -8140,6 +8140,7 @@ export type WebhookCreate = {
   methods?: Array<WebhookMethod>
   entrypoint_ref?: string | null
   allowlisted_cidrs?: Array<string>
+  include_headers?: boolean
 }
 
 export type WebhookMethod = "GET" | "POST"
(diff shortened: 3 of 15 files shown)


topher-lo:
Tested live with chrome devtools 
Screenshot 2026-06-09 at 12 04 19 PM
Screenshot 2026-06-09 at 12 01 12 PM

topher-lo:
Addressed the CI failure and the review feedback in `caf01166d`:

- **CI (`test_get_workflow_*` 500s):** `WebhookRead.include_headers` now coerces `None → False` via a before-validator, matching the existing `methods`/`allowlisted_cidrs`/`filters` validators. The failures came from the in-memory `mock_webhook` fixture leaving the unflushed ORM attribute as `None`.
- **codex + cubic — `/wait` ignored the flag:** extracted a shared `_wrapped_payload` dependency and applied it to both `/wait` and `/draft`, so `TRIGGER` has the same `{status_code, headers, data}` shape across all webhook trigger endpoints (root POST/GET keeps its per-NDJSON-item wrapping).
- **codex — `create_webhook` dropped the field:** now passes `include_headers=params.include_headers` into the `Webhook(...)` constructor.
- **cubic — stale rollback:** the UI error path now restores the pre-toggle value (`!checked`) instead of the possibly-stale `webhook` prop.

Added unit coverage for the `_wrapped_payload` dependency (wrap/passthrough + api-key stripping) and the `WebhookRead` None-coercion.

topher-lo:
Both follow-up findings were valid — addressed in `b1ff382`:

- **codex + cubic — `include_headers` lost on store export/import:** added it to the `RemoteWebhook` schema, the export (`store/service.py`), and the import (`import_service.py::_update_webhook`). It defaults to `False`, so older exports without the field import cleanly. Added round-trip coverage (export carries it, import restores it, missing field → `False`).
- **cubic — `/wait` and `/draft` trace-logged wrapped headers:** those handlers now receive an already-wrapped payload via the `_wrapped_payload` dependency, so their `logger.trace` no longer logs the payload value (the adjacent `logger.info` still records the hit). The root POST/GET path is unaffected — it logs the raw body before wrapping. Real-world severity was low (trace is below DEBUG, off by default), but it's a cheap, legitimate hardening.

Note: `allowlisted_cidrs` is also not round-tripped through the store today — a pre-existing gap unrelated to this PR, so I left it out of scope. Happy to file a separate issue if you'd like.

PR #2840, opened 2026-06-09
feat(agents): agent runs in inbox

jordan-umusu:
---
## Summary by cubic
Show agent runs in the Inbox with grouped results, server-side search/sort, and per‑group cursor pagination. Replaces the federated model with a single EE provider, prevents duplicate chat in terminal sessions, and polishes the runs table UI.

- New Features
  - Backend: Single EE `AgentRunsInboxProvider`; SQL keyset pagination; global sort by `created_at`/`updated_at`; server-side search, date filters, group filter (`InboxGroup`), and entity-type filter; approvals aggregated across all chat sources; `/inbox/pending` returns 0 when unavailable.
  - API: New `agent_run` item type, `InboxGroup` enum, and `created_by`; `GET /inbox/items` adds `search`, `group`, `entityType`, `orderBy`, `createdAfter`, `updatedAfter` (removed status sort).
  - UI: Grouped `RunsTable` (Review required, Running, Error, Completed) with per‑group “Show more” via server cursors; debounced search with keep‑previous‑data; improved row interactions and timestamp tooltips; expanded entity-type options; sidebar renamed to “Inbox”; only resume live sessions to avoid chat duplication in terminal sessions.

- Migration
  - Regenerate the frontend API client for new inbox types and params.
  - Deploy `@tracecat-ee` with the new `AgentRunsInboxProvider`.

<sup>Written for commit 6da24b6e0a20bd8be3a2ddfde59abd17fb1fa62d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2840?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

image


packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
```diff
@@ -323,6 +355,16 @@ async def _enrich_sessions(
             workflows = workflow_result.scalars().all()
             workflows_by_id = {w.id: w for w in workflows}
 
+        # Fetch creators for user-initiated sessions
+        creator_ids = {s.created_by for s in sessions if s.created_by}
+        users_by_id: dict[uuid.UUID, User] = {}
+        if creator_ids:
+            user_stmt = select(User).where(
```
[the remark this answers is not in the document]
jordan-umusu: The creator id's are derived from rows that are already workspace scoped preventing one from maliciously enumerating through arbitrary users.

frontend/src/components/inbox/index.tsx
```diff
@@ -2,21 +2,23 @@
...
 import {
   type DateFilterValue,
+  type InboxGroupState,
   type UseInboxFilters,
   useDeleteApproval,
 } from "@/hooks/use-inbox"
 import type { InboxSessionItem } from "@/lib/agents"
-import { ActivityAccordion } from "./activity-accordion"
 import { InboxHeader } from "./inbox-header"
+import { RunsTable } from "./runs-table"
```
daryllimyt: nit: absolute path

packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
```diff
@@ -30,14 +31,28 @@
...
     from tracecat.auth.types import Role
 
 
-class ApprovalsInboxProvider(BaseCursorPaginator):
-    """Provides approval items for the inbox.
+class AgentRunsInboxProvider(BaseCursorPaginator):
+    """Provides agent run items for the inbox.
 
-    Filters to workflow-initiated sessions only and enriches with workflow metadata.
+    Lists root Claude Code agent sessions (plus any legacy sessions with
```
daryllimyt: keep this harness agnostic

packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
```diff
@@ -96,6 +111,50 @@ async def describe_status(
                 statuses[session_id] = status
         return statuses
 
+    def _base_query(self, search: str | None):
+        """Base statement selecting inbox-eligible root sessions."""
+        # Root sessions only: all Claude Code runs, plus legacy sessions that
```
daryllimyt: harness agnostic

packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
```diff
@@ -96,6 +111,50 @@ async def describe_status(
                 statuses[session_id] = status
         return statuses
 
+    def _base_query(self, search: str | None):
```
daryllimyt: return type?

packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
```diff
@@ -96,6 +111,50 @@ async def describe_status(
...
+        """Base statement selecting inbox-eligible root sessions."""
+        # Root sessions only: all Claude Code runs, plus legacy sessions that
+        # already have approvals so existing inbox items don't disappear.
+        has_approvals = (
+            select(Approval.id).where(Approval.session_id == AgentSession.id).exists()
+        )
+        base_stmt = select(AgentSession).where(
```
daryllimyt, line 141: Consider narrowing this ORM load with `load_only(...)` instead of selecting the full `AgentSession` entity. The grouped path can scan up to 300 sessions, and the inbox polls multiple groups, so this can repeatedly hydrate JSONB fields that are not used here (`tools`, `mcp_integrations`, `agents_binding`, `work_dir_snapshot`, and especially `artifacts`).

Using `load_only` keeps the existing `AgentSession` instances and avoids rewriting `_enrich_sessions` / `_resolve_temporal_statuses` around row tuples:

```py
from sqlalchemy.orm import load_only

base_stmt = select(AgentSession).options(
    load_only(
        AgentSession.id,
        AgentSession.title,
        AgentSession.created_by,
        AgentSession.entity_type,
        AgentSession.entity_id,
        AgentSession.curr_run_id,
        AgentSession.created_at,
        AgentSession.updated_at,
        raiseload=True,
    )
).where(...)
```

One important detail: include `AgentSession.id` explicitly. In this model `surrogate_id` is the primary key, so SQLAlchemy auto-loads `surrogate_id`, not the UUID `id`. Columns used only in filters (`workspace_id`, `parent_session_id`, `harness_type`) do not need to be loaded.

packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
```diff
@@ -269,6 +330,174 @@ async def list_items(
...
+            sessions = list(result.scalars().all())
+            if not sessions:
+                exhausted = True
+                break
+
+            scanned += len(sessions)
+            last_session = sessions[-1]
+            last_key = (getattr(last_session, sort_col), last_session.id)
+
+            items = await self._enrich_sessions(sessions)
```
daryllimyt: This grouped path does a lot of work for rows that may be discarded immediately after classification. Each batch is fully enriched first, then filtered by `_classify_item()`, so non-matching scanned sessions still pay for approval loading, workflow/user lookups, Temporal status resolution, metadata construction, and `InboxItemRead` serialization.

A lower-waste shape would be to split lightweight group classification from full enrichment:

```py
while len(matching_sessions) < limit + 1 and scanned < GROUP_SCAN_MAX_SESSIONS:
    sessions = await fetch_next_session_batch(...)

    classifications = await self._classify_sessions_for_group(
        sessions,
        group=group,
    )

    matching_sessions.extend(
        session
        for session in sessions
        if classifications[session.id] == group
    )

page_sessions = matching_sessions[:limit]
page_items = await self._enrich_sessions(page_sessions)
```

The helper can stay narrowly scoped to group membership data, for example:

```py
async def _classify_sessions_for_group(
    self,
    sessions: Sequence[AgentSession],
    *,
    group: InboxGroup,
) -> dict[uuid.UUID, InboxGroup]:
    session_ids = [s.id for s in sessions]
    pending_counts = await self._fetch_pending_approval_counts(session_ids)

    if group is InboxGroup.REVIEW_REQUIRED:
        return {
            s.id: InboxGroup.REVIEW_REQUIRED
            if pending_counts.get(s.id, 0) > 0
            else InboxGroup.COMPLETED
            for s in sessions
        }

    temporal_statuses = await self._resolve_temporal_statuses(sessions)
    return {
        s.id: self._classify_session_light(
            session=s,
            pending_count=pending_counts.get(s.id, 0),
            temporal_status=temporal_statuses.get(s.id),
        )
        for s in sessions
    }
```

The exact helper shape can vary, but the important part is that `_enrich_sessions()` remains the full item builder and runs only for the final page instead of every scanned row.

daryllimyt (changes requested):

frontend/src/hooks/use-inbox.ts
```diff
@@ -177,154 +204,222 @@ function getDateFromFilter(filter: DateFilterValue): Date | null {
...
+    sort,
+    enabled: baseEnabled,
+    autoRefresh,
+    pollMs: 10000,
+  })
 
-    const updatedAfterDate = getDateFromFilter(updatedAfter)
-    const createdAfterDate = getDateFromFilter(createdAfter)
-    const query = searchQuery.toLowerCase().trim()
+  // Client-side filtering (entity type, date filters) applied per group
```
daryllimyt: This filtering should happen server-side instead of after each group page is fetched. The server has already chosen the keyset page by the time this runs, so the client can drop all `limit` rows for `entityType` / date filters while matching sessions still exist later in the cursor stream. That makes groups look empty or short and leaves `hasMore` tied to the unfiltered page.

Please add `entity_type`, `created_after`, and `updated_after` query params to the inbox endpoint, pass them through to the provider, apply them in `_base_query()` before pagination/group scanning, and have the UI send absolute ISO timestamps for the relative date filters.

daryllimyt (changes requested):

frontend/src/components/inbox/inbox-detail.tsx
```diff
@@ -58,6 +58,19 @@ export function InboxDetail({
...
+  // when a pendingMessage is queued for it; a fork discovered on mount has no
+  // pendingMessage and may have completed earlier. For the parent run, "live"
+  // means a streaming Temporal status (RUNNING or CONTINUED_AS_NEW). Resuming a
+  // terminal session replays its last persisted turn on top of the DB-seeded
+  // history, duplicating the chat. `resume` is read once on ChatSessionPane
+  // mount (the pane is keyed by sessionId), so the later pendingMessage clear
+  // does not flip it.
+  const shouldResume = isForkedSession
+    ? pendingMessage !== undefined
+    : isLiveAgentStatus(session.derivedStatus)
```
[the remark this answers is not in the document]
jordan-umusu: valid but edge case that is better addressed with a proper followup refactor.

daryllimyt (approved):

daryllimyt (approved):
lgtm

PR #2841, opened 2026-06-09
fix(mcp): preserve org context across OAuth resume

### LINKED ISSUE #2808: Scaling mcp with more than 1 replica loose the session
**Describe the bug**
If I run Tracecat with 1 replica for mcp all is working fine
If I scale up mcp I cannot connect to mcp with 403 forbidden

**Provide logs**
logs from mcp:
IdP callback error: access_denied - User cannot be resolved to a single organization

**To reproduce**
Deploy helm chart with 2 replicas for mcp
Create a token
Try to connect from terminal (tried with copilot)

**Expected behavior**
We should be able to connect
(it works well with 1 replica)

- Tracecat version  `1.0.0-beta.48`
- Where did you deploy Tracecat? k8s
- Browser type : non (terminal)

RitwijParmar:
Fixes #2808.

This keeps MCP OAuth org context stable across login and resume, which is the path that breaks when MCP is scaled beyond one replica.

What changed:
- Resolve MCP authorize sessions from an explicit `org` hint or the active Tracecat org cookie before falling back to the current single-org behavior.
- Store the `org` hint inside the resume transaction and preserve it through `/oauth/mcp/continue` and `/authorize/resume`.
- Validate org hints and active-org cookies against the signed-in user's active org memberships before issuing the MCP auth code.
- Add endpoint and resolver coverage for org preservation, query override, cookie resolution, invalid hints, and ambiguous multi-org fallback.
- Document the multi-organization self-hosted MCP behavior.

Why this shape:
The existing resume transaction is already shared storage, so this avoids relying on sticky sessions. A resumed authorization request carries enough org context for any replica to finish the flow, while still failing closed if the user is not a member of the requested org.

Local checks:
- `uv run ruff check tracecat/mcp/oidc/session.py tracecat/mcp/oidc/endpoints.py tests/unit/test_mcp_oidc_session.py tests/unit/test_mcp_oidc_endpoints.py`
- `uv run basedpyright tracecat/mcp/oidc/session.py tracecat/mcp/oidc/endpoints.py tests/unit/test_mcp_oidc_session.py tests/unit/test_mcp_oidc_endpoints.py`
- `python3.11 -m py_compile tracecat/mcp/oidc/session.py tracecat/mcp/oidc/endpoints.py tests/unit/test_mcp_oidc_session.py tests/unit/test_mcp_oidc_endpoints.py`
- `cd frontend && bun test src/lib/auth-return-url.test.ts`
- `git diff --check`

I also tried the focused backend pytest command, but this local machine does not have Postgres running on localhost:5432, so Tracecat's session-level DB fixture fails before the test bodies run.

---
## Summary by cubic
Keeps MCP OAuth org context stable across login and resume so multi‑org auth works across replicas without sticky sessions. Rejects malformed resume state and ambiguous org hints; UUID hints are preferred over slugs.

- **Bug Fixes**
  - Pass and persist `org` through `/authorize`, `/authorize/resume`, and the resume transaction; login redirect to `/oauth/mcp/continue` carries `org`; resume uses stored `org` unless a query `org` overrides it.
  - Resolve session org via explicit `org` (UUID > slug) or the active‑org cookie; require active membership; treat blank hints as absent; only fall back to single‑org when unambiguous; ignore cookies that don’t match a current membership.
  - Validate resume transactions have required OAuth params before resuming; reject invalid hints, ambiguous slug matches, and multi‑org users without explicit org. Frontend preserves `org` in continue/sign‑in flows and the return‑URL sanitizer keeps `org` (including encoded); docs updated and tests added for redirects, hint passing, resume overrides, cookies, ambiguous hints, and malformed resume.

<sup>Written for commit 545629907d5c5ccf30e28c5fa4496513952d7dee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2841?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


topher-lo:
Have a duplicate.

PR #2878, opened 2026-06-18
fix(case-triggers): stop stale trigger replays

daryllimyt:
## Summary
- require online case triggers to target the workflow's current committed, runnable definition
- self-heal stale invalid case triggers by setting them offline when the consumer encounters missing, empty, invalid, or non-current definitions
- treat those invalid trigger configs as terminal skips so one bad trigger cannot keep a Redis case event pending after valid triggers already dispatched

## Root Cause
A stale online case trigger can point at a workflow without a usable current definition. The consumer previously treated that as a retryable dispatch failure, returned `should_ack=false`, and left the Redis stream message pending. Valid triggers for the same case event could still dispatch and write a temporary dedupe key; after that key expired, the still-pending event could replay the valid workflow.

## Validation
- `uv run ruff check --fix tracecat/workflow/case_triggers/service.py tracecat/cases/triggers/consumer.py tests/unit/test_case_trigger_consumer.py tests/unit/test_case_triggers.py tests/unit/test_workflow_import_service.py`
- `uv run ruff format --check tracecat/workflow/case_triggers/service.py tracecat/cases/triggers/consumer.py tests/unit/test_case_trigger_consumer.py tests/unit/test_case_triggers.py tests/unit/test_workflow_import_service.py`
- `uv run basedpyright tracecat/workflow/case_triggers/service.py tracecat/cases/triggers/consumer.py tests/unit/test_case_trigger_consumer.py tests/unit/test_case_triggers.py tests/unit/test_workflow_import_service.py`
- `uv run pytest tests/unit/test_case_trigger_consumer.py tests/unit/test_case_triggers.py tests/unit/test_workflow_import_service.py -k 'case_trigger'`
- `uv run pytest tests/unit/test_workflow_import_service.py`

Pre-commit also passed during commit after installing frontend dependencies for the generated-client hook.

---
## Summary by cubic
Prevents stale case trigger replays by forcing online triggers to use the workflow’s current runnable definition, and by ACKing and auto‑disabling invalid triggers to clear pending replays. External imports with online triggers now publish a create‑scoped, version‑locked initial definition so the trigger stays online.

- **Bug Fixes**
  - Consumer: If a trigger’s workflow definition is missing, non‑current, empty, or has invalid DSL, set the trigger offline, treat as a terminal skip, write the done key, and ACK the event to stop replays.
  - Service: When enabling a trigger, require the workflow’s current, published, valid definition; otherwise reject with “Publish the workflow…”.
  - Management: For external imports with an online trigger, publish the initial definition via create scope, freeze with the workflow’s `registry_lock`, set `workflow.version`, then upsert the trigger (keeps it online).

<sup>Written for commit bf096312c39e519bd2cd30d5fa81e3ca60e20534. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2878?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/unit/test_case_trigger_consumer.py
+++ b/tests/unit/test_case_trigger_consumer.py
@@ -383,6 +383,72 @@ async def test_dispatch_workflow_skips_non_current_definition(
     workflow_service_connect.assert_not_called()
 
 
+@pytest.mark.anyio
+async def test_dispatch_workflow_disables_structurally_invalid_definition(
+    monkeypatch: pytest.MonkeyPatch,
+):
+    workflow_id = uuid.uuid4()
+    workspace_id = uuid.uuid4()
+    consumer = CaseTriggerConsumer(client=AsyncMock())
+    consumer._disable_invalid_case_trigger = AsyncMock()
+    workflow_service_connect = AsyncMock()
+
+    monkeypatch.setattr(
+        "tracecat.cases.triggers.consumer.WorkflowDefinitionsService.get_definition_by_workflow_id",
+        AsyncMock(
+            return_value=SimpleNamespace(
+                workflow=SimpleNamespace(version=1),
+                version=1,
+                content={
+                    "title": "invalid",
+                    "description": "",
+                    "entrypoint": {"ref": "start"},
+                    "actions": [],
+                },
+                registry_lock=None,
+            )
+        ),
+    )
+    monkeypatch.setattr(
+        "tracecat.cases.triggers.consumer.WorkflowExecutionsService.connect",
+        workflow_service_connect,
+    )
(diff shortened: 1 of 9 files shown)


jordan-umusu (approved):
lgtm

PR #2892, opened 2026-06-22
feat(api): sync rbac assignments with memberships

jordan-umusu:
## Summary by cubic
Syncs workspace memberships with workspace-scoped RBAC so users are auto-added/removed via direct or group role paths. Adds `via_group` and `via_direct` flags in the API/UI, keeps “Change role” editable for mixed-source members, and reconciles under workspace-scoped RLS with transaction-scoped advisory locks (ENG-1499).

- New Features
  - Added reconciliation primitives and wiring: `workspace_scoped_path_exists`, `group_scoped_path_exists`, `MembershipService.reconcile_workspace_membership`, `MembershipService.reconcile_group_members`, `pg_advisory_xact_lock`, and `workspace_rls_context`. Reconciliation runs on RBAC writes (user/group assignment create/delete, add/remove group member, group delete) and invitation acceptance, with set-based, locked fanout for groups.
  - Members API now returns `via_group` and `via_direct`, and picks role as direct > group > default. UI shows a “Via group” badge, gates Change role by `via_direct` and Remove by `via_group`, with tooltips.

- Bug Fixes
  - Allow changing role for mixed-source members (direct + group). Removing group-derived members is blocked with 409 until group access changes.
  - Re-adding a member is idempotent and preserves any existing non-default direct role; org-wide roles do not create membership.
  - Deleting groups or group assignments, and adding/removing group members, reconciles affected users under workspace RLS to avoid stale memberships.

<sup>Written for commit fb3c961f6b66e1306b0256875c3edf1195bc98a8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/components/workspaces/workspace-members-table.tsx
+++ b/frontend/src/components/workspaces/workspace-members-table.tsx
@@ -21,6 +21,7 @@ import {
   AlertDialogTitle,
   AlertDialogTrigger,
 } from "@/components/ui/alert-dialog"
+import { Badge } from "@/components/ui/badge"
 import { Button } from "@/components/ui/button"
 import {
   Dialog,
@@ -44,6 +45,11 @@ import {
   SelectTrigger,
   SelectValue,
 } from "@/components/ui/select"
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from "@/components/ui/tooltip"
 import { toast } from "@/components/ui/use-toast"
 import {
   useWorkspaceMembers,
@@ -218,8 +224,21 @@ export function WorkspaceMembersTable({
                 />
               ),
               cell: ({ row }) => (
-                <div className="text-xs capitalize">
+                <div className="flex items-center gap-2 text-xs capitalize">
                   {row.getValue<string>("role_name")}
+                  {row.original.via_group && (
+                    <Tooltip>
+                      <TooltipTrigger asChild>
+                        <Badge variant="secondary" className="font-normal">
+                          Via group
+                        </Badge>
+                      </TooltipTrigger>
+                      <TooltipContent>
+                        Access granted through a group. Manage this role from
+                        the group's settings.
+                      </TooltipContent>
+                    </Tooltip>
+                  )}
                 </div>
               ),
               enableSorting: true,
@@ -230,6 +249,9 @@ export function WorkspaceMembersTable({
               id: "actions",
               enableHiding: false,
               cell: ({ row }) => {
+                // Managed via the group, and the backend rejects these edits —
+                // disable the row actions rather than offer a guaranteed failure.
+                const viaGroup = row.original.via_group ?? false
                 return (
                   <DropdownMenu>
                     <DropdownMenuTrigger asChild>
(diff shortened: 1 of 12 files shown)


tracecat/authz/service.py
```diff
@@ -51,26 +102,59 @@ async def list_memberships(self, workspace_id: WorkspaceID) -> Sequence[Membersh
...
+                    UserRoleAssignment.user_id == User.id,
                     UserRoleAssignment.workspace_id == Membership.workspace_id,
                     UserRoleAssignment.organization_id == Workspace.organization_id,
                 ),
             )
             .outerjoin(DBRole, DBRole.id == UserRoleAssignment.role_id)
+            .outerjoin(group_role, group_role.c.user_id == User.id)
             .where(Membership.workspace_id == workspace_id)
         )
         rows = (await self.session.execute(statement)).all()
```
daryllimyt: type the tuple destructuring in the loop below
```suggestion
        rows = (await self.session.execute(statement)).tuples().all()
```

tracecat/authz/service.py
```diff
@@ -156,49 +241,152 @@ async def create_membership(
...
         )
         await self.session.commit()
+        await self.reconcile_workspace_membership(user_id, workspace_id)
+
+    async def reconcile_workspace_membership(
+        self, user_id: UserID, workspace_id: WorkspaceID
+    ) -> None:
+        """Upsert/delete Membership(user, W) to match workspace-scoped RBAC.
+
+        The row must exist iff the user holds at least one ws-scoped role path
```
daryllimyt, line 310: From my local Codex review: this patch establishes Membership as the expected materialized state for workspace-scoped direct/group RBAC paths, but only reconciliation handlers create those rows. On upgrade, existing user_role_assignment.workspace_id and group_role_assignment.workspace_id rows will not have Membership rows, so those users can be absent from member lists and fail workspace membership checks until each assignment or group is touched. Please add an Alembic migration/backfill for both existing direct and group paths.

frontend/src/components/workspaces/workspace-members-table.tsx
```diff
@@ -247,32 +269,78 @@ export function WorkspaceMembersTable({
...
-                              setSelectedUser(row.original)
-                              setIsChangeRoleOpen(true)
-                            }}
-                          >
-                            Change role
-                          </DropdownMenuItem>
-                        </DialogTrigger>
-                      )}
+                      {canUpdateMembers &&
+                        (viaGroup ? (
```
daryllimyt: From my local Codex review: when a member has both a direct workspace assignment and group-derived access, the API can return via_group=true while still returning the direct role_name because direct roles win. This disables Change role solely on via_group, even though handleChangeRole updates the direct assignment and the backend only rejects removal. Admins should still be able to edit the direct role for mixed-source members.

daryllimyt (changes requested):
lgtm, few comments

frontend/src/components/workspaces/workspace-members-table.tsx
jordan-umusu: addressed in the latest commit, added a flag for these mixed-source members and handled accordingly.

tracecat/authz/service.py
jordan-umusu: dont think this is necessary. in the case where a user has a role assignment but not a membership, this state is obvious and easily recoverable. this pr makes this previous immersion bug impossible and heals on write. tl:dr the migration is unnecessary because this state is not valid for actual users on the platform.

PR #2897, opened 2026-06-23
fix(agents): stream replay hardening

jordan-umusu:
## Summary by cubic
Hardens agent stream resume/replay with per‑turn Redis keys, run tagging, composite SSE ids + cursor parsing, and turn‑aware reconnects to stop gaps and duplicate assistant bubbles.

- **New Features**
  - Per‑turn streaming: added session `active_stream_id` and history `curr_run_id`; switched to per‑turn Redis keys (key now includes `stream_id`); plumbed through executor/loopback, `mcp`, and the adapter; new `finalize_turn_activity` guarded by `FINALIZE_TURN_PATCH` compare‑and‑clears `curr_run_id`/`active_stream_id` at terminal.
  - SSE replay: `vercel` adapter emits `id: <redis-id>:<frame-index>` and parses cursors; replays from the composite cursor and self‑heals resumes by synthesizing missing `TEXT_START` or lazily opening parts on orphan deltas (repair frames don’t advance the composite index).
  - Stable UI: router uses `session_id:curr_run_id` for a stable assistant bubble; `ChatResponse` returns `curr_run_id` so the SDK can upsert across reconnects.

- **Bug Fixes**
  - Mid‑turn visibility: persisted history (incl. tool results) is tagged with `curr_run_id`, and mid‑turn loads hide the active run’s partial rows.
  - Reconnects: derive a `TurnLifecycle` to resume, 204, or emit a terminal error; the frontend stops sending Last‑Event‑ID; the server replays the active turn from Redis; `mcp` joins the per‑turn stream and starts at `0-0`; EE error frames now emit to the active per‑turn stream.
  - Key isolation: per‑turn Redis keys prevent cross‑turn cursor collisions; client cursors are parsed and cross‑checked; `mcp` mints/shares the `stream_id` for preset runs so producers/consumers join the same stream.

<sup>Written for commit 92f86c02d9fc18f3be3cfde8f67782426c08c281. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2897?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tracecat/agent/session/router.py
+++ b/tracecat/agent/session/router.py
@@ -25,7 +25,7 @@ from tracecat.agent.session.schemas import (
     AgentSessionUpdate,
 )
 from tracecat.agent.session.service import AgentSessionService
-from tracecat.agent.session.types import AgentSessionEntity
+from tracecat.agent.session.types import AgentSessionEntity, TurnLifecycle
 from tracecat.agent.stream.artifacts import artifact_stream_event
 from tracecat.agent.stream.connector import AgentStream
 from tracecat.agent.stream.events import StreamFormat
@@ -46,6 +46,16 @@ from tracecat.exceptions import EntitlementRequired, TracecatNotFoundError
 from tracecat.logger import logger
 
 router = APIRouter(prefix="/agent/sessions", tags=["agent-sessions"])
+
+
+def _bubble_id(session_id: uuid.UUID, curr_run_id: uuid.UUID | None) -> str | None:
+    """Stable assistant-bubble id for a turn, if the turn is known.
+
+    ``session_id:curr_run_id`` is stable for the whole run, so the AI SDK upserts
+    the live assistant in place across reconnects instead of spawning a duplicate
+    bubble.
+    """
+    return f"{session_id}:{curr_run_id}" if curr_run_id else None
 
 
 async def _require_workspace_chat_entitlement_for_session_tree(
@@ -438,7 +448,7 @@ async def send_message(
                 detail="Workspace access required",
             )
 
-        stream = await AgentStream.new(session_id, workspace_id)
+        message_id: str | None = None
         async with AgentSessionService.with_session(role=role) as svc:
             # Check if this is a legacy chat (read-only)
             if await svc.is_legacy_session(session_id):
@@ -459,16 +469,19 @@ async def send_message(
             )
 
             if isinstance(request, ContinueRunRequest):
-                # Continuations should follow only newly appended events. Resuming
-                # from the persisted DB cursor can replay the approval request that
-                # the active client already rendered before clicking approve/deny.
+                # Continuations resume the same workflow + per-turn stream. Reuse
+                # the existing stream id and follow only newly appended events;
+                # resuming from "0-0" would replay the approval request the active
+                # client already rendered before clicking approve/deny.
+                stream_id = agent_session.active_stream_id
+                stream = await AgentStream.new(session_id, workspace_id, stream_id)
                 start_id = "$"
             else:
-                # Each fresh execution turn gets a new Redis stream buffer so
-                # stale events from the prior turn are never replayed.
-                await stream.reset_for_new_turn()
-                # Read from the beginning of the freshly cleared stream so we still
-                # pick up events emitted before the SSE response starts consuming.
+                # Mint the per-turn stream id at the HTTP layer (turn start) so the
+                # seed artifact and the worker producer both write to the same
+                # fresh per-turn key. No reset/reuse of a prior turn's buffer.
+                stream_id = uuid.uuid4()
+                stream = await AgentStream.new(session_id, workspace_id, stream_id)
                 start_id = "0-0"
                 if await svc.should_seed_initial_artifact(agent_session) and (
                     artifact := await svc.build_initial_artifact(agent_session)
@@ -481,14 +494,23 @@ async def send_message(
 
             # Run session turn (spawns DurableAgentWorkflow)
             try:
-                await svc.run_turn(
+                turn_response = await svc.run_turn(
                     session_id=session_id,
                     request=request,
+                    active_stream_id=stream_id,
                 )
             except Exception:
                 if not isinstance(request, ContinueRunRequest):
+                    # Startup failed after we minted the stream: surface a terminal
+                    # frame so reconnecting clients don't hang, and clear pointers.
+                    # Non-continue turns always mint a fresh per-turn stream id.
+                    assert stream_id is not None
                     try:
-                        await stream.abort_new_turn()
+                        await stream.error("Failed to start agent turn")
+                        await stream.done()
+                        await svc.clear_active_turn(
+                            session_id, expected_stream_id=stream_id
+                        )
                     except Exception as rollback_exc:
                         logger.warning(
                             "Failed to clear stream state after turn startup failure",
@@ -496,6 +518,17 @@ async def send_message(
                             error=str(rollback_exc),
                         )
                 raise
+
+            # Build a bubble id stable for this turn. Prefer the run id returned by
+            # run_turn (new turns) — terminal cleanup may already have nulled the
+            # session row on a fast turn. Continuations return None and reuse the
+            # in-progress run id still pinned on the session row.
+            if turn_response is not None and turn_response.curr_run_id is not None:
+                run_id = turn_response.curr_run_id
+            else:
+                refreshed = await svc.get_session(session_id)
+                run_id = refreshed.curr_run_id if refreshed else None
+            message_id = _bubble_id(session_id, run_id)
 
         logger.info(
             "Starting Vercel streaming session",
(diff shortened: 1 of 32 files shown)


tracecat/agent/stream/events.py
```diff
@@ -13,6 +13,24 @@
 AgentStreamEventTA: TypeAdapter[AgentStreamEvent] = TypeAdapter(AgentStreamEvent)
 
 
+@dataclass(frozen=True, slots=True)
+class VercelFrameCursor:
```
daryllimyt: Vercel related constructs/fns should remain in `vercel` module

tracecat/agent/session/router.py
```diff
@@ -481,14 +494,23 @@ async def send_message(
...
                     session_id=session_id,
                     request=request,
+                    active_stream_id=stream_id,
                 )
             except Exception:
                 if not isinstance(request, ContinueRunRequest):
+                    # Startup failed after we minted the stream: surface a terminal
+                    # frame so reconnecting clients don't hang, and clear pointers.
+                    # Non-continue turns always mint a fresh per-turn stream id.
+                    assert stream_id is not None
```
daryllimyt, line 515: nit: raise with a clearer error message

tracecat/agent/session/service.py
```diff
@@ -1320,9 +1393,58 @@ async def run_turn(
...
+        # the post-run refresh on a fast turn.
         stream_url = f"/api/agent/sessions/{session_id}/stream"
-        return ChatResponse(stream_url=stream_url, chat_id=session_id)
+        return ChatResponse(
+            stream_url=stream_url, chat_id=session_id, curr_run_id=run_id
+        )
+
+    async def get_turn_lifecycle(
+        self, agent_session: AgentSession
+    ) -> tuple[TurnLifecycle, uuid.UUID | None]:
```
daryllimyt: **Typing nit (non-blocking):** consider returning a `NamedTuple` here instead of a bare `tuple[TurnLifecycle, uuid.UUID | None]`.

The two elements are coupled — `run_id` is `None` exactly when the lifecycle is `NONE` — but that contract isn't visible at the call site, which reads as an anonymous positional unpack:

```python
lifecycle, curr_run_id = await svc.get_turn_lifecycle(agent_session)
```

A named result documents the fields and keeps positional unpacking working unchanged:

```python
class TurnLifecycleResult(NamedTuple):
    lifecycle: TurnLifecycle
    run_id: uuid.UUID | None
```

Readability only — no behavior change.

tracecat/agent/types.py
```diff
@@ -39,10 +39,12 @@ def __new__(
         cls,
         workspace_id: uuid.UUID | str,
         session_id: uuid.UUID | str,
+        stream_id: uuid.UUID | str | None = None,
```
daryllimyt: **Typing nit (non-blocking):** the `| str` on `stream_id` is dead width. Every caller passes a `uuid.UUID` (or `None`), and `active_stream_id` is `uuid.UUID | None` end to end, so this is never a bare string — the `str(stream_id)` coercion in the body just makes the `str` arm reachable-but-unused. Could narrow to:

```python
stream_id: uuid.UUID | None = None,
```

Note `workspace_id` and `session_id` above carry the same pre-existing `| str`, so for a consistent signature you'd narrow all three (or leave them as-is). No behavior change either way.

tracecat/chat/schemas.py
```diff
@@ -76,6 +76,14 @@ class ChatResponse(BaseModel):
 
     stream_url: str = Field(..., description="URL to connect for SSE streaming")
     chat_id: uuid.UUID = Field(..., description="Unique chat identifier")
+    curr_run_id: uuid.UUID | None = Field(
+        default=None,
+        description=(
+            "Run id of the turn just spawned. Returned so callers can build the "
```
daryllimyt: clarity: is this the Temporal run_id?

tracecat/agent/session/service.py
```diff
@@ -980,14 +1033,19 @@ async def claim_external_channel_approval_sink(
...
+        active_stream_id: uuid.UUID | None,
+    ) -> None:
         """Emit an end marker so duplicate/no-op continuations don't hang SSE clients."""
         if self.workspace_id is None:
             return
         try:
             redis_client = await get_redis_client()
             await redis_client.xadd(
-                str(StreamKey(self.workspace_id, session_id)),
+                str(StreamKey(self.workspace_id, session_id, active_stream_id)),
```
daryllimyt: nit: str() cast here seems redundant because StreamKey inherits from `str`

tracecat/agent/stream/connector.py
```diff
@@ -39,16 +52,23 @@ def __init__(
...
         self.client = client
         self.workspace_id = workspace_id
         self.session_id = session_id
-        self._stream_key = StreamKey(workspace_id, session_id)
+        self.stream_id = stream_id
+        self._stream_key = StreamKey(workspace_id, session_id, stream_id)
 
     @classmethod
-    async def new(cls, session_id: uuid.UUID, workspace_id: uuid.UUID) -> AgentStream:
+    async def new(
```
daryllimyt, line 67: **Typing nit (non-blocking):** `__init__` orders these as `(workspace_id, session_id, ...)` while `new` orders them `(session_id, workspace_id, ...)` and re-swaps when delegating to `cls(...)` below. All three ids are bare `uuid.UUID`, so a caller (or a future edit) that transposes them is invisible to the type checker.

Cheapest guard: make `new`/`__init__` (and `StreamKey`) **keyword-only** (`*`), which removes the positional-transposition risk at zero cost.

If you want a stronger, refactor-proof guarantee across the HTTP→workflow→producer id-threading, distinct `NewType`s would make them non-interchangeable:

```python
SessionId = NewType("SessionId", uuid.UUID)
RunId     = NewType("RunId", uuid.UUID)
StreamId  = NewType("StreamId", uuid.UUID)
```

Tradeoff: zero runtime cost and Pydantic v2 validates them fine, but `Mapped[...]` in `db/models.py` needs `type_annotation_map` care — so it only pays off adopted across the whole id-set (with a single wrap at the repo edge), not piecemeal.

tracecat/agent/session/service.py
```diff
@@ -1320,9 +1393,58 @@ async def run_turn(
...
+            )
+            return TurnLifecycle.FAILED, curr_run_id
+
+        match description.status:
+            case WorkflowExecutionStatus.RUNNING:
+                return TurnLifecycle.RUNNING, curr_run_id
+            case WorkflowExecutionStatus.COMPLETED:
+                return TurnLifecycle.COMPLETED, curr_run_id
+            case _:
+                # FAILED | TERMINATED | CANCELED | TIMED_OUT | CONTINUED_AS_NEW
```
daryllimyt: should continued-as-new count as a failure? our Turns don't usually hit that limit but if we do, is this the least surprising behavior?

tracecat/mcp/server.py
```diff
@@ -9608,21 +9609,27 @@ async def run_agent_preset(
...
             )
+            # Mint the stream id in the caller so the producer (run_turn pins
+            # this id) and the consumer (below) share the same suffixed key.
+            stream_id = uuid.uuid4()
             # BasicChatRequest is handled at runtime by run_turn's match statement
             # but not included in the ChatRequest type alias
             await svc.run_turn(
-                session.id, cast(ChatRequest, BasicChatRequest(message=prompt))
+                session.id,
+                cast(ChatRequest, BasicChatRequest(message=prompt)),
```
daryllimyt: nit: can we drop the cast() here

daryllimyt (approved):
lgtm, nothing blocking but few nits

tracecat/agent/session/service.py
jordan-umusu: changed this! the semantics of continued-as-new seem least surprising grouped with the running status.

tracecat/chat/schemas.py
jordan-umusu: clarified in the docstring, its the temporal workflow id (agent/{id}), a stable handle compared to the temporal run id which is autogenerated

tracecat/agent/session/service.py
```diff
@@ -771,6 +771,59 @@ async def update_last_stream_id(
...
+        NONE -> 204.
+        """
+        stmt = (
+            update(AgentSession)
+            .where(
+                AgentSession.id == session_id,
+                AgentSession.workspace_id == self.workspace_id,
+                AgentSession.curr_run_id == run_id,
+            )
+            .values(curr_run_id=None, active_stream_id=None)
```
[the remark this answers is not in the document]
jordan-umusu: done in a followup PR

tracecat/agent/stream/events.py
jordan-umusu: moved. couldn't move it directly into adapter/vercel.py though, since stream/connector.py (the generic, protocol-agnostic Redis stream layer) also needs to parse this cursor for resume and adapter/vercel.py already imports from stream/, so reversing that would risk a circular import.

PR #2917, opened 2026-06-29
feat(worker): expose Temporal worker concurrency tuning env knobs

daryllimyt:
## Summary

The DSL worker (`tracecat/dsl/worker.py`) constructed its Temporal `Worker`
with SDK-default concurrency and an activity thread pool that was not surfaced
in any deployment target, so there was no first-class way to tune worker
concurrency per deployment.

On small CPU allocations, high worker concurrency (100 workflow-task slots +
100 activity slots + a 100-thread activity pool, all sharing one CPython GIL)
can starve the synchronous `DSLScheduler` construction enough to trip
Temporal's 2-second workflow deadlock detector (`[TMPRL1101]`).

This PR exposes three env-configurable knobs so operators can right-size
concurrency to the worker's CPU allocation. **Defaults are unchanged (100),
so there is no behavior change** — this only adds the ability to tune.

## Knobs

| Env var | Worker setting | Default |
| --- | --- | --- |
| `TEMPORAL__THREADPOOL_MAX_WORKERS` | activity `ThreadPoolExecutor` size | 100 |
| `TEMPORAL__MAX_CONCURRENT_ACTIVITIES` | `max_concurrent_activities` | 100 |
| `TEMPORAL__MAX_CONCURRENT_WORKFLOW_TASKS` | `max_concurrent_workflow_tasks` | 100 |

## Changes

- `tracecat/dsl/worker.py`: read the two new env vars and pass
  `max_concurrent_activities` / `max_concurrent_workflow_tasks` to
  `Worker(...)` (mirrors the executor and agent workers).
- `docker-compose.yml`, `docker-compose.dev.yml`, `docker-compose.local.yml`:
  surface the three knobs on the `worker` service.
- `deployments/fargate/`: add `worker_threadpool_max_workers`,
  `worker_max_concurrent_activities`, `worker_max_concurrent_workflow_tasks`
  (root + ECS module variables, `main.tf` passthrough, `worker_env`).
- `.env.example`: document the knobs.

## Tuning guidance

On a 2 vCPU worker the activity thread pool is the highest-leverage lever: the
worker's sync activities are CPU-bound expression/template evaluation, which
cannot run in parallel under a single GIL, so lowering the pool removes
contention without reducing real throughput. Keep `max_concurrent_activities`
generous — it is mostly consumed by async I/O activities that release the GIL.

## Follow-up

Helm and EKS manifests live in the `deployments/k8s` submodule
(`TracecatHQ/k8s`) and will get the same wiring in that repo.

[...]

PR #2928, opened 2026-06-30
perf(dsl): build loop regions in a single pass

daryllimyt:
## Summary

- `DSLScheduler._build_loop_regions` rescanned **every** action and walked the scope hierarchy for **each** `core.loop.end`, making it O(loops × actions × depth). `DSLScheduler.__init__` is a fully synchronous constructor (no `await`) that Temporal re-runs in full on every workflow replay (cache eviction, worker failover, reset), so this cost repeated on every activation for loop-heavy DAGs.
- Rewrite it as two passes: validate each `loop.end` (unchanged checks/error messages), then do a single O(actions × depth) sweep that walks each action's scope chain once, claiming membership for every loop scope encountered along the way.
- `_scope_is_descendant` is now unused (it was only called from the old membership comprehension) and is removed.
- Output is unchanged: same `LoopRegion` membership semantics, including the `end_ref` union (the loop-end action's own scope is its parent, not a descendant, so it never appears via the walk and still needs the explicit union).

This addresses the algorithmic half of a previously investigated `[TMPRL1101]` deadlock-detector trip (synchronous scheduler construction not yielding within Temporal's 2s budget); the worker-concurrency mitigation for the same incident shipped separately in #2917. A follow-up "don't rebuild the static plan on every replay" change is intentionally out of scope for this PR — it needs a workflow-definition-version cache key that isn't currently plumbed into the scheduler.

## Test plan

- [x] `uv run ruff check` / `ruff format --check` on the changed file
- [x] `uv run basedpyright` — 0 errors
- [x] `tests/unit/test_dsl_scope_validation.py` — 26 passed
- [x] `tests/temporal/test_loop_control_flow.py` — 10 passed (nested loops, loop-inside-scatter, skip propagation)
- [x] `tests/temporal` (full suite, `-n auto`) — 168 passed, 16 skipped, 5 failed; the 5 failures (`requires_api`-marked table/secret tests) reproduce identically on unmodified `main` in this environment, confirming they're pre-existing and unrelated to this change

---
## Summary by cubic
Build loop regions in a single pass to reduce scheduler construction from O(loops × actions × depth) to O(actions × depth). This cuts the synchronous `DSLScheduler.__init__` cost on Temporal replays and addresses the algorithmic side of `[TMPRL1101]`.

- **Refactors**
  - Rewrote `DSLScheduler._build_loop_regions` into two phases: validate `loop.end`s, then a single sweep that assigns membership while walking each action’s scope chain once.
  - Preserved behavior: same `LoopRegion` membership semantics, including explicit `end_ref` union.
  - Removed unused `_scope_is_descendant`.

<sup>Written for commit 5308aab0558e996b2cc271adea21358b2491d799. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2928?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tracecat/dsl/scheduler.py
+++ b/tracecat/dsl/scheduler.py
@@ -225,16 +225,9 @@ class DSLScheduler:
         dst_ref, edge_type = adj
         return dst_ref, edge_type.value
 
-    def _scope_is_descendant(self, scope: str, ancestor_scope: str) -> bool:
-        curr_scope = scope
-        while curr_scope is not None:
-            if curr_scope == ancestor_scope:
-                return True
-            curr_scope = self.scope_hierarchy.get(curr_scope)
-        return False
-
     def _build_loop_regions(self) -> dict[str, LoopRegion]:
-        regions: dict[str, LoopRegion] = {}
+        # Phase 1: validate each loop.end and map loop scope -> end ref.
+        loop_end_by_scope: dict[str, str] = {}
         for stmt in self.dsl.actions:
             if stmt.action != PlatformAction.LOOP_END:
                 continue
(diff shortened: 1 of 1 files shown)


daryllimyt:
On the deferred "don't rebuild the static plan on every replay" follow-up mentioned above — is process-wide caching of that plan an antipattern here? Short answer: no, but the risk is concentrated in one place worth flagging for whoever picks it up.

Temporal's own SDK already does this for a superset of this state: the sticky workflow cache (`max_cached_workflows`, default 1000) holds full execution state process-wide, keyed by run ID, for the exact same reason (avoid replay cost). A secondary cache for just the static/derived plan (tasks/adj/scopes/loop_regions), keyed by definition identity, is a natural complement, not a foreign pattern — and since the cached value would be a pure, immutable function of the DSL, the usual "shared mutable state" objections mostly don't apply: a race just recomputes it twice (wasted CPU, not corruption), not a torn read.

It becomes an antipattern under three specific failure modes:

1. **Wrong cache key → silent determinism violation.** If the key doesn't precisely capture DSL content (e.g., `wf_id` alone, without a version), two different DSL edits under the same key would serve stale scheduling logic — and because Temporal workflows are expected to be deterministic, this kind of bug wouldn't throw, it'd just silently misschedule. This is exactly why Layer B is deferred: `DSLScheduler` only receives `dsl` + `run_context`, and `RunContext` has no version field today, so there's no cheap, correct key available yet without new plumbing (the definition version is already resolved elsewhere via `get_workflow_definition_activity` / `resolve_registry_lock_activity` — threading it into the scheduler is the actual work).
2. **Unbounded growth.** Workers are long-running processes; a plain dict keyed by every definition version ever seen leaks memory over the worker's lifetime. Needs an LRU bound, not a dict.
3. **Caching something that isn't actually pure.** Easy to accidentally capture a reference to per-execution state (role, context) in the "static" plan during refactor — would cause cross-execution interference. Worth a frozen/slotted dataclass plus a type-level boundary to keep it honest.

So: sound pattern, the real risk is entirely in getting #1 right — that's design work, not a smell to avoid.

jordan-umusu (approved):
lgtm

daryllimyt:
@cubic review

PR #2930, opened 2026-07-01
feat(agents): chat interrupts

jordan-umusu:
## Summary by cubic
Adds a Stop control to interrupt an active agent turn with fast, graceful cancellation across workflow, runtime, and UI. Cancels use a low-latency Redis signal, preserve runtime errors, render an “Interrupted” divider, and mark affected tool calls as interrupted without surfacing abort noise.

- **New Features**
  - API: `POST /workspaces/{workspace_id}/agent/sessions/{session_id}/cancel` accepts `{ reason: "user_cancel" }`, returns `{ session_id, run_id, reason }`, responds `409` if no turn is running; enforces workspace chat entitlement.
  - Cancellation flow: Workflow `request_cancel` works mid-turn, during approvals, and before tool execution; writes a Redis signal keyed by `run_id` that the executor polls; emits a `data-cancelled` stream event with `tool_call_ids`; persists a `CANCELLED` message pinned to the cancelled `run_id` for reloads.
  - UI: Stop button added; submit is disabled while generating/cancelling; shows a centered “Interrupted” divider; aborted tool calls display “Interrupted” via a new `output-interrupted` state. Added `useCancelChatTurn` and support for `data-cancelled` parts in live streams and saved messages.

- **Bug Fixes**
  - Executor/Broker/Runtime: Eagerly queues `interrupt()` before client connect or query send, with a 30s hard-cancel fallback; tracks active turns with runtime handles; reports success only when the runtime had no error; derives `interrupted_tool_call_ids` from cancel-then-error ordering.
  - Router/Workflow/Streaming: Treats Temporal `CANCELLED` as terminal and emits a final frame; centralized cancelled-turn output helper ensures consistent final notices and history; worker caps heartbeat throttle at 5s so SDK-driven cancels arrive faster.

<sup>Written for commit 889627398faa7be388aa14d91153a10b8eaf1870. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2930?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -27,5 +27,8 @@
         "*.github.com"
       ]
     }
+  },
+  "enabledPlugins": {
+    "codex@openai-codex": true
   }
 }
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -27,8 +27,5 @@
         "*.github.com"
       ]
     }
-  },
-  "enabledPlugins": {
-    "codex@openai-codex": true
   }
 }
--- a/packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
+++ b/packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
 import uuid
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
@@ -1052,34 +1051,24 @@ class DurableAgentWorkflow:
             heartbeat_timeout=timedelta(seconds=60),
             retry_policy=RETRY_POLICIES["activity:fail_fast"],
         )
-        cancel_wait_task = asyncio.create_task(
-            workflow.wait_condition(lambda: self._cancel_requested)
+        # ActivityHandle is an asyncio.Task subclass, so .done() is valid.
+        # Neither wait_condition nor the handle poll emits history commands,
+        # so this race stays replay-safe.
+        await workflow.wait_condition(
+            lambda: activity_handle.done() or self._cancel_requested
         )
-        try:
-            done, _pending = await asyncio.wait(
-                [activity_handle, cancel_wait_task],
-                return_when=asyncio.FIRST_COMPLETED,
-            )
-
-            if activity_handle in done:
-                return await activity_handle
-
+        if not activity_handle.done():
             activity_handle.cancel()
-            try:
-                return await activity_handle
-            except ActivityError as e:
-                if self._cancel_requested and isinstance(
-                    e.cause, TemporalCancelledError
-                ):
-                    return AgentExecutorResult(
-                        success=True,
-                        cancelled=True,
-                        cancelled_reason=self._cancel_reason or "user_cancel",
-                    )
-                raise
-        finally:
-            if not cancel_wait_task.done():
-                cancel_wait_task.cancel()
+        try:
+            return await activity_handle
+        except ActivityError as e:
+            if self._cancel_requested and isinstance(e.cause, TemporalCancelledError):
+                return AgentExecutorResult(
+                    success=True,
+                    cancelled=True,
+                    cancelled_reason=self._cancel_reason or "user_cancel",
+                )
+            raise
 
     async def _run_with_agent_executor(
         self, args: AgentWorkflowArgs, cfg: AgentConfig
@@ -1241,13 +1230,9 @@ class DurableAgentWorkflow:
                     session_id=self.session_id,
                     reason=result.cancelled_reason,
                 )
-                message_history = await self._load_terminal_message_history(result)
-                return AgentOutput(
-                    output=None,
-                    message_history=message_history,
-                    duration=(datetime.now(UTC) - info.start_time).total_seconds(),
-                    usage=RunUsage(requests=0, input_tokens=0, output_tokens=0),
-                    session_id=self.session_id,
+                # Executor loopback already emitted the cancelled notice.
+                return await self._cancelled_turn_output(
+                    result, info, emit_cancelled=False
                 )
 
             if not result.success:
@@ -1305,14 +1290,8 @@ class DurableAgentWorkflow:
                         }
                     )
                     await self.approvals.handle_decisions()
-                    await self._emit_session_cancelled()
-                    message_history = await self._load_terminal_message_history(result)
-                    return AgentOutput(
-                        output=None,
-                        message_history=message_history,
-                        duration=(datetime.now(UTC) - info.start_time).total_seconds(),
-                        usage=RunUsage(requests=0, input_tokens=0, output_tokens=0),
-                        session_id=self.session_id,
+                    return await self._cancelled_turn_output(
+                        result, info, emit_cancelled=True
                     )
                 # Persist approval decisions to DB (atomic with chat messages)
                 await self.approvals.handle_decisions()
@@ -1322,14 +1301,8 @@ class DurableAgentWorkflow:
                         session_id=self.session_id,
                         reason=self._cancel_reason,
                     )
-                    await self._emit_session_cancelled()
-                    message_history = await self._load_terminal_message_history(result)
-                    return AgentOutput(
-                        output=None,
-                        message_history=message_history,
-                        duration=(datetime.now(UTC) - info.start_time).total_seconds(),
-                        usage=RunUsage(requests=0, input_tokens=0, output_tokens=0),
-                        session_id=self.session_id,
+                    return await self._cancelled_turn_output(
+                        result, info, emit_cancelled=True
                     )
 
                 # Execute approved tools and reconcile the SDK transcript.
@@ -1343,14 +1316,8 @@ class DurableAgentWorkflow:
                         session_id=self.session_id,
                         reason=self._cancel_reason,
                     )
-                    await self._emit_session_cancelled()
-                    message_history = await self._load_terminal_message_history(result)
-                    return AgentOutput(
-                        output=None,
-                        message_history=message_history,
-                        duration=(datetime.now(UTC) - info.start_time).total_seconds(),
-                        usage=RunUsage(requests=0, input_tokens=0, output_tokens=0),
-                        session_id=self.session_id,
+                    return await self._cancelled_turn_output(
+                        result, info, emit_cancelled=True
                     )
 
                 tool_results: list[ToolExecutionResult] = []
@@ -1416,6 +1383,31 @@ class DurableAgentWorkflow:
                 ),
                 session_id=self.session_id,
             )
+
+    async def _cancelled_turn_output(
+        self,
+        result: AgentExecutorResult,
+        info: workflow.Info,
+        *,
+        emit_cancelled: bool,
+    ) -> AgentOutput:
+        """Build the terminal output for a cancelled turn.
+
+        The executor-cancel loopback already emitted the cancelled notice, so
+        that path passes emit_cancelled=False. Approval-wait cancels have not
+        emitted yet and must do so before loading history to preserve the
+        activity command order at every call site.
+        """
+        if emit_cancelled:
+            await self._emit_session_cancelled()
+        message_history = await self._load_terminal_message_history(result)
+        return AgentOutput(
+            output=None,
+            message_history=message_history,
+            duration=(datetime.now(UTC) - info.start_time).total_seconds(),
+            usage=RunUsage(requests=0, input_tokens=0, output_tokens=0),
+            session_id=self.session_id,
+        )
 
     async def _load_terminal_message_history(
         self,
--- a/packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
+++ b/packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import uuid
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
@@ -8,7 +9,13 @@ from typing import Any, Literal
 from pydantic import BaseModel, ConfigDict, Field
 from temporalio import workflow
 from temporalio.common import TypedSearchAttributes
-from temporalio.exceptions import ActivityError, ApplicationError
+from temporalio.exceptions import (
+    ActivityError,
+    ApplicationError,
+)
+from temporalio.exceptions import (
+    CancelledError as TemporalCancelledError,
+)
 
 with workflow.unsafe.imports_passed_through():
     from pydantic_ai.messages import ToolCallPart
@@ -125,6 +132,12 @@ EMIT_PRE_STREAM_SESSION_ERRORS_PATCH = (
 PERSIST_SESSION_ERROR_PATCH = (
     "tracecat_ee.agent.workflows.durable.persist_session_error"
 )
+# Temporal patch IDs are persisted in each workflow execution's history. Use a
+# stable, unique ID for every command-producing workflow change, and never reuse
+# an ID for another change. Keep both branches until old histories that lack the
+# marker have aged out, then use workflow.deprecate_patch(...) before removing
+# the marker entirely in a later cleanup.
+AGENT_REQUEST_CANCEL_PATCH = "durable-agent-request-cancel-v1"
 
 
 @dataclass(frozen=True, slots=True)
@@ -416,6 +429,10 @@ class WorkflowApprovalSubmission(BaseModel):
     decision_metadata: dict[str, dict[str, Any]] | None = None
 
 
+class WorkflowCancelRequest(BaseModel):
+    reason: Literal["user_cancel"] = "user_cancel"
+
+
 def _resolve_agent_output(
     *,
     output: Any,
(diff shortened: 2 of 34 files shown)


tracecat/agent/runtime/claude_code/broker.py
```diff
@@ -176,18 +189,36 @@ async def run_turn_in_session_lease(
...
+        """Gracefully interrupt an active turn's Claude SDK client, if any.
+
+        No-op if the turn hasn't constructed its runtime yet (still hydrating
+        the work dir) - the caller's timeout/hard-cancel fallback covers that
+        narrow window.
+        """
+        async with self._lock:
+            active = self._active_turns.get(session_id)
+        if active is None or active.runtime is None:
+            return
```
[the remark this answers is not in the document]
jordan-umusu: not true, the cancellation happens not primarily at the harness level (which is a graceful attempt) but instead as a temporal activity cancellation on the parent activity task.

packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
```diff
@@ -931,6 +982,68 @@ def validate_set_approvals(self, submission: WorkflowApprovalSubmission) -> None
...
+            activity_handle.cancel()
+        try:
+            return await activity_handle
+        except ActivityError as e:
+            if self._cancel_requested and isinstance(e.cause, TemporalCancelledError):
+                return AgentExecutorResult(
+                    success=True,
+                    cancelled=True,
+                    cancelled_reason=self._cancel_reason or "user_cancel",
+                )
```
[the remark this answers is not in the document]
jordan-umusu: low-sev nearly impossible to reproduce, the cancel can't even be sent until the workflow is RUNNING and describable, and the frontend stop button is gated on that same lifecycle so.

packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
```diff
@@ -1083,16 +1207,22 @@ async def _run_with_agent_executor(
...
-                task_queue=config.TRACECAT__AGENT_EXECUTOR_QUEUE,
-                start_to_close_timeout=timedelta(
-                    seconds=config.TRACECAT__AGENT_SANDBOX_TIMEOUT
-                ),
-                heartbeat_timeout=timedelta(seconds=60),
-                retry_policy=RETRY_POLICIES["activity:fail_fast"],
-            )
+            result = await self._run_agent_activity_turn(executor_input)
+
+            if result.cancelled:
```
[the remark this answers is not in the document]
jordan-umusu: edge-case that is not owrht the squeeze, both a user would have to cnacel and the runtime would have to die in the teardown window for this to be possible, can implement as a follow up if needed.

packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
```diff
@@ -931,6 +982,68 @@ def validate_set_approvals(self, submission: WorkflowApprovalSubmission) -> None
...
+        try:
+            return await activity_handle
+        except ActivityError as e:
+            if self._cancel_requested and isinstance(e.cause, TemporalCancelledError):
+                return AgentExecutorResult(
+                    success=True,
+                    cancelled=True,
+                    cancelled_reason=self._cancel_reason or "user_cancel",
+                )
+            raise
```
daryllimyt: nit: can be inlined

packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
```diff
@@ -897,6 +916,38 @@ async def _emit_session_error(self, message: str) -> None:
                 error=str(emit_error),
             )
 
+    async def _emit_session_cancelled(self) -> None:
+        """Emit the advisory cancelled-turn stream event for approval-wait cancels.
```
daryllimyt: redundant? we have the validator below

tracecat/agent/runtime/claude_code/broker.py
```diff
@@ -62,7 +75,7 @@ async def stop(self) -> None:
         """Cancel any remaining active turns and reject future work."""
         async with self._lock:
             self._closed = True
-            active_tasks = list(self._active_turns.values())
+            active_tasks = [active.task for active in self._active_turns.values()]
         for task in active_tasks:
             task.cancel()
         for task in active_tasks:
```
daryllimyt: Pre-existing (this PR only touches the line above), but since `stop()` is in scope: if one of these tasks is failing with a non-`CancelledError`, `await task` re-raises it here and aborts the loop, leaving the remaining tasks un-awaited — shutdown stops mid-list and the rest surface as "Task exception was never retrieved" warnings. The `gather(..., return_exceptions=True)` in `executor_worker._stop_runtime_services` contains the blast, but `stop()` could be robust on its own:

```suggestion
        await asyncio.gather(*active_tasks, return_exceptions=True)
```
(replacing both the second loop and the try/except)

tracecat/agent/runtime/claude_code/runtime.py
```diff
@@ -872,6 +877,33 @@ async def _handle_approval_request(
...
+    async def interrupt(self, *, reason: str) -> None:
+        """Gracefully stop the in-flight turn via the Claude SDK client.
+
+        Safe to call before the client has connected or before the query has
+        been sent - the request is recorded and applied once both have
+        happened, so a cancel requested early in a turn's lifecycle still
+        takes effect instead of racing the SDK client's own setup.
+        """
+        self._interrupt_requested = reason
+        await self._client_connected_event.wait()
```
daryllimyt, line 895: If `run()` raises before the client connects (transport spawn failure, `_prepare_resume_and_mcp` error, etc.), `_client_connected_event` is never set and this wait never completes — `interrupt()` hangs forever. Today that's only bounded by the caller: the activity wraps `broker.interrupt_turn` in `asyncio.timeout(GRACEFUL_CANCEL_TIMEOUT_SECONDS)` with a hard-cancel fallback. That contract lives entirely in the caller, though — worth stating in this docstring that `interrupt()` may never return if the turn dies before connect and callers must bound it with a timeout (or alternatively give `broker.interrupt_turn` a `timeout` param that falls back to `task.cancel()`).

packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
```diff
@@ -957,6 +1011,68 @@ def validate_set_approvals(self, submission: WorkflowApprovalSubmission) -> None
                     + ", ".join(sorted(unexpected_metadata_ids))
                 )
 
+    @workflow.update
+    def request_cancel(self, request: WorkflowCancelRequest) -> None:
+        request = WorkflowCancelRequest.model_validate(request)
```
daryllimyt: nit: looks redundant given the below validator

daryllimyt:
## Triage: verified review findings on cancellation paths

Traced all three reported findings through the code. All three point at real code facts, but each needs qualification.

### Finding 1 — cancelled marker can attach to the wrong run: **valid race, Medium (not High)**

- `append_cancelled_marker` reads `agent_session.curr_run_id` at write time (`tracecat/agent/session/service.py:865`).
- Ordering is as claimed: the loopback emits `done` from inside the executor activity (`tracecat/agent/executor/loopback.py:797`), while the marker is persisted later by the `emit_session_cancelled` activity, scheduled only after the executor activity resolves.
- The window is open: `validate_turn_request` (`service.py:1543`) only blocks on pending approvals — nothing stops a new turn from committing `curr_run_id = new_run_id` (`service.py:1452`) while the cancelled workflow is still finalizing.

Impact is cosmetic: a mis-tagged marker is hidden by the mid-turn filter (`service.py:2099-2103`) while the *next* turn runs, then revealed when that turn finalizes. Hydration skips markers by `kind`, so no transcript corruption — the "stopped by user" divider is just missing on reload during the next turn, then reappears. Fix is right and trivial: the workflow already derives `curr_run_id` replay-safely from its own workflow id (`durable.py:1093`), so pass it through `EmitSessionCancelledInputs` explicitly. Worth doing.

### Finding 2 — skipped terminal stream frame: **valid in substance, but both stated scenarios are wrong**

- **"Graceful interrupt timed out" doesn't reach the cancelled path.** On timeout, `activity.py:676` re-raises `TimeoutError`, so the activity completes **failed**, not cancelled. The `isinstance(e.cause, TemporalCancelledError)` check at `durable.py:1068` doesn't match; the error propagates to `run()`'s handler (`durable.py:852`) and `_finalize_session_error(should_stream=True)` streams error+done. Wrong UX (error banner instead of "stopped"), but not a hang.
- **Hard-cancel after runtime attach also emits frames.** `runtime.run()` has `finally: await self._event_writer.send_done()` (`runtime.py:1557`), so `broker.cancel_turn()` → `task.cancel()` still drives the loopback done frame.

The gap that *is* real: **the activity is cancelled before it ever starts** — cancel lands while the executor activity is queued (executor workers saturated, exactly when users mash stop). The handle resolves with a clean `CancelledError`, `durable.py:1068` returns `cancelled=True`, and `_cancelled_turn_output(emit_cancelled=False)` persists the marker with no stream frames — no executor ever existed to emit any. The live SSE client hangs until its reconnect logic kicks in (cold reconnect self-heals). Narrower variant: hard-cancel during pre-runtime hydration, where `runtime.run` was never entered.

Caution on the proposed fix: unconditional `emit_cancelled=True` in that branch would double-emit `done` where the runtime's `finally` *did* fire — the code explicitly warns a second done races client teardown (`activities.py:152-157`). A `stream_emitted` flag can't ride on a cancelled activity (no result payload). Cleaner fix: short-circuit when `_cancel_requested` is set while the activity is unstarted and have the workflow emit, or make `done` idempotent stream-side.

### Finding 3 — one-shot interrupt in the Temporal fallback branch: **technically valid, narrow in practice; Low/Medium**

Factually correct: `activity.py:667` ignores `interrupt_turn`'s `False` return; the watcher at `activity.py:737` retries. Mitigating detail the review missed: the watcher task stays alive during the `CancelledError` handler (only cancelled in the `finally` at `activity.py:687`), so whenever the Redis signal was written — every endpoint-initiated cancel — the retry loop is still running concurrently and covers the gap. The one-shot weakness only bites when the best-effort Redis write failed (`service.py:1788`) *and* Temporal delivers the cancel during the pre-attach hydration window. Even then it degrades to the 30s timeout → hard cancel → streamed error, not a hang. Cheap to fix by sharing the retry helper.

### On the design suggestions

Mostly sound, one pushback: the "grace window before `activity_handle.cancel()`" suggestion solves a non-problem — Temporal cancellation lands in the activity's *own* graceful branch (`except CancelledError` → interrupt → 30s window), and the runtime dedupes interrupts via `_interrupt_sent` (`runtime.py:896`), so the eager cancel doesn't destructively race the Redis path. Conversely, "short-circuit before sandbox startup when stop is requested pre-activity" is the strongest item — it's the actual fix for finding 2's reachable case.

**Bottom line:** fix finding 1 (trivial, real), fix finding 2 but for the queued-activity scenario rather than the ones stated, and take finding 3 as cheap hardening. None of the three blocks the branch as-is.

daryllimyt:
## Triage: verified review findings on cancellation paths

Traced all three reported findings through the code. All three point at real code facts, but each needs qualification.

### Finding 1 — cancelled marker can attach to the wrong run: **valid race, Medium (not High)**

- `append_cancelled_marker` reads `agent_session.curr_run_id` at write time (`tracecat/agent/session/service.py:865`).
- Ordering is as claimed: the loopback emits `done` from inside the executor activity (`tracecat/agent/executor/loopback.py:797`), while the marker is persisted later by the `emit_session_cancelled` activity, scheduled only after the executor activity resolves.
- The window is open: `validate_turn_request` (`service.py:1543`) only blocks on pending approvals — nothing stops a new turn from committing `curr_run_id = new_run_id` (`service.py:1452`) while the cancelled workflow is still finalizing.

Impact is cosmetic: a mis-tagged marker is hidden by the mid-turn filter (`service.py:2099-2103`) while the *next* turn runs, then revealed when that turn finalizes. Hydration skips markers by `kind`, so no transcript corruption — the "stopped by user" divider is just missing on reload during the next turn, then reappears. Fix is right and trivial: the workflow already derives `curr_run_id` replay-safely from its own workflow id (`durable.py:1093`), so pass it through `EmitSessionCancelledInputs` explicitly. Worth doing.

### Finding 2 — skipped terminal stream frame: **valid in substance, but both stated scenarios are wrong**

- **"Graceful interrupt timed out" doesn't reach the cancelled path.** On timeout, `activity.py:676` re-raises `TimeoutError`, so the activity completes **failed**, not cancelled. The `isinstance(e.cause, TemporalCancelledError)` check at `durable.py:1068` doesn't match; the error propagates to `run()`'s handler (`durable.py:852`) and `_finalize_session_error(should_stream=True)` streams error+done. Wrong UX (error banner instead of "stopped"), but not a hang.
- **Hard-cancel after runtime attach also emits frames.** `runtime.run()` has `finally: await self._event_writer.send_done()` (`runtime.py:1557`), so `broker.cancel_turn()` → `task.cancel()` still drives the loopback done frame.

The gap that *is* real: **the activity is cancelled before it ever starts** — cancel lands while the executor activity is queued (executor workers saturated, exactly when users mash stop). The handle resolves with a clean `CancelledError`, `durable.py:1068` returns `cancelled=True`, and `_cancelled_turn_output(emit_cancelled=False)` persists the marker with no stream frames — no executor ever existed to emit any. The live SSE client hangs until its reconnect logic kicks in (cold reconnect self-heals). Narrower variant: hard-cancel during pre-runtime hydration, where `runtime.run` was never entered.

Caution on the proposed fix: unconditional `emit_cancelled=True` in that branch would double-emit `done` where the runtime's `finally` *did* fire — the code explicitly warns a second done races client teardown (`activities.py:152-157`). A `stream_emitted` flag can't ride on a cancelled activity (no result payload). Cleaner fix: short-circuit when `_cancel_requested` is set while the activity is unstarted and have the workflow emit, or make `done` idempotent stream-side.

### Finding 3 — one-shot interrupt in the Temporal fallback branch: **technically valid, narrow in practice; Low/Medium**

Factually correct: `activity.py:667` ignores `interrupt_turn`'s `False` return; the watcher at `activity.py:737` retries. Mitigating detail the review missed: the watcher task stays alive during the `CancelledError` handler (only cancelled in the `finally` at `activity.py:687`), so whenever the Redis signal was written — every endpoint-initiated cancel — the retry loop is still running concurrently and covers the gap. The one-shot weakness only bites when the best-effort Redis write failed (`service.py:1788`) *and* Temporal delivers the cancel during the pre-attach hydration window. Even then it degrades to the 30s timeout → hard cancel → streamed error, not a hang. Cheap to fix by sharing the retry helper.

### On the design suggestions

Mostly sound, one pushback: the "grace window before `activity_handle.cancel()`" suggestion solves a non-problem — Temporal cancellation lands in the activity's *own* graceful branch (`except CancelledError` → interrupt → 30s window), and the runtime dedupes interrupts via `_interrupt_sent` (`runtime.py:896`), so the eager cancel doesn't destructively race the Redis path. Conversely, "short-circuit before sandbox startup when stop is requested pre-activity" is the strongest item — it's the actual fix for finding 2's reachable case.

**Bottom line:** fix finding 1 (trivial, real), fix finding 2 but for the queued-activity scenario rather than the ones stated, and take finding 3 as cheap hardening. None of the three blocks the branch as-is.

packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
```diff
@@ -1230,6 +1387,31 @@ async def _run_with_agent_executor(
...
+        yet and pass emit_cancelled=True. Either way the marker is persisted
+        before loading history so the terminal history includes it.
+        """
+        await self._emit_session_cancelled(emit_stream=emit_cancelled)
+        message_history = await self._load_terminal_message_history(result)
+        return AgentOutput(
+            output=None,
+            message_history=message_history,
+            duration=(datetime.now(UTC) - info.start_time).total_seconds(),
+            usage=RunUsage(requests=0, input_tokens=0, output_tokens=0),
```
daryllimyt, line 1470: note: we should eventually calculate this somehow

.claude/settings.json
```diff
@@ -27,5 +27,8 @@
         "*.github.com"
       ]
     }
+  },
```
daryllimyt: maybe keep this in your user settings

daryllimyt (approved):
lgtm, small non-blocking comments. only:
- drop codex plugin from project level

packages/tracecat-ee/tracecat_ee/agent/workflows/durable.py
jordan-umusu: Agreed — the cancel path doesn't carry usage back through `AgentExecutorResult` today, so cancelled turns report zero usage. Will plumb partial-turn usage from the runtime into the cancelled terminal output as a follow-up.

Issue #1214, opened 2025-06-22
[Feature Request] Bring Your Own Key Vault (BYOKV) as Primary Secrets Manager


Zalmez:
**Is your feature request related to a problem? Please describe.**
Organizations, particularly those with mature cloud security postures and compliance requirements, need centralized and highly controlled secret management. Tracecat's current in-app secrets manager, while functional for basic use cases, presents several challenges for these environments such as: 
* Compliance & Trust:
Storing sensitive API keys, credentials and certificates within traacecat's database (even if encrypted) may not meet strict compliance mandates for individual companies
* Infrastructure as code (IaC) & Automation:
In environments heavily reliant on IaC (Terraform, Bicep, etc.) for deploying infrastructure and applications, secrets are typically managed and injected via established cloud key vault services. The current model requires manual secret entry into Tracecat, breaking IaC principles and hindering full automation of Tracecat deployments and configurations.
* Operational overhead:
Managing secrets in disparate systems (Tracecat's internal manager + external key vaults for other applications) increases operational complexity, leads to potential inconsistencies, and complicates auditing and rotation processes.
* Security posture:
Relying solely on Tracecat's internal secret management limits the ability to leverage existing enterprise-grade security controls, such as granular IAM/RBAC, detailed audit logging, and hardware security module (HSM) backing provided by cloud key vaults.

**Describe the solution you'd like**
I suggest implementing the ability to bring your own key vault supporting the major secrets managers out there:
* Azure Key vault
* AWS Secrets Manager
* Google Cloud Secret Manager 

**Describe alternatives you've considered**
Current Tracecat In-App Secrets Manager: This is the existing alternative, but it falls short for enterprises requiring stronger control, IaC integration, and compliance adherence.

**Links to relevant documentation or examples**

* AWS Secrets Manager: https://aws.amazon.com/secrets-manager/
* Azure Key Vault: https://azure.microsoft.com/en-us/products/key-vault/
* Google Cloud Secret Manager: https://cloud.google.com/secret-manager

**Additional context**
Implementing BYOKV as the primary secret management mechanism would significantly enhance Tracecat's appeal to enterprise customers and security-conscious organizations. It aligns Tracecat with industry best practices for secret management, strengthens its security posture, and facilitates seamless integration into existing cloud infrastructure and IaC pipelines.

topher-lo:
[...]

PR #2940, opened 2026-07-03
feat(api): pluggable secrets backend + read-only HashiCorp Vault support

TyTcoH:
Follows up on #1214 — a **draft / proof of concept** for a pluggable secrets
layer with a read-only HashiCorp Vault backend. Built for
isolated k8s cluster deployment and shared as-is: adopt it, rework it, or lift the
parts that are useful. I've enabled "Allow edits by maintainers" so you can
push straight to the branch.

### The idea

Resolve workflow secret values from Vault at execution time instead of storing
them (Fernet-encrypted) in Postgres. For teams with a corporate Vault that
kills the second copy of crown-jewel credentials, the single non-rotating
master key, and the DB/backups as an extra attack surface — secrets stay under
Vault's own ACLs, audit, and rotation.

`TRACECAT__SECRETS_BACKEND=db` stays the default and is untouched: the DB
backend is a behavior-preserving extraction of today's
`search_secrets` + `decrypt_keyvalues` flow.

### How it fits

A `SecretsBackend` protocol (`tracecat/secrets/backend.py`) picked by
`TRACECAT__SECRETS_BACKEND` (`db` | `vault`):

- **`DatabaseSecretsBackend`** — today's logic, extracted.
- **`VaultSecretsBackend`** — KV v2 over `httpx` (no new dependency) at
  `{mount}/data/{prefix}/{scope}/{workspace|org uuid}/{environment}/{name}`,
  with a short-TTL, size-bounded in-process cache (default 45s). `can_write`
  is `False`.

Both runtime read paths route through it — `AuthSandbox` for workspace secrets
(the `.secrets` contract and "Missing workspace secrets" error are preserved
byte-for-byte) and `SecretsService.get_registry_ssh_key` for org SSH keys.

In Vault mode the DB rows become **value-less registrations** (name + key names
+ type + environment; requests carrying actual values are rejected). A new
`backend` column (additive migration, default `'db'`) marks ownership. The
workspace "Create credential" dialog picks up a `secrets_backend_readonly` flag
from `/info` and submits a registration, so `custom` secrets work from the UI.

Vault auth defaults to `jwt` — a projected Kubernetes service account token
validated by Vault *offline* via `jwt_validation_pubkeys`, which is the whole
point for isolated clusters where Vault can't call back into the kube-apiserver
(only outbound Tracecat→Vault connectivity needed). A static `token` method is
there for local dev. The Vault client and token stay in the executor service
process and never reach action sandboxes.

### What's covered

- New `tests/unit/test_secrets_backend.py` (23 tests): factory, KV v2 reads,
  cache hit/expiry/disable, JWT login + reuse + invalidation, path safety,
  write blocking. Existing suites still green
  (`test_sandbox`/`test_secrets_service`/`test_secrets`/`api/test_api_secrets`,
  101/101). `ruff` clean; frontend `tsc` + biome clean.
- Alembic migration verified from a clean DB, incl. downgrade/upgrade round-trip.
- Smoke-tested against a live `hashicorp/vault` dev server: a `custom` secret
  resolved straight from KV v2, missing secrets returned the usual error.

### Deliberately out of scope

Writing values to Vault; OAuth integrations (stay in DB); other providers
(AWS/Azure/GCP — the protocol is provider-agnostic); auto-syncing registrations
from Vault; the read-only UI treatment for org-secret dialogs and structured
ssh/mtls/ca types (the API already accepts value-less registrations for those).

Docs live in `docs/self-hosting/secrets-backends.mdx`; `.env.example` is updated.

---

---
## Summary by cubic
Introduces a pluggable secrets backend with a read-only HashiCorp Vault option so secret values can be resolved at runtime instead of stored in Postgres. The default behavior remains unchanged (`db`), and the UI now supports value-less registrations when an external backend is active.

- **New Features**
  - Added `SecretsBackend` with `db` (encrypted in DB) and `vault` (KV v2 via `httpx`, read-only) implementations.
  - Vault: JWT or static token auth, optional namespace, short-TTL in-process cache, safe path encoding, and clear 403 handling.
  - Unified resolution: `AuthSandbox` and `SecretsService` read via the selected backend; org SSH keys resolve from the backend in read-only mode.
  - DB: new `backend` column; external-backend rows are registrations (no values). The `db` backend refuses to serve these as values.
  - UI: workspace “Create credential” dialog detects read-only mode from `/info` (`secrets_backend_readonly`) and registers key names only.
  - Docs and `.env.example` updated; extensive unit tests for backend factory, Vault flow, caching, and guards.

- **Migration**
  - Run DB migrations (adds the `backend` column; existing rows default to `db`).
  - To use Vault: set `TRACECAT__SECRETS_BACKEND=vault`, `TRACECAT__VAULT_ADDR`, and configure auth (`TRACECAT__VAULT_AUTH_METHOD=jwt|token`; JWT needs `TRACECAT__VAULT_JWT_ROLE` and a projected token).
  - In Vault mode, create secrets as registrations with empty values; store values in Vault. Keys map 1:1 (SSH uses `PRIVATE_KEY`).

<sup>Written for commit fa2da2609c90098508cedc267b09688b86457821. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2940?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -168,3 +168,21 @@ PYTHONPATH=/home/apiuser/.local:/app/local_registry
 # Feature flags (comma-separated list)
 TRACECAT__FEATURE_FLAGS=""
 TRACECAT__EE_MULTI_TENANT=false
+
+# --- Secrets backend (optional) ---
+# Where workflow secret values live: 'db' (default, encrypted in Postgres)
+# or 'vault' (resolved from HashiCorp Vault at runtime, read-only).
+TRACECAT__SECRETS_BACKEND=db
+# Required when TRACECAT__SECRETS_BACKEND=vault
+TRACECAT__VAULT_ADDR=
+TRACECAT__VAULT_KV_MOUNT=secret
+TRACECAT__VAULT_PATH_PREFIX=tracecat
+# Auth: 'jwt' (projected Kubernetes service account token) or 'token' (dev only)
+TRACECAT__VAULT_AUTH_METHOD=jwt
+TRACECAT__VAULT_JWT_AUTH_MOUNT=jwt
+TRACECAT__VAULT_JWT_ROLE=
+TRACECAT__VAULT_JWT_TOKEN_PATH=/var/run/secrets/tokens/vault-token
+TRACECAT__VAULT_TOKEN=
+TRACECAT__VAULT_NAMESPACE=
+TRACECAT__VAULT_CACHE_TTL_SECONDS=45
+TRACECAT__VAULT_CACHE_MAX_SIZE=1024
--- a/alembic/versions/b3d1a9c47e02_add_secret_backend_marker.py
+++ b/alembic/versions/b3d1a9c47e02_add_secret_backend_marker.py
@@ -1,10 +1,15 @@
 """add secret backend marker
 
 Adds a ``backend`` column to all secret tables. It records which secrets
-backend owns a secret's values: ``db`` (default, values Fernet-encrypted in
+backend a row was created under: ``db`` (default, values Fernet-encrypted in
 ``encrypted_keys``) or an external backend such as ``vault``, in which case
 the row is a value-less registration (name, key names, type, environment)
 and values are resolved from the external backend at runtime.
+
+Runtime dispatch is currently global (``TRACECAT__SECRETS_BACKEND``); the
+marker is an integrity guard (the database backend refuses to serve
+external-backend registrations as empty values) and lays the groundwork for
+future per-secret backend dispatch.
 
 Purely additive: existing rows are backfilled to ``db`` via the server
 default, which matches their actual storage.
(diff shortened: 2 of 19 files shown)


topher-lo:
@TyTcoH thanks for taking a crack at this! Are you in our Discord channel? Because this is a major feature and one that touches the core system with real security impacts, we'd like to go through the design and approach with you together in our Discord #contributors channel first 🙏

PR #2959, opened 2026-07-06
feat(integrations): test stdio mcp connections

jordan-umusu:
## Summary by cubic
Adds sandboxed stdio MCP connection probing with Temporal-based background verification and a unified “test connection” flow for both HTTP and stdio. Catalog rows show “configured” until verified; verified stdio tool summaries are persisted, invalid/approval-gated tools are excluded at runtime, and networking auto-enables when any stdio MCP is used (including subagents).

- **New Features**
  - Stdio probe: runs under `nsjail`; falls back to PID-namespace isolation when unavailable; cached host checks; sanitized error messages; 60s timeout cap.
  - Verification: Temporal workflow with distinct probe and persist activities; saved stdio integrations verify in the background; verification-status API plus a React hook poll via a Temporal handle every 3s with toasts on status changes.
  - Test connection: one flow; HTTP tests work with unsaved configs; stdio tests require a saved integration ID and persist discovered tools on success; dialog clarifies save-before-test.
  - Tool inventory/runtime: persist non-secret stdio tool summaries and include them in workflow payloads; runtime uses only verified tools, filters invalid tool names, and ignores approval-gated stdio tools; UI shows an “approvals unsupported” hint for stdio.
  - Networking: presets and runtime auto-enable internet if any agent or subagent uses a `stdio` MCP.
  - Schema/UI/infra: catalog connect returns “configured” until verification (non-OAuth or no tools yet); new HTTP test and verification-status models/endpoints; improved polling/toasts; Docker adds `git` and `openssh-client`; request logging records metadata only.

- **Migration**
  - Handle the `configured` status for MCP catalog entries and poll verification status until tools are available.

<sup>Written for commit ddfeeeab382baa457b499035735acad1cdd46e56. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2959?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/hooks/use-stdio-mcp-verification-status.ts
+++ b/frontend/src/hooks/use-stdio-mcp-verification-status.ts
@@ -1,0 +1,197 @@
+import {
+  type Query,
+  type QueryClient,
+  useQueries,
+  useQueryClient,
+} from "@tanstack/react-query"
+import { useEffect, useMemo } from "react"
+
+import {
+  type MCPVerificationStatusRead,
+  mcpIntegrationsGetMcpIntegrationVerificationStatus,
+} from "@/client"
+import { toast } from "@/components/ui/use-toast"
+import { MCP_STDIO_VERIFICATION_POLL_INTERVAL_MS } from "@/lib/integrations"
+
+const previousVerificationStatuses = new Map<
+  string,
+  MCPVerificationStatusRead["status"]
+>()
+const MCP_VERIFICATION_TOAST_DETAIL_MAX_LENGTH = 280
+
+type UseStdioMcpVerificationStatusParams = {
+  workspaceId: string
+  pendingIntegrationIds: string[]
+}
+
+export type StdioMcpVerificationStatuses = ReadonlyMap<
+  string,
+  MCPVerificationStatusRead
+>
+
+/**
+ * Polls Temporal-backed stdio MCP verification status for pending integrations.
+ */
+export function useStdioMcpVerificationStatus({
+  workspaceId,
+  pendingIntegrationIds,
+}: UseStdioMcpVerificationStatusParams): StdioMcpVerificationStatuses {
+  const queryClient = useQueryClient()
+  const mcpIntegrationIds = useMemo(
+    () => Array.from(new Set(pendingIntegrationIds)).sort(),
+    [pendingIntegrationIds]
+  )
+
+  const statusQueries = useQueries({
+    queries: mcpIntegrationIds.map((mcpIntegrationId) => ({
+      queryKey: [
+        "mcp-verification-status",
+        workspaceId,
+        mcpIntegrationId,
+      ] as const,
+      queryFn: async () =>
+        await mcpIntegrationsGetMcpIntegrationVerificationStatus({
+          workspaceId,
+          mcpIntegrationId,
+        }),
+      enabled: Boolean(workspaceId && mcpIntegrationId),
+      refetchInterval: stdioMcpVerificationRefetchInterval,
+      refetchOnWindowFocus: false,
+    })),
+  })
+
+  useEffect(() => {
+    for (const [index, query] of statusQueries.entries()) {
+      const mcpIntegrationId = mcpIntegrationIds[index]
+      const statusRead = query.data
+      if (!workspaceId || !mcpIntegrationId || !statusRead) {
+        continue
+      }
+
+      const key = verificationStatusKey(workspaceId, mcpIntegrationId)
+      const previousStatus = previousVerificationStatuses.get(key)
+      if (previousStatus === statusRead.status) {
+        continue
+      }
+
+      previousVerificationStatuses.set(key, statusRead.status)
+
+      if (statusRead.status === "succeeded") {
+        invalidateMcpVerificationQueries(
+          queryClient,
+          workspaceId,
+          mcpIntegrationId
+        )
+        continue
+      }
+
+      if (statusRead.status === "failed" && previousStatus !== "failed") {
+        toast({
+          title: "MCP server verification failed",
+          description: verificationFailureToastDescription(statusRead.error),
+          variant: "destructive",
+        })
+      }
+    }
+  }, [mcpIntegrationIds, queryClient, statusQueries, workspaceId])
+
+  return new Map(
+    mcpIntegrationIds.flatMap((mcpIntegrationId, index) => {
+      const statusRead = statusQueries[index]?.data
+      return statusRead ? [[mcpIntegrationId, statusRead] as const] : []
+    })
+  )
+}
+
+function verificationFailureToastDescription(
+  error: string | null | undefined
+): string {
+  if (!error) {
+    return "Stdio MCP verification failed. Open the failure badge for details."
+  }
+
+  const meaningfulLine = error
+    .split("\n")
+    .map((line) => line.trim())
+    .reverse()
+    .find((line) => line && !isStdioProbeStartupNoise(line))
+  if (!meaningfulLine) {
+    return "Stdio MCP verification failed. Open the failure badge for details."
+  }
+
+  const detail =
+    meaningfulLine.length > MCP_VERIFICATION_TOAST_DETAIL_MAX_LENGTH
+      ? `${meaningfulLine.slice(0, MCP_VERIFICATION_TOAST_DETAIL_MAX_LENGTH - 1)}…`
+      : meaningfulLine
+  return `${detail} — Open the failure badge for details.`
+}
+
+function isStdioProbeStartupNoise(line: string): boolean {
+  return (
+    /^(Downloading|Downloaded|Installed)\b/.test(line) ||
+    /^AuthlibDeprecationWarning:/.test(line) ||
+    /^[╭╮╰╯│─\s]+$/.test(line) ||
+    (line.startsWith("│") && line.endsWith("│"))
+  )
+}
+
+/**
+ * Seeds a newly started verification in the query cache and resumes polling.
+ *
+ * Failed verification queries are terminal and stop polling. A reconnect uses
+ * the same integration and query key, so explicitly moving that key back to
+ * `verifying` prevents a cached failure from hiding the replacement workflow.
+ */
+export function markStdioMcpVerificationStarted(
+  queryClient: QueryClient,
+  workspaceId: string,
+  mcpIntegrationId: string
+): void {
+  const queryKey = [
+    "mcp-verification-status",
+    workspaceId,
+    mcpIntegrationId,
+  ] as const
+  previousVerificationStatuses.set(
+    verificationStatusKey(workspaceId, mcpIntegrationId),
+    "verifying"
+  )
+  queryClient.setQueryData<MCPVerificationStatusRead>(queryKey, {
+    status: "verifying",
+    error: null,
+  })
+  void queryClient.invalidateQueries({ queryKey })
+}
+
+function verificationStatusKey(
+  workspaceId: string,
+  mcpIntegrationId: string
+): string {
+  return `${workspaceId}:${mcpIntegrationId}`
+}
+
+function stdioMcpVerificationRefetchInterval(
+  query: Query<MCPVerificationStatusRead>
+): number | false {
+  const status = query.state.data?.status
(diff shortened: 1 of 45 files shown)


tracecat/agent/mcp/stdio_probe.py
```diff
@@ -0,0 +1,374 @@
...
+            message="Failed to connect to the stdio MCP server",
+            error=error,
+        )
+
+    tools = _parse_probe_tools(result.output)
+    return StdioMCPProbeResult(
+        success=True,
+        tools=tools,
+        message=f"Connected successfully — {len(tools)} tools available",
+    )
```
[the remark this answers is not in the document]
jordan-umusu: @zeropath-ai accepted-risk because many users do not have access to sandboxes and must rely on best effort isolation.

tracecat/agent/executor/activity.py
```diff
@@ -907,6 +924,103 @@ async def _hydrate_stdio_env(
...
+) -> StdioMCPProbeResult:
+    """Resolve env templates, validate, and probe an stdio config on the sandbox.
+
+    ``stdio_env`` carries template references and workspace-owned plaintext
+    config loaded from the saved integration row; secrets are resolved here
+    inside the executor, never across the workflow boundary.
+    """
+    if stdio_env:
+        try:
+            stdio_env = await preset_svc._resolve_stdio_env(
```
daryllimyt: nit: public

tracecat/agent/executor_worker.py
```diff
@@ -33,7 +36,10 @@
 
 def get_activities() -> list:
     """Load runtime activities registered by the agent-executor worker."""
-    return [run_agent_activity]
+    return [
+        run_agent_activity,
+        probe_stdio_mcp_connection_activity,
```
daryllimyt, line 41: **Non-blocking capacity note: stdio probes share agent runtime slots**

This does not need to block the PR if stdio verification volume is expected to stay low, but worth being aware of the operational tradeoff: registering `probe_stdio_mcp_connection_activity` here means stdio verification runs on `TRACECAT__AGENT_EXECUTOR_QUEUE`, which is documented as reserved for `run_agent_activity` so agent runtime capacity can be provisioned independently. Stdio probes are bounded, but they can still launch a user-configured process with network access for up to the probe timeout cap, so a burst of catalog/save verification could consume the same limited activity slots as interactive agent turns. If this becomes visible in practice, the follow-up would be to move probes to a separate sandbox-capable queue/worker or add a small probe-specific concurrency cap.

tracecat/integrations/service.py
```diff
@@ -3340,6 +3435,58 @@ async def _record_mcp_verification_failure(
...
+    def start_mcp_stdio_verification(self, *, mcp_integration: MCPIntegration) -> None:
+        """Launch saved stdio MCP verification without blocking the request."""
+        if mcp_integration.server_type != "stdio":
+            raise ValueError(
+                "Only stdio MCP integrations can be verified asynchronously"
+            )
+        if mcp_integration.id is None:
+            raise ValueError("MCP integration must be saved first")
+
+        task = asyncio.create_task(
```
daryllimyt: **blocking:** The probe itself is durable (it runs as `StdioMCPProbeWorkflow` on Temporal), but the waiter-and-persister is a process-local `asyncio.create_task` in the API process. `execute_workflow` is awaited here, and the `configured` → `connected` transition (lock, merge tools, commit) only happens on the API side after that await returns.

If the API restarts mid-verification (deploy, crash, OOM), Temporal runs the probe to completion and returns the tools to nobody — the row is stranded in `configured` with no retry, and the only recovery is the user manually re-triggering. Deploys are exactly when this will fire.

Suggestion: draw the durable boundary around the whole operation instead:

1. API calls `client.start_workflow(...)` and returns immediately — no in-process task set, no done-callbacks, no bypass-RLS background session (`_verify_mcp_integration_in_background` goes away).
2. Move the persist step (lock → merge → commit) into the workflow as an activity. The probe activity (`probe_stdio_mcp_connection_activity`) already opens a role-scoped session and uses `IntegrationService` to load the row and decrypt the env, so the worker demonstrably has everything it needs to also persist. Keep the merge-base semantics intact (sync test-connection snapshots `previous_tools` before probing; this path uses the row's current tools).
3. Derive the workflow id from the integration's scope + id instead of `uuid4()`, e.g. `mcp-stdio-probe/ws/{workspace_id}/{mcp_integration_id}` (or `mcp-stdio-probe/org/{org_id}/{mcp_integration_id}` for org-scoped rows). That gives Temporal-level dedupe: double-saving an integration can't race two concurrent probes for the same row. Today that race is only serialized late, at the row lock in the merge step.

The frontend doesn't need a new polling surface — the row state remains the poll target — but with a deterministic workflow id the handle is also trivially recoverable if we ever want to expose it.

tracecat/middleware/request.py
```diff
@@ -21,30 +86,33 @@ async def dispatch(self, request: Request, call_next):
...
+                # Log the incoming request with parameters. Bodies can carry provider
+                # credentials, so redact secret-bearing keys before logging.
+                request_logger.debug(
+                    "Incoming request",
+                    method=request.method,
+                    scheme=request.url.scheme,
+                    hostname=request.url.hostname,
+                    path=request.url.path,
+                    params=request_params,
+                    body=redact_request_body(request_body),
```
daryllimyt: why not just stop logging `body` entirely? this adds recursive parsing over every single request with a body, even though capped at 6 depth can be very wide (think webhook payloads?)

daryllimyt:
Missing test coverage: we should add a Docker-backed nsjail smoke test for stdio MCP probing that runs `probe_stdio_mcp_tools_in_sandbox()` against a real FastMCP stdio server.

Current coverage mocks `NsjailExecutor` for the nsjail branch and only runs a real FastMCP server through the non-nsjail fallback path (`tests/unit/test_stdio_probe.py`).

daryllimyt (changes requested):
see comments

tracecat/agent/executor/activity.py
jordan-umusu: done, also made `validate_stdio_server_config` public since it's the same pattern right below

tracecat/agent/executor_worker.py
jordan-umusu: yeah decided to just fix it now. probes run on their own task queue (`TRACECAT__AGENT_EXECUTOR_PROBE_QUEUE`) with a second worker in the same executor process, small concurrency cap (default 4). agent turns keep their slots.

tracecat/integrations/service.py
jordan-umusu: done, went with your suggestion. API just does `start_workflow` and returns, persist (lock/merge/commit) is an activity in the workflow now, and `_verify_mcp_integration_in_background` is gone. workflow id is `mcp-stdio-probe/ws/{workspace_id}/{mcp_integration_id}` so double-saves dedupe in Temporal. merge semantics unchanged, sync test-connection still snapshots `previous_tools`.

also switched the conflict policy to `TERMINATE_EXISTING` so a save during an in-flight probe supersedes it instead of being skipped — otherwise a stale probe of the old config could persist its tools onto the updated row.

tracecat/middleware/request.py
jordan-umusu: fair, dropped it entirely. body is never read now, redaction stuff deleted too.

jordan-umusu:
added `tests/unit/test_stdio_probe_nsjail.py`, runs `probe_stdio_mcp_tools_in_sandbox()` through real nsjail against a real FastMCP stdio server. success + bad-command cases, gated like the other nsjail tests (nsjail + rootfs + in docker).

tracecat/middleware/request.py
```diff
@@ -1,9 +1,21 @@
...
 from starlette.middleware.base import BaseHTTPMiddleware
 
 from tracecat.contexts import ctx_client_ip
 
 
+def _is_debug_logging_enabled(logger: Any) -> bool:
+    return (
+        logger.level(os.environ.get("LOG_LEVEL", "INFO").upper()).no
+        <= logger.level("DEBUG").no
+    )
```
daryllimyt: isn't this already what logger.debug does?

tracecat/config.py
```diff
@@ -820,6 +820,17 @@ def _parse_auth_types() -> set[AuthType]:
 This queue is reserved for `run_agent_activity` so agent runtime capacity can be
 provisioned independently from agent workflow/control-plane work."""
 
+TRACECAT__AGENT_EXECUTOR_PROBE_QUEUE = os.environ.get(
```
daryllimyt: we don't need this - should just go on the agent-worker queue, each temporal worker can only poll a single queue

tracecat/agent/workflows/mcp_probe.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+
+@workflow.defn
+class StdioMCPProbeWorkflow:
+    """Probe a saved stdio MCP integration via the executor sandbox."""
+
+    @workflow.run
+    async def run(self, input: StdioMCPProbeWorkflowInput) -> StdioMCPProbeResult:
+        probe_result = await workflow.execute_activity(
```
daryllimyt: nit: should type the return value, i think there's a generic param?

tracecat/agent/workflows/mcp_probe.py
```diff
@@ -0,0 +1,57 @@
...
+            MCP_STDIO_PROBE_ACTIVITY_NAME,
+            StdioMCPProbeInput(
+                mcp_integration_id=input.mcp_integration_id,
+                role=input.role,
+            ),
+            task_queue=config.TRACECAT__AGENT_EXECUTOR_PROBE_QUEUE,
```
daryllimyt: probe should run on the agent-executor queue since it needs the sandbox

tracecat/agent/workflows/mcp_probe.py
```diff
@@ -0,0 +1,57 @@
...
+            return probe_result
+
+        return await workflow.execute_activity(
+            MCP_STDIO_PERSIST_ACTIVITY_NAME,
+            StdioMCPPersistInput(
+                mcp_integration_id=input.mcp_integration_id,
+                role=input.role,
+                tools=probe_result.tools,
+            ),
+            task_queue=config.TRACECAT__AGENT_EXECUTOR_PROBE_QUEUE,
```
daryllimyt: nit: this is crud/db ops so it can run on agent-worker queue if we want to optimize

tracecat/agent/executor_worker.py
```diff
@@ -77,6 +91,10 @@ async def main(shutdown_event: asyncio.Event | None = None) -> None:
     max_concurrent = int(
         os.environ.get("TRACECAT__AGENT_EXECUTOR_MAX_CONCURRENT_ACTIVITIES") or 1
     )
+    probe_max_concurrent = int(
+        os.environ.get("TRACECAT__AGENT_EXECUTOR_PROBE_MAX_CONCURRENT_ACTIVITIES")
```
daryllimyt: shouldn't need this

daryllimyt (changes requested):
just need to set the right queues

frontend/src/hooks/use-stdio-mcp-verification-status.ts
```diff
@@ -0,0 +1,124 @@
...
+
+      const key = verificationStatusKey(workspaceId, mcpIntegrationId)
+      const previousStatus = previousVerificationStatuses.get(key)
+      if (previousStatus === statusRead.status) {
+        continue
+      }
+
+      previousVerificationStatuses.set(key, statusRead.status)
+
+      if (statusRead.status === "succeeded") {
```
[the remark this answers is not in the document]
jordan-umusu: Refetching on failure would be a no-op with network cost.

daryllimyt (approved):
lgtm

PR #2983, opened 2026-07-07
fix(ui): include pending approval as streamable status

jordan-umusu:
## Summary by cubic
Treat `PENDING_APPROVAL` sessions as live and resume them by rotating to a new per-turn stream after approval. Clients attach to the returned `active_stream_id`, and we mark an approval-continuation start so panes always stream the correct, open turn.

- **Bug Fixes**
  - Frontend: adopt the server transcript wholesale on `ready` via `useAdoptServerTranscript`; always refetch on mount in `useGetChatVercel`; simplify Inbox resume to the session’s live status; invalidate chat-turn queries from the approvals dialog via `invalidateChatTurnQueries`; validate approval-request payloads with `isApprovalCardArray` and dedupe duplicates (keep newest); compact tool calls across the DB/stream seam.
  - Backend: rotate per-turn streams on approval and return `active_stream_id` in `ChatResponse`; mark approval-continuation start in Redis and use reverse-range checks (`xrevrange`) to detect open vs paused; add `emit_session_done` and defer `stream.done` to the workflow for approval closure; unify Vercel SSE headers; ensure the router attaches to rotated streams and stabilizes continuation even if the previous stream expired.

<sup>Written for commit 71e491e8b0e232b5b0e5adf42f4657b840952925. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/2983?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tracecat/agent/session/service.py
```diff
@@ -2100,35 +2162,78 @@ async def list_messages(
...
+        if not include_active and current_run_id is not None:
+            all_entries = [
+                entry
+                for entry in all_entries
+                if entry.curr_run_id != current_run_id
+                or (
+                    current_run_approval_boundary is not None
+                    and entry.surrogate_id <= current_run_approval_boundary
+                )
+            ]
```
daryllimyt: Could we split this method into query/visibility and rendering responsibilities, and drop this in-memory filter entirely?

The SQL predicate immediately above already enforces the same rule. Keeping this list comprehension gives production two sources of truth, and the current mock-based tests only need it because their mocked query returns rows that the real database would exclude.

Suggested structural sketch:

```python
async def list_messages(
    self,
    session_id: uuid.UUID,
    *,
    kinds: Sequence[MessageKind] | None = None,
    include_active: bool = False,
) -> list[ChatMessage]:
    agent_session = await self.get_session(session_id)
    if agent_session is None:
        return await ChatService(
            self.session, self.role
        ).list_legacy_messages(session_id, kinds=kinds)

    session_ids = [session_id]
    if agent_session.parent_session_id:
        session_ids.insert(0, agent_session.parent_session_id)

    approvals_by_tool_id = await self._approvals_by_tool_id(session_ids)
    entries = await self._visible_history_entries(
        session_ids=session_ids,
        current_run_id=agent_session.curr_run_id,
        approval_tool_call_ids=set(approvals_by_tool_id),
        include_active=include_active,
    )

    return self._render_history_entries(
        entries,
        approvals_by_tool_id=approvals_by_tool_id,
        kinds=kinds,
    )


async def _latest_approval_boundary(
    self,
    *,
    session_ids: Sequence[uuid.UUID],
    current_run_id: uuid.UUID,
    approval_tool_call_ids: set[str],
) -> int | None:
    stmt = (
        select(
            AgentSessionHistory.surrogate_id,
            AgentSessionHistory.content,
        )
        .where(
            AgentSessionHistory.session_id.in_(session_ids),
            AgentSessionHistory.curr_run_id == current_run_id,
        )
        .order_by(AgentSessionHistory.surrogate_id.desc())
    )
    result = await self.session.execute(stmt)

    for surrogate_id, content in result.tuples():
        inner_message = content.get("message") or content
        if any(
            tool_use.get("id") in approval_tool_call_ids
            for tool_use in self._extract_tool_uses_from_message(inner_message)
        ):
            return surrogate_id

    return None


async def _visible_history_entries(
    self,
    *,
    session_ids: Sequence[uuid.UUID],
    current_run_id: uuid.UUID | None,
    approval_tool_call_ids: set[str],
    include_active: bool,
) -> list[AgentSessionHistory]:
    approval_boundary = None
    if (
        not include_active
        and current_run_id is not None
        and approval_tool_call_ids
    ):
        approval_boundary = await self._latest_approval_boundary(
            session_ids=session_ids,
            current_run_id=current_run_id,
            approval_tool_call_ids=approval_tool_call_ids,
        )

    stmt = (
        select(AgentSessionHistory)
        .where(AgentSessionHistory.session_id.in_(session_ids))
        .order_by(AgentSessionHistory.surrogate_id)
    )

    if not include_active and current_run_id is not None:
        visible = or_(
            AgentSessionHistory.curr_run_id.is_(None),
            AgentSessionHistory.curr_run_id != current_run_id,
        )
        if approval_boundary is not None:
            visible = or_(
                visible,
                and_(
                    AgentSessionHistory.curr_run_id == current_run_id,
                    AgentSessionHistory.surrogate_id <= approval_boundary,
                ),
            )
        stmt = stmt.where(visible)

    result = await self.session.execute(stmt)
    return list(result.scalars().all())  # No post-query Python re-filter.


def _render_history_entries(
    self,
    entries: Sequence[AgentSessionHistory],
    *,
    approvals_by_tool_id: Mapping[str, Approval],
    kinds: Sequence[MessageKind] | None,
) -> list[ChatMessage]:
    messages: list[ChatMessage] = []
    internal_uuids: set[str] = set()

    for entry in entries:
        content = entry.content
        if not content:
            continue

        if entry.kind == MessageKind.INTERNAL.value:
            if line_uuid := session_line_uuid(content):
                internal_uuids.add(line_uuid)
            continue

        if is_continuation_control_artifact(content, internal_uuids):
            if line_uuid := session_line_uuid(content):
                internal_uuids.add(line_uuid)
            continue

        if entry.kind == MessageKind.COMPACTION.value:
            emitted = [self._compaction_message(entry)]
        elif entry.kind == MessageKind.CANCELLED.value:
            emitted = [self._cancelled_message(entry)]
        else:
            emitted = self._chat_and_approval_messages(
                entry,
                approvals_by_tool_id=approvals_by_tool_id,
            )

        messages.extend(
            message
            for message in emitted
            if not kinds or message.kind in kinds
        )

    return messages
```

The block on this thread would therefore be deleted, not moved:

```python
if not include_active and current_run_id is not None:
    all_entries = [
        entry
        for entry in all_entries
        if entry.curr_run_id != current_run_id
        or (
            current_run_approval_boundary is not None
            and entry.surrogate_id <= current_run_approval_boundary
        )
    ]
```

Rendering unit tests should mock only the rows that SQL would return. A database-backed test should own the `NULL`, current-run, and approval-boundary visibility semantics.

tracecat/agent/session/service.py
```diff
@@ -1720,13 +1760,29 @@ async def _continue_with_approvals(
...
+        new_stream_id = uuid.uuid4()
+        continuation_stream = await AgentStream.new(
+            session_id=session_id,
+            workspace_id=self.workspace_id,
+            stream_id=new_stream_id,
+        )
+        await continuation_stream.mark_approval_continuation()
+        agent_session.active_stream_id = new_stream_id
+        self.session.add(agent_session)
+        await self.session.commit()
```
daryllimyt: This makes Stream B the persisted session pointer before Temporal accepts the update, but the rotation is not transactional.

I would rate this as a **medium reliability risk**: normal-path probability is low, but the failure is significant.

- A definite Temporal rejection returns 500, leaves approvals pending, and leaves marker-only B looking live to reconnecting clients, which can then wait indefinitely. A manual retry usually recovers because the dedup key is cleared.
- An ambiguous RPC failure is more serious: Temporal may have accepted the update even though the client raises. Retrying then rotates to Stream C while the workflow may already be writing to B, potentially splitting the continuation across streams.
- The setup operations before the current `try` can also fail while leaving the dedup reservation in place for five minutes, temporarily suppressing retries.

I think this should be hardened before merge because it can recreate the same “approval stream appears live but never finishes” failure this PR addresses. For a definite rejection, close B, compare-and-restore the previous pointer, and clear dedup. For an unknown outcome, reuse B with a stable Temporal update ID instead of rotating again.

tracecat/chat/schemas.py
```diff
@@ -76,6 +76,15 @@ class ChatResponse(BaseModel):
 
     stream_url: str = Field(..., description="URL to connect for SSE streaming")
     chat_id: uuid.UUID = Field(..., description="Unique chat identifier")
+    active_stream_id: uuid.UUID | None = Field(
+        default=None,
+        description=(
+            "Per-turn Redis stream id the caller should attach to. On approval "
+            "continuations this is the freshly rotated stream id (a suffix-only "
```
daryllimyt: Could we avoid the **prefix/suffix** terminology here? Those terms describe stream position, but they do not explain that the boundary is specifically the approval pause. I think this would be clearer:

> Per-turn Redis stream ID the caller should attach to. For approval continuations, this is a newly rotated stream containing only events emitted after approval resumes. Callers must attach to this stream instead of the stream that ended at the approval pause, which may already have expired.

The same prefix/suffix wording appears in `durable.py` and `session/service.py`; using approval-specific lifecycle language consistently would make the flow easier to understand without already knowing the Stream A/Stream B model.

frontend/src/lib/chat.ts
```diff
@@ -350,6 +350,22 @@ type CompactionData = {
   phase?: "started" | "completed" | "failed"
 }
 
+type KeptToolPart = {
+  position: string
+  hasOutput: boolean
+}
+
+function getApprovalToolCallId(approval: unknown): string | null {
```
daryllimyt: no unknown types - the types are known? `ApprovalCard`

frontend/src/lib/chat.ts
```diff
@@ -478,11 +526,36 @@ export function transformMessages(messages: ai.UIMessage[]): ai.UIMessage[] {
...
     const newParts: ai.UIMessagePart<ai.UIDataTypes, ai.UITools>[] = []
     for (const [j, part] of message.parts.entries()) {
       const posKey = `${i}-${j}`
       if (ignorePos.has(posKey)) continue
+      if (
+        shouldFilterApprovalParts &&
+        part.type === "data-approval-request" &&
+        Array.isArray(part.data)
+      ) {
+        const remaining = part.data.filter((approval) => {
```
daryllimyt: if this ends up being inferred as `any` means you aren't doing enough to type narrow it

frontend/src/components/chat/chat-session-pane.tsx
```diff
@@ -172,11 +172,22 @@ function isCancelAttributionContentPart(
...
-  return messages
-    .map((m) => `${m.id}:${m.parts.map((p) => p.type).join(",")}`)
-    .join("|")
+/**
+ * Identity signature of a transcript: message ids plus the id, type, state, and
+ * text of each part. Two transcripts with the same signature carry the same
+ * rendered content, so an unchanged signature means there is nothing to adopt.
+ */
+function transcriptSignature(messages: UIMessage[]): string {
+  return JSON.stringify(
```
daryllimyt: is this ordered/stable?

daryllimyt (changes requested):

tracecat/agent/session/service.py
```diff
@@ -1215,6 +1202,154 @@ def _approval_dedup_key(
...
+            )
+            .values(active_stream_id=stream_id)
+            .returning(AgentSession.active_stream_id)
+        )
+        try:
+            result = await self.session.execute(stmt)
+            installed_stream_id = result.scalar_one_or_none()
+            await self.session.commit()
+        except SQLAlchemyError:
+            await self.session.rollback()
```
daryllimyt, line 1277: does this rollback more than necessary?

tracecat/agent/session/service.py
```diff
@@ -1215,6 +1202,154 @@ def _approval_dedup_key(
...
+    ) -> ApprovalContinuationAttempt:
+        """Reuse an installed attempt or atomically install a fresh stream."""
+        if attempt := await self._existing_approval_continuation_attempt(
+            agent_session,
+            submission_key=submission_key,
+        ):
+            return attempt
+
+        previous_stream_id = agent_session.active_stream_id
+        stream_id = uuid.uuid4()
```
daryllimyt, line 1245: if this runs in the temporal workflow use workflow.uuid4()

daryllimyt (approved):

tracecat/agent/session/service.py
jordan-umusu: Not called within a temporal workflow, Its only production caller is _continue_with_approvals (service.py:2005), reached only via run_turn (service.py:1589).

tracecat/agent/session/service.py
jordan-umusu: no, every statement before this on all continuation paths is a SELECT, so this only rolls back the attempted stream id rotation.

PR #3015, opened 2026-07-10
feat(slack): add optional hash input to update_view action

chr-ge:
## Summary

Adds an optional `hash` input to the `tools.slack.update_view` action so callers can pass Slack's view state hash to protect `views.update` against race conditions (per Slack API docs).

- New optional input `hash: str | None` (default `null`), described as "A string that represents view state to protect against possible race conditions."
- When `hash` is omitted, the step falls back to the previous behavior of sending a UTC timestamp (`FN.to_isoformat(FN.utcnow())`), so existing workflows keep working unchanged.

---
## Summary by cubic
Add an optional `hash` input to `tools.slack.update_view` so callers can pass Slack’s view state hash to avoid race conditions on views.update. When `hash` is omitted, we now send `null` so Slack performs an unguarded update (no timestamp fallback).

<sup>Written for commit 681469c777f12d60be1079d8169abaffa882bf5d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3015?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


daryllimyt:
@codex review

PR #3019, opened 2026-07-11
fix(cases): serialize attachment quota checks per case

### LINKED ISSUE #2970: cases: serialize attachment quota checks during uploads
## Problem

Case attachment uploads enforce per-case attachment count and storage limits with aggregate reads before the new attachment row is inserted. Without serializing concurrent uploads for the same case, two uploads can both observe the same pre-insert count/storage and then both commit, allowing the case to exceed its configured limits.

## Evidence

The upload path currently calls `CaseAttachmentService._assert_case_limits()` from `create_attachment()` before creating or restoring the file/attachment records. `_assert_case_limits()` uses count and storage aggregate queries, but the service does not acquire a per-case row lock or equivalent serialization point around the quota check and subsequent insert.

This differs from other case mutation paths that already use `SELECT FOR UPDATE` when updating a case.

## Reproduction

Safe high-level reproduction:

1. Configure a case near its maximum attachment count or storage limit.
2. Start two uploads to the same case at the same time.
3. Both requests can pass the aggregate limit check before either request commits its attachment.
4. The case can end up over the intended count or storage limit.

## Actual behavior

Concurrent uploads can make quota decisions from stale aggregate state.

## Expected behavior

For a single case, attachment quota checks and the subsequent attachment write should be serialized so each upload sees the latest committed/in-flight case attachment state before deciding whether the upload is allowed.

## Impact

This is a production data-integrity and reliability issue. Per-case limits are used to bound storage and attachment fan-out; concurrent uploads should not be able to bypass those limits by racing.

## Proposed fix

Acquire a row-level lock on the target `Case` before running the aggregate count/storage limit checks and hold that lock through the attachment write transaction. This keeps the fix small and matches the existing `for_update` pattern used by case mutation code.

## Validation plan

- Add focused unit coverage proving the lock query uses `SELECT FOR UPDATE` and is scoped to the case/workspace.
- Add regression coverage proving `create_attachment()` acquires the case write lock before running `_assert_case_limits()`.
- Run focused backend tests plus lint/type checks for the changed files.

## Contribution

I can open a focused PR with the row-locking fix and regression coverage.

abhay-codes07:
## Checklist

## Description

`CaseAttachmentService.create_attachment` validates the per-case attachment count and storage limits with aggregate reads (`COUNT(*)` / `SUM(size)`) before inserting the new attachment. Nothing serializes concurrent uploads for the same case, so two requests can both observe the same pre-insert aggregates, both pass `_assert_case_limits`, and both commit — leaving the case over its configured limits.

This PR takes a row-level lock on the case (`SELECT ... FOR UPDATE`) right before the limit checks, mirroring what other case mutation paths already do via `CasesService.get_case(..., for_update=True)`. The lock is scoped to the upload transaction, so concurrent uploads to the same case queue behind each other and each one decides against committed state, while uploads to different cases stay fully parallel.

## Related Issues

Fixes #2970

## Screenshots / Recordings

N/A (no UI changes)

## Steps to QA

Added a regression test that reproduces the race with real concurrency — two uploads racing on independent database connections (not the savepoint-bound test session), with `TRACECAT__MAX_ATTACHMENTS_PER_CASE=1`:

```
uv run pytest tests/unit/test_case_attachments_service.py
```

- `test_concurrent_uploads_cannot_exceed_attachment_limit`: exactly one upload wins, the other gets `MaxAttachmentsExceededError`, and the case ends with exactly one attachment.
- The other 12 tests in the file pass unchanged, confirming single-uploader behavior is untouched.

---
## Summary by cubic
Serialize per-case attachment quota checks by taking a row-level lock on the case in `CaseAttachmentService.create_attachment`. Concurrent uploads to the same case now queue, enforcing count/storage limits correctly; includes a regression test proving only one upload succeeds when the per-case limit is 1.

<sup>Written for commit bb1645e8ea3f6fe4a96fbf778e4ea972bddf0c37. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/unit/test_case_attachments_service.py
+++ b/tests/unit/test_case_attachments_service.py
@@ -1,11 +1,14 @@
+import asyncio
 import base64
 import os
 import uuid
+from datetime import UTC, datetime
 from io import BytesIO
 from urllib.parse import parse_qs, urlparse
 
 import pytest
 from dotenv import dotenv_values
+from sqlalchemy import delete
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from tracecat import config
@@ -16,6 +19,8 @@ from tracecat.cases.attachments.service import CaseAttachmentService
 from tracecat.cases.enums import CaseEventType, CasePriority, CaseSeverity, CaseStatus
 from tracecat.cases.schemas import CaseCreate
 from tracecat.cases.service import CaseEventsService, CasesService
+from tracecat.db.engine import get_async_session_context_manager
+from tracecat.db.models import Case, CaseAttachment, CaseEvent, File
 from tracecat.exceptions import TracecatAuthorizationError, TracecatException
 from tracecat.storage.blob import ensure_bucket_exists
 from tracecat.storage.exceptions import (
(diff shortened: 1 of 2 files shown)


topher-lo:
@cubic rview

topher-lo (approved):

PR #3023, opened 2026-07-11
fix(api): return adjacent page for reverse cursor pagination

### LINKED ISSUE #3022: Reverse cursor pagination returns the wrong page in cases, tables, and case rows endpoints
## Problem

Basically people have two 

* [https://github.com/TracecatHQ/tracecat/pull/3131 ](<https://github.com/TracecatHQ/tracecat/pull/3131>)
* [https://github.com/TracecatHQ/tracecat/pull/3023 ](<https://github.com/TracecatHQ/tracecat/pull/3023>)

Consolidate and close ^ and create new PR.

Several cursor-paginated list endpoints mishandle `reverse=true` (the "previous page" direction), each in a slightly different way:

1. `CasesService.search_cases` (`tracecat/cases/service.py`) flips the cursor comparison for reverse but keeps the forward `ORDER BY`. With `WHERE sort < cursor ORDER BY sort DESC LIMIT n`, the query returns the first `n` rows of the entire backward prefix — i.e. page 1 — instead of the rows immediately before the cursor. It also never swaps `has_more`/`has_previous`, and its reverse branch overwrites `next_cursor` while leaving `prev_cursor` as `null`.
2. `TablesService.list_rows` (`tracecat/tables/service.py`, the search-style method used by `search_rows`) has the same missing `ORDER BY` flip, then reverses the wrong window, so the returned rows are both the wrong page and (relative to the requested sort) mis-ordered.
3. `CaseTableRowsService.list_rows` (`tracecat/cases/rows/service.py`) tries to flip the `ORDER BY` for reverse, but the base statement is already built with `.order_by(created_at.desc(), id.desc())` and the reverse branch *appends* ascending keys instead of replacing them — SQLAlchemy keeps the first sort keys, so the flip never takes effect and the same wrong-window problem occurs. On top of that, `next_cursor`/`prev_cursor` are computed from the already-reversed items and then swapped again, so after one backward page "next" points at the previous page and "prev" points at the next page.

For comparison, `WorkflowsManagementService.list_workflows` (`tracecat/workflow/management/management.py`) and `TableEditorService.list_rows` implement this correctly: flip the scan direction, fetch `limit + 1`, compute cursors from scan order, then reverse the items and swap the cursors/flags

## Reproduction

Using the cases list as an example (same shape for the other two):

1. Create 9 cases in a workspace and page forward with `limit=3` (`GET /cases?limit=3`, then follow `next_cursor` twice) so you're on page 3.
2. Request the previous page: `GET /cases?limit=3&cursor=<page3.prev_cursor>&reverse=true`.

## Actual behavior

The response contains the cases from **page 1**, not page 2. `has_previous` is `true` even at the top of the list, and `prev_cursor` is `null` so you cannot keep paging backward.

For case table rows, going back from page 3 returns the two newest links (page 1's window) in ascending order, and the swapped cursors point the wrong way.

## Expected behavior

`reverse=true` should return the rows immediately preceding the cursor in display order, with `next_cursor`/`prev_cursor`/`has_more`/`has_previous` all oriented in the response (display) direction — matching what `TableEditorService.list_rows` and `WorkflowsManagementService.list_workflows` already do.

Existing reverse-pagination tests only navigate back from page 2, where "previous page" and "page 1" coincide, which is why none of this is caught.

I have a fix ready with regression tests that page three deep and walk back — PR incoming.

abhay-codes07:
## Checklist

## Description

Fixes reverse (`reverse=true`, i.e. "previous page") cursor pagination in three list endpoints. The reference implementations already in the codebase (`TableEditorService.list_rows`, `WorkflowsManagementService.list_workflows`) do it in four steps: flip the scan direction, fetch `limit + 1`, compute cursors from scan order, then reverse the items and swap cursors/flags. Each of the fixed endpoints was missing a different piece:

- **`CasesService.search_cases`**: flipped the cursor comparison but not the `ORDER BY`, so `LIMIT` kept the first rows of the whole backward prefix — "previous page" returned page 1 from any page 3 or deeper. It also never swapped `has_more`/`has_previous` and its reverse branch overwrote `next_cursor` while leaving `prev_cursor` as `null`. Now uses the canonical scan-flip + reverse + swap.
- **`TablesService.list_rows`** (the search-style method behind `search_rows`): same missing `ORDER BY` flip; the post-fetch reversal/swap was already there, so it reversed the wrong window. Now flips the scan direction to match.
- **`CaseTableRowsService.list_rows`**: had two problems. Its reverse branch appended ascending sort keys to a statement already built with `.order_by(created_at.desc(), id.desc())` — SQLAlchemy keeps the first sort keys, so the intended flip never took effect and the same wrong window came back. It also computed cursors from the already-reversed items and then swapped them again, so "next" pointed backward and "prev" pointed forward. The reverse branch now replaces the ORDER BY (`order_by(None)` first), and cursors are computed from scan order before the reversal, like the siblings.

Forward pagination is untouched in all three (the reverse handling is a no-op when `reverse` is false).

## Related Issues

Fixes #3022

## Screenshots / Recordings

N/A (API behavior fix)

## Steps to QA

```
uv run pytest tests/unit/test_cases_service.py tests/unit/test_tables_service.py tests/unit/test_case_table_rows_service.py
```

The new regression tests page three deep and then walk back — the existing reverse tests only went back from page 2, where "previous page" and "page 1" coincide, which is why the bug never surfaced:

- `test_search_cases_reverse_pagination_returns_adjacent_page`: back from page 3 returns page 2 (previously page 1), flags are oriented in display direction, and the swapped cursors round-trip forward to page 3 and back to page 1.
- `test_list_rows_reverse_pagination_returns_adjacent_page` (tables): same walk over 7 rows with `limit=2`.
- `test_list_rows_reverse_pagination_cursors_point_forward_and_back` (case table rows): after going back a page, `next_cursor` returns to page 3 and `prev_cursor` continues to page 1 (previously each pointed the other way).

All pre-existing pagination tests in the three files pass unchanged.

---
## Summary by cubic
Fixes reverse cursor pagination so “previous page” returns the adjacent page with correct cursors and flags across `CasesService.search_cases`, `TablesService.list_rows`, and `CaseTableRowsService.list_rows`. Forward pagination is unchanged.

- **Bug Fixes**
  - Implemented the canonical reverse flow: flip scan direction, fetch `limit + 1`, compute cursors from scan order, then reverse items and swap `next/prev` and `has_more/has_previous`. Added regression tests that walk back from page 3 to page 2.
  - `CasesService.search_cases` and `TablesService.list_rows`: now flip `ORDER BY` for reverse scans; case search also fixes swapped/overwritten cursors and flags.
  - `CaseTableRowsService.list_rows`: replaces (not appends) the base `ORDER BY` when reversing; computes cursors before reversal and swaps correctly so `next` points forward and `prev` points back.

<sup>Written for commit 8613676562ee8ce66a77613c99bee6952d4a2ec1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3023?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/unit/test_case_table_rows_service.py
+++ b/tests/unit/test_case_table_rows_service.py
@@ -312,3 +312,76 @@ async def test_list_rows_cursor_uses_created_at_and_id_order(
         include_row_data=False,
     )
     assert [item.id for item in legacy_page.items] == [older_small_id, oldest_small_id]
+
+
+@pytest.mark.anyio
+async def test_list_rows_reverse_pagination_cursors_point_forward_and_back(
+    session: AsyncSession,
+    cases_service: CasesService,
+    case_rows_service: CaseTableRowsService,
+    tables_service: TablesService,
+) -> None:
+    """After paging backward, next/prev cursors must keep their meaning.
+
+    Regression test: cursors were computed from the already-reversed items and
+    then swapped again, so after one backward page "next" pointed at the
+    previous page and "prev" pointed at the next page.
+    """
+    case = await _create_case(cases_service)
+    table_id, _ = await _create_table_with_row(
+        tables_service,
+        name=f"case_rows_reverse_{uuid.uuid4().hex[:8]}",
+        value="seed",
+    )
+
+    base_time = datetime(2026, 1, 1, tzinfo=UTC)
+    links = [
+        CaseTableRow(
+            workspace_id=case_rows_service.workspace_id,
+            case_id=case.id,
+            table_id=table_id,
+            row_id=uuid.uuid4(),
+            created_at=base_time + timedelta(minutes=i),
+            updated_at=base_time + timedelta(minutes=i),
+        )
+        for i in range(6)
+    ]
+    session.add_all(links)
+    await session.commit()
+
+    # Display order is created_at desc: newest link first
+    display_ids = [
+        link.id for link in sorted(links, key=lambda x: x.created_at, reverse=True)
+    ]
+
+    async def page(cursor: str | None = None, reverse: bool = False):
+        return await case_rows_service.list_rows(
+            case_id=case.id,
+            limit=2,
+            cursor=cursor,
+            reverse=reverse,
+            include_row_data=False,
+        )
+
+    page1 = await page()
+    assert [item.id for item in page1.items] == display_ids[0:2]
+    page2 = await page(cursor=page1.next_cursor)
+    assert [item.id for item in page2.items] == display_ids[2:4]
+    page3 = await page(cursor=page2.next_cursor)
+    assert [item.id for item in page3.items] == display_ids[4:6]
+    assert page3.prev_cursor is not None
+
+    # Going back from page 3 must return page 2
+    back = await page(cursor=page3.prev_cursor, reverse=True)
+    assert [item.id for item in back.items] == display_ids[2:4]
+    assert back.has_more is True
+    assert back.has_previous is True
+
+    # next continues forward to page 3, prev continues backward to page 1
+    forward_again = await page(cursor=back.next_cursor)
+    assert [item.id for item in forward_again.items] == display_ids[4:6]
+
+    back_to_first = await page(cursor=back.prev_cursor, reverse=True)
+    assert [item.id for item in back_to_first.items] == display_ids[0:2]
+    assert back_to_first.has_more is True
+    assert back_to_first.has_previous is False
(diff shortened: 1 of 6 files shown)


topher-lo:
@cubic review

PR #3077, opened 2026-07-16
test: simplify registry import reload coverage

daryllimyt:
## Summary

- remove the permanently skipped reproduction of the historical import race
- remove the stress test that imported and evicted real `tracecat_registry` modules
- replace the remaining race harness with focused tests alongside `import_and_reload`
- verify both reload-failure preservation and process-wide reload serialization

## Why

The deleted `test_import_reload_no_race_with_lock` did not actually prove that the process-wide lock was effective. It still passed when `_import_reload_lock` was replaced with `contextlib.nullcontext()`, so removing the lock would not have caused the regression test to fail.

The file also contained:

- a permanently skipped test of deliberately unsafe historical behavior
- a pop-window test that waited two seconds for an event that should never occur
- repeated mutation of real `tracecat_registry` entries in `sys.modules`, adding broad state and timing dependence to a unit test

Registry sync is now subprocess-isolated, but `import_and_reload` is still used by local-repository and recovery paths. This PR therefore keeps the relevant contracts and tests them directly:

1. A reload failure must return and preserve the existing module without attempting a fresh import.
2. Two concurrent reload calls must not enter `importlib.reload` simultaneously.

The new serialization test fails when the lock is disabled, unlike the old stress test. In the local timing run, the removed tests accounted for about 4.8 seconds of test-call time; the new concurrency test completes in about 0.1 seconds.

## Validation

- `uv run pytest tests/unit/test_registry_repository.py -q --durations=10`
- `uv run ruff check tests/unit/test_registry_repository.py`
- `uv run ruff format --check tests/unit/test_registry_repository.py`
- `uv run basedpyright tests/unit/test_registry_repository.py`
- mutation check: confirmed the new serialization test fails when `_import_reload_lock` is disabled

---
## Summary by cubic
Simplifies registry import/reload tests by removing the flaky race harness and adding fast, focused tests that verify reload failure handling and process-wide serialization.

- **Refactors**
  - Removed `tests/unit/test_registry_import_race.py` and its flaky race/stress cases against `tracecat_registry`.
  - Added two unit tests alongside `import_and_reload`:
    - Preserves the existing module on reload failure (no fresh `import_module`).
    - Serializes concurrent reloads so `importlib.reload` never runs in parallel.
  - New serialization test fails when the process-wide lock is disabled, and reduces test time from ~4.8s to ~0.1s.

<sup>Written for commit 076a7144fc6df6bdfdd7c89454449706cecbe21f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3077?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


daryllimyt:
@codex review

PR #3082, opened 2026-07-17
feat(audit): expand audit log coverage

jordan-umusu:
## Summary by cubic
Expanded audit logging with task‑per‑event async delivery, bounded retries, typed callbacks, broader control‑plane coverage, stricter privacy, and normalized request attribution (including MCP tool calls). Also optimized audit settings caching and kept batch‑case audit metadata accurate.

- **New Features**
  - Async delivery: task‑per‑event posting with transient retries (exponential backoff), burst queue decoupled from socket cap; `flush_audit_deliveries` for tests.
  - Typed `AuditEventDetails` and privacy‑bounded metadata via `sanitize_audit_metadata` and shared sanitizers (`redact_sensitive_text`, `sanitize_urls_in_text`); `changed_fields`/`changed_sections`; batch case audits use `is_batch`, `case_count`, `succeeded_count`, `failed_count`; dry‑run/deferred commits suppress emission.
  - Coverage: workflow edits/YAML/graph ops, schedules (create/update/delete), case triggers (upsert/update), webhooks (create/update/API key rotate), executions (cancel/terminate/reset); audit payloads exclude `content` and `workflow_alias`.
  - Attribution: request middleware sets `ctx_request_audit` with normalized `client_ip` and bounded `user_agent` family; MCP tool calls attributed via `MCPRequestAuditMiddleware`.

- **Refactors**
  - Centralized sanitizers in `tracecat.sanitization`; removed ad‑hoc regex; MCP stdio probe and integrations now use shared sanitizers.
  - Expanded audit types (`AuditMetadata`, new actions/resources); centralized emission; moved webhook create/update to `tracecat.webhooks.service` with scope enforcement; Temporal RPC error handling moved into services; routers simplified.
  - Cached audit settings reads with `async_lru` to reduce repeated DB lookups and fixed the cache key by dropping the nil‑UUID sentinel.

<sup>Written for commit 9f2c75b1a93673bffa942f577cd11fea51790822. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3082?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tracecat/audit/logger.py
+++ b/tracecat/audit/logger.py
@@ -5,7 +5,7 @@ import inspect
 import uuid
 from collections.abc import Awaitable, Callable, Mapping
 from dataclasses import dataclass
-from typing import Any, cast
+from typing import Any, Concatenate, cast
 
 from sqlalchemy.ext.asyncio import AsyncSession
 
@@ -21,26 +21,6 @@ from tracecat.contexts import ctx_role
 from tracecat.logger import logger
 from tracecat.sanitization import redact_sensitive_text
 from tracecat.service import BaseService
-
-
-@dataclass(frozen=True)
-class AuditCallContext:
-    """Information passed to an audit detail callback.
-
-    Attributes:
-        target: The service associated with the call, when one can be found.
-            For methods this is normally ``self``. For decorated free functions,
-            it may be the bound ``service`` argument. Otherwise, it is the first
-            positional argument or ``None``.
-        arguments: Arguments supplied to the decorated function, keyed by their
-            parameter names. Default values that the caller omitted are not
-            included. These arguments may contain sensitive application data;
-            callbacks must select only identifiers, changed-field names, and
-            other metadata permitted by the audit policy.
-    """
-
-    target: Any | None
-    arguments: Mapping[str, Any]
 
 
 @dataclass(frozen=True)
@@ -76,12 +56,9 @@ class AuditEventDetails:
     emit: bool = True
 
 
-type AuditAttemptMetadataFn = Callable[
-    [AuditCallContext], AuditEventDetails | Awaitable[AuditEventDetails]
-]
-type AuditTerminalMetadataFn[T] = Callable[
-    [AuditCallContext, T], AuditEventDetails | Awaitable[AuditEventDetails]
-]
+type AuditDetailsResult = AuditEventDetails | Awaitable[AuditEventDetails]
+type AuditAttemptMetadataFn[**P] = Callable[P, AuditDetailsResult]
+type AuditTerminalMetadataFn[R, **P] = Callable[Concatenate[R, P], AuditDetailsResult]
 
 # Mapping of resource types to their default resource_id_attr names
 _RESOURCE_ID_ATTR_MAP: dict[str, str] = {
@@ -106,8 +83,8 @@ def audit_log[**P, R](
     resource_type: AuditResourceType,
     action: AuditAction,
     resource_id_attr: str | None = None,
-    attempt_metadata: AuditAttemptMetadataFn | None = None,
-    terminal_metadata: AuditTerminalMetadataFn[R] | None = None,
+    attempt_metadata: AuditAttemptMetadataFn[P] | None = None,
+    terminal_metadata: AuditTerminalMetadataFn[R, P] | None = None,
 ) -> Callable[
     [Callable[P, Awaitable[R]]],
     Callable[P, Awaitable[R]],
@@ -123,8 +100,9 @@ def audit_log[**P, R](
 
     The actor role is resolved from the service instance or a bound ``role``
     argument, then from the role context. Auditing is skipped when there is no
-    authenticated actor or when the resolved role belongs to an internal
-    service. User and service-account roles remain auditable. Event creation and
+    auditable actor. Internal service roles that carry the initiating user's ID
+    are audited and attributed to that user; unattributed service traffic has
+    no actor and is skipped. Event creation and
     detail callbacks are best-effort: their failures are logged without changing
     the result of the decorated operation. A service session is reused when
     available; decorated free functions can instead supply a bound ``session``
@@ -150,17 +128,18 @@ def audit_log[**P, R](
             the call, extraction checks positional objects and bound arguments;
             after the call, the return value is checked first.
         attempt_metadata: Optional synchronous or asynchronous callback invoked
-            before the attempt event. It receives an
-            :class:`AuditCallContext` and returns :class:`AuditEventDetails`.
-            Only ``action``, ``resource_id``, ``data``, and ``emit`` are used in
-            this phase. If it raises, the decorator logs the error and continues
-            with its default fields.
-        terminal_metadata: Optional synchronous or asynchronous callback
-            invoked after the decorated function returns. It receives the same
-            call context plus the function result and returns
-            :class:`AuditEventDetails`. Only ``resource_id``, ``data``, and
-            ``status`` are used in this phase. If it raises, the terminal event
-            uses the fields already derived.
+            before the attempt event. It takes the decorated function's own
+            arguments and returns :class:`AuditEventDetails`. Only ``action``,
+            ``resource_id``, ``data``, and ``emit`` are used in this phase. If it
+            raises, the decorator logs the error and continues with its default
+            fields.
+        terminal_metadata: Optional synchronous or asynchronous callback invoked
+            after the decorated function returns. It takes the result followed by
+            the decorated function's arguments; method-attached callbacks should
+            declare ``(result, *args: Any, **kwargs: Any)`` because a named
+            parameter spelling fails pyright's name check against P. Only
+            ``resource_id``, ``data``, and ``status`` are used in this phase. If
+            it raises, the terminal event uses the fields already derived.
 
     Returns:
         A decorator that preserves the wrapped function's signature and return
@@ -202,17 +181,13 @@ def audit_log[**P, R](
                 else ctx_role.get()
             )
 
-            if role is None or role.actor_id is None or role.type == "service":
+            if role is None or role.actor_id is None:
                 return await func(*args, **kwargs)
 
             raw_session = getattr(service, "session", None)
             if raw_session is None:
                 raw_session = bound_arguments.get("session")
             session = cast(AsyncSession | None, raw_session)
-            call_context = AuditCallContext(
-                target=service if service is not None else first_arg,
-                arguments=bound_arguments,
-            )
 
             resource_id: uuid.UUID | None = None
             try:
@@ -234,7 +209,7 @@ def audit_log[**P, R](
             should_emit = True
             if attempt_metadata is not None:
                 try:
-                    details = attempt_metadata(call_context)
+                    details = attempt_metadata(*args, **kwargs)
                     if inspect.isawaitable(details):
                         details = await details
                     if details.action is not None:
@@ -298,7 +273,7 @@ def audit_log[**P, R](
                     )
                     if terminal_metadata is not None:
                         try:
-                            details = terminal_metadata(call_context, result)
+                            details = terminal_metadata(result, *args, **kwargs)
                             if inspect.isawaitable(details):
                                 details = await details
                             if details.resource_id is not None:
--- a/tracecat/audit/logger.py
+++ b/tracecat/audit/logger.py
@@ -4,23 +4,90 @@ import functools
 import inspect
 import uuid
 from collections.abc import Awaitable, Callable, Mapping
-from typing import Any, Concatenate, ParamSpec, TypeVar
+from dataclasses import dataclass
+from typing import Any, cast
+
+from sqlalchemy.ext.asyncio import AsyncSession
 
 from tracecat.audit.enums import AuditEventStatus
 from tracecat.audit.service import AuditService
-from tracecat.audit.types import AuditAction, AuditResourceType
+from tracecat.audit.types import (
+    AuditAction,
+    AuditMetadata,
+    AuditResourceType,
+)
 from tracecat.auth.types import PlatformRole, Role
 from tracecat.contexts import ctx_role
 from tracecat.logger import logger
-from tracecat.service import BasePlatformService, BaseService
+from tracecat.sanitization import redact_sensitive_text
+from tracecat.service import BaseService
 
-P = ParamSpec("P")
-R = TypeVar("R")
+
+@dataclass(frozen=True)
+class AuditCallContext:
+    """Information passed to an audit detail callback.
+
+    Attributes:
+        target: The service associated with the call, when one can be found.
+            For methods this is normally ``self``. For decorated free functions,
+            it may be the bound ``service`` argument. Otherwise, it is the first
+            positional argument or ``None``.
+        arguments: Arguments supplied to the decorated function, keyed by their
+            parameter names. Default values that the caller omitted are not
+            included. These arguments may contain sensitive application data;
+            callbacks must select only identifiers, changed-field names, and
+            other metadata permitted by the audit policy.
+    """
+
+    target: Any | None
+    arguments: Mapping[str, Any]
+
+
+@dataclass(frozen=True)
+class AuditEventDetails:
+    """Audit fields returned by an audit metadata callback.
+
+    Attributes:
+        action: Replaces the decorator's default action when returned by
+            ``attempt_metadata``. It is ignored when returned by
+            ``terminal_metadata``.
+        resource_id: Replaces the automatically extracted resource ID. A value
+            from ``attempt_metadata`` applies to every event; a value from
+            ``terminal_metadata``
+            applies only to the terminal event.
+        data: Metadata to store on the event. Data from ``attempt_metadata`` is
+            used for both the attempt and terminal events. Non-``None`` data
+            from ``terminal_metadata`` replaces it for the terminal event only.
+            The audit service applies its metadata allowlist before delivery;
+            callers must still avoid constructing metadata from raw content or
+            secret values.
+        status: Replaces the inferred terminal status when returned by
+            ``terminal_metadata``. It is ignored when returned by
+            ``attempt_metadata``.
+        emit: When ``False`` from ``attempt_metadata``, executes the decorated
+            function without emitting any audit events. It is ignored by
+            ``terminal_metadata``.
+    """
+
+    action: AuditAction | None = None
+    resource_id: uuid.UUID | None = None
+    data: AuditMetadata | None = None
+    status: AuditEventStatus | None = None
+    emit: bool = True
+
+
+type AuditAttemptMetadataFn = Callable[
+    [AuditCallContext], AuditEventDetails | Awaitable[AuditEventDetails]
+]
+type AuditTerminalMetadataFn[T] = Callable[
+    [AuditCallContext, T], AuditEventDetails | Awaitable[AuditEventDetails]
+]
 
 # Mapping of resource types to their default resource_id_attr names
 _RESOURCE_ID_ATTR_MAP: dict[str, str] = {
     "workflow": "workflow_id",
     "workflow_execution": "wf_id",
+    "schedule": "schedule_id",
     "organization_member": "user_id",
     "organization_session": "session_id",
     "organization": "org_id",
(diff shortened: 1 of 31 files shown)


tracecat/audit/logger.py
```diff
@@ -63,23 +179,40 @@ def decorator(
...
             role: Role | PlatformRole | None = (
-                self.role if isinstance(self, BasePlatformService) else ctx_role.get()
+                candidate
+                if isinstance(candidate, (Role, PlatformRole))
+                else ctx_role.get()
             )
 
-            if role is None or role.actor_id is None:
-                return await func(self, *args, **kwargs)
+            if role is None or role.actor_id is None or role.type == "service":
```
daryllimyt: why do we start skipping 'service' Roles?

tracecat/workflow/executions/service.py
```diff
@@ -1937,6 +1960,11 @@ async def _resolve_reset_event_id(
             )
         return reset_event_id
 
+    @audit_log(
+        resource_type="workflow_execution",
+        action="create",
```
daryllimyt: should this be 'reset'?

tracecat/workflow/management/management.py
```diff
@@ -96,6 +96,17 @@
 from tracecat.workflow.schedules.service import WorkflowSchedulesService
 
 
+def _workflow_update_audit_details(
+    context: AuditCallContext,
+) -> AuditEventDetails:
+    params = cast(WorkflowUpdate, context.arguments["params"])
+    return AuditEventDetails(
+        data={
+            "changed_fields": sorted(params.model_dump(exclude_unset=True)),
```
daryllimyt: `model_dump(exclude_unset=True)` serializes every set value (including potentially large ones like `expects`/`config`/`returns`) only to discard them and keep the keys. `model_fields_set` yields the same field names without the serialization — `WorkflowUpdate` is a plain `BaseModel` (no aliases, computed fields, or `extra` config), so this is a drop-in equivalent.

```suggestion
            "changed_fields": sorted(params.model_fields_set),
```

tracecat/webhooks/service.py
```diff
@@ -1,18 +1,90 @@
...
+    if role.workspace_id is not None:
+        webhook = await get_webhook(
+            session,
+            workspace_id=role.workspace_id,
+            workflow_id=workflow_id,
+        )
+    return AuditEventDetails(
+        resource_id=webhook.id if webhook is not None else None,
+        data={
+            "changed_fields": sorted(params.model_dump(exclude_unset=True)),
```
daryllimyt: Same as in `management.py`: `model_fields_set` gives the same keys without serializing the values (`WebhookUpdate` is a plain `BaseModel`).

```suggestion
            "changed_fields": sorted(params.model_fields_set),
```

tracecat/workflow/case_triggers/service.py
```diff
@@ -15,6 +17,40 @@
...
+
+
+def _case_trigger_update_audit_details(
+    context: AuditCallContext,
+) -> AuditEventDetails:
+    if context.arguments.get("commit", True) is False:
+        return AuditEventDetails(emit=False)
+    params = cast(CaseTriggerUpdate, context.arguments["params"])
+    return AuditEventDetails(
+        data={"changed_fields": sorted(params.model_dump(exclude_unset=True))}
```
daryllimyt: Same as in `management.py`: `model_fields_set` gives the same keys without serializing the values (`CaseTriggerUpdate` is a plain `BaseModel`).

```suggestion
        data={"changed_fields": sorted(params.model_fields_set)}
```

tracecat/workflow/case_triggers/service.py
```diff
@@ -131,6 +167,12 @@ async def get_case_trigger(self, workflow_id: WorkflowID) -> CaseTrigger:
         return await self._ensure_case_trigger_exists(workflow_id)
 
     @requires_entitlement(Entitlement.CASE_ADDONS)
+    @audit_log(
+        resource_type="case_trigger",
+        action="create",
```
daryllimyt: should this be 'upsert'

tracecat/workflow/executions/service.py
```diff
@@ -2003,14 +2031,40 @@ async def _reset(
...
+            if exc.status != RPCStatusCode.NOT_FOUND:
+                raise
+            self.logger.info(
+                "Workflow execution already completed, ignoring cancellation request",
+                wf_exec_id=wf_exec_id,
+            )
 
+    @audit_log(
+        resource_type="workflow_execution",
+        action="cancel",
```
daryllimyt: `terminate_workflow_execution` is audited with `action="cancel"` — only `data.operation` distinguishes it from a real cancel. Consumers filtering audit events by `action` will conflate graceful cancels with forceful terminations.

Since this PR already extends the `AuditAction` literal with `publish`/`cancel`/`rotate`, adding `terminate` and using it here is a one-line wire-contract change that's much cheaper now than after sinks start ingesting events. (Reset auditing as `create` + `operation="reset"` is fine to keep — a reset genuinely creates a new run — but cancel vs terminate is a real semantic distinction for security review.)

If keeping `cancel` is deliberate, a short comment on the decorator would prevent this from being "fixed" later into a breaking change.

tracecat/audit/service.py
```diff
@@ -213,8 +219,8 @@ async def _post_event(self, *, webhook_url: str, payload: AuditEvent) -> None:
         except Exception as exc:
             self.logger.warning(
                 "Failed to deliver audit webhook",
-                error=str(exc),
-                webhook_url=webhook_url,
+                error=redact_sensitive_text(str(exc), redact_emails=True),
+                webhook_url=redact_sensitive_text(webhook_url),
```
daryllimyt: [P1] Do not include the webhook URL path in this log field.

redact_sensitive_text removes URL userinfo, query strings, and fragments, but deliberately preserves the path. Many webhook providers encode the credential in that path, so the current failure log can expose a usable secret to anyone with application-log access.

Please log only a parsed origin or hostname, or omit webhook_url entirely.

tracecat/workflow/schedules/service.py
```diff
@@ -67,6 +69,11 @@ async def list_schedules(
         return list(schedules)
 
     @require_scope("schedule:create")
+    @audit_log(
```
daryllimyt, line 92: [P1] This decorator can emit SUCCESS before the transaction actually commits.

create_schedule accepts commit=False, and import and service-composition paths call it that way. The wrapper emits terminal success as soon as the method returns after flush; if the outer transaction later rolls back, the external audit stream permanently reports a schedule creation that never happened. delete_schedule has the same issue.

Please emit these events only after the owning transaction commits, or suppress the per-schedule audit for commit=False and let the transaction owner audit the mutation.

tracecat/audit/logger.py
```diff
@@ -91,61 +224,130 @@ async def wrapper(self: Any, *args: P.args, **kwargs: P.kwargs) -> R:
...
+                        "Audit attempt metadata derivation failed",
+                        resource_type=resource_type,
+                        action=action,
+                        error_type=type(exc).__name__,
+                    )
+
+            if not should_emit:
+                return await func(*args, **kwargs)
 
             async with AuditService.with_session(role, session=session) as svc:
```
daryllimyt, line 232: [P2] Audit webhook delivery synchronously blocks every emitting audited operation.

The wrapper awaits an ATTEMPT delivery before the mutation and a SUCCESS or FAILURE delivery afterward. Each delivery can spend 10 seconds in _post_event. This PR adds 20 production decorators across graph and workflow edits, schedules, execution controls, webhook and API-key operations, case triggers, and MCP updates. A degraded sink can therefore add roughly 20 seconds to all these request paths; version-conflict retries can repeat the delay.

Please persist events to a durable outbox and deliver them asynchronously so an external audit sink cannot determine control-plane request latency.

tracecat/audit/logger.py
```diff
@@ -4,23 +4,90 @@
...
+            positional argument or ``None``.
+        arguments: Arguments supplied to the decorated function, keyed by their
+            parameter names. Default values that the caller omitted are not
+            included. These arguments may contain sensitive application data;
+            callbacks must select only identifiers, changed-field names, and
+            other metadata permitted by the audit policy.
+    """
+
+    target: Any | None
+    arguments: Mapping[str, Any]
```
daryllimyt: [P3] Consider making audit metadata callbacks type-safe by forwarding the decorated function ParamSpec instead of exposing Mapping[str, Any].

**Before:**
```python
def details(context: AuditCallContext) -> AuditEventDetails:
    params = cast(WorkflowUpdate, context.arguments["params"])
```
A parameter rename still typechecks. At runtime the stale key raises KeyError, the wrapper catches it, and the audit event falls back to its default fields.

**After:**
```python
type AuditDetailsResult = AuditEventDetails | Awaitable[AuditEventDetails]

def audit_log[**P, R](
    *,
    attempt_metadata: Callable[P, AuditDetailsResult] | None = None,
    terminal_metadata: Callable[Concatenate[R, P], AuditDetailsResult] | None = None,
): ...

details = attempt_metadata(*args, **kwargs)
details = terminal_metadata(result, *args, **kwargs)
```
Callbacks then match the decorated signature directly:
```python
def details(
    service: WorkflowsManagementService,
    workflow_id: WorkflowID,
    params: WorkflowUpdate,
) -> AuditEventDetails: ...
```
This lets Pyright catch incompatible signature changes and removes the current inspect.signature -> Mapping[str, Any] -> string lookup -> cast chain across nine files. This is non-blocking because migrating all callback attachments is broader than the coverage change, but it would be a safer direction for this API.

tracecat/audit/service.py
daryllimyt: alternatively we could make redact_sensitive_text() specifically handle webhook urls

tracecat/middleware/request.py
```diff
@@ -1,26 +1,63 @@
...
-        forwarded_for = request.headers.get("X-Forwarded-For")
-        if forwarded_for:
-            # X-Forwarded-For format: "client, proxy1, proxy2"
-            # First IP is the original client
-            client_ip = forwarded_for.split(",")[0].strip()
-        # Fallback to direct connection IP
-        if not client_ip and request.client:
-            client_ip = request.client.host
+        if forwarded_for := request.headers.get("X-Forwarded-For"):
+            client_ip = _normalize_client_ip(forwarded_for.split(",")[0].strip())
```
daryllimyt: since we only take the first item we can split once
```suggestion
            client_ip = _normalize_client_ip(forwarded_for.split(",", 1)[0].strip())
```

tracecat/middleware/request.py
```diff
@@ -1,26 +1,63 @@
...
+        if client_ip is None:
+            client_ip = _normalize_client_ip(
+                request.client.host if request.client is not None else None
+            )
+
+        user_agent = _normalize_audit_user_agent(request.headers.get("User-Agent"))
 
-        token = ctx_client_ip.set(client_ip)
+        client_ip_token = ctx_client_ip.set(client_ip)
+        user_agent_token = ctx_user_agent.set(user_agent)
```
daryllimyt: Consolidate `ctx_client_ip` and `ctx_user_agent` into a single request-context object rather than two disparate contextvars. They share an identical lifecycle (set together at request ingress, reset together in the same `finally`), which signals they're one concept — audit attribution captured at ingress.

Something like:

```python
@dataclass(frozen=True, slots=True)
class RequestAuditContext:
    client_ip: str | None
    user_agent: str | None

ctx_request_audit: ContextVar[RequestAuditContext | None] = ContextVar(
    "request-audit", default=None
)
```

This gives one set/reset pair (no token-ordering bookkeeping here), an atomic snapshot for the audit service instead of two independent reads, and a natural home for future attribution fields (e.g. request ID) without growing the flat registry in `contexts.py`. Blast radius is small — only this middleware and `audit/service.py` touch these vars.

tracecat/workflow/executions/internal_router.py
```diff
@@ -645,23 +645,17 @@ async def update_webhook(
     Mirrors the public ``PATCH /workflows/{id}/webhook`` route so the workflows
     SDK (which resolves under ``/internal``) can enable/disable webhooks.
     """
-    if role.workspace_id is None:
-        raise HTTPException(
-            status_code=HTTP_400_BAD_REQUEST, detail="Workspace ID is required"
+    try:
```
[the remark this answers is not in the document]
daryllimyt: why was this marked resolved?

daryllimyt (changes requested):

tracecat/audit/logger.py
daryllimyt: for now - use asyncio tassks + Set[asayncio.Task] in the module to hold strong refs to the emission event proc wide, then add done_callback to pop from the set. i think this pattern exists elsewhere

tracecat/audit/logger.py
jordan-umusu: it was meant to skip true machine traffic but might've been too broad and included user-attributed executor calls, narrowed.

tracecat/audit/logger.py
jordan-umusu: agree with the implications of expanding coverage, think though given the extra day to work on this, persisting in postgres and batch delivering in temporal is probably most sound.

jordan-umusu: proceeded with the asyncio tasks set approach.

tracecat/audit/service.py
```diff
@@ -12,19 +21,109 @@
...
+        logger.warning(
+            "Failed to deliver audit webhook",
+            error_type=type(exc).__name__,
+            status_code=getattr(response, "status_code", None),
+            resource_type=delivery.resource_type,
+            action=delivery.action,
+        )
+
+
+async def flush_audit_deliveries() -> None:
```
daryllimyt: move into tests module

tracecat/audit/service.py
```diff
@@ -179,44 +276,57 @@ async def _get_payload_attribute(self) -> str | None:
...
                 "audit_webhook_payload_attribute must be a string",
-                value=value,
                 value_type=type(value),
             )
             return None
 
         cleaned = value.strip()
         return cleaned or None
 
+    async def _build_delivery(
```
daryllimyt, line 386: we'd likely want to cache the configs here so we avoid overloading the DB. possible candidates `_get_audit_setting`

tracecat/audit/service.py
```diff
@@ -292,13 +408,47 @@ async def create_event(
...
+                user-agent from context. This field is sensitive security
+                metadata and may identify client software or devices.
+        """
+
         if self.role is None or getattr(self.role, "actor_id", None) is None:
             self.logger.debug(
-                "Skipping audit log", reason="non_auditable_role", role=self.role
+                "Skipping audit log",
+                reason="non_auditable_role",
+                role_type=getattr(self.role, "type", None),
```
daryllimyt: nit: replace getattr

tracecat/workflow/management/draft.py
```diff
@@ -844,19 +845,47 @@ def parse_workflow_edit_request(
     return request
 
 
+def _workflow_edit_audit_details(
+    *,
+    role: Any,
```
daryllimyt: can this be typed?

tracecat/audit/service.py
```diff
@@ -12,19 +21,109 @@
...
+    action: AuditAction
+
+
+# Strong refs to in-flight delivery tasks; done callbacks release them.
+_delivery_tasks: set[asyncio.Task[None]] = set()
+
+# Each in-flight delivery holds one socket/FD. Sized to <=25% of the smallest
+# deployment FD envelope (Fargate default soft ulimit 1024); drains ~25/s even
+# against a hung sink (10s httpx timeout). Recompute if either input changes.
+_MAX_IN_FLIGHT_DELIVERIES = 256
```
daryllimyt: **Follow-up idea: decouple the burst cap from the socket cap with a per-loop semaphore.**

`_MAX_IN_FLIGHT_DELIVERIES = 256` currently serves two roles at once:

1. **FD/socket cap** — each in-flight `_deliver` holds one httpx connection.
2. **Burst admission cap** — event #257 in a burst is dropped *even when the sink is fast and healthy*, because tasks can't complete as fast as they're spawned.

These want different values: the socket cap should be small (a few dozen), the burst cap much larger (payloads are KB-sized identifier dicts, so buffering is cheap).

Sketch — keep the task-per-event structure and gate only the HTTP call:

```python
# Pending-task bound (burst absorption; drop threshold, replaces 256).
_MAX_PENDING_DELIVERIES = 2048
# Concurrent-socket bound (FD budget).
_MAX_CONCURRENT_POSTS = 32

# asyncio primitives bind to a loop; keyed per loop and evicted alongside
# the closed-loop task sweep in _spawn_delivery.
_semaphores: dict[asyncio.AbstractEventLoop, asyncio.Semaphore] = {}

def _get_loop_semaphore() -> asyncio.Semaphore:
    loop = asyncio.get_running_loop()
    if (sem := _semaphores.get(loop)) is None:
        sem = _semaphores[loop] = asyncio.Semaphore(_MAX_CONCURRENT_POSTS)
    return sem

async def _deliver(delivery: _AuditDelivery) -> None:
    async with _get_loop_semaphore():
        ...  # existing httpx post, unchanged
```

Pending tasks queue themselves on the semaphore's FIFO waiter list, so `_delivery_tasks`, the done-callback discard, the closed-loop sweep, and `flush_audit_deliveries` all keep working unchanged — no worker pool or dispatcher task to manage.

Sizing rationale (2048/32): a full backlog against a *healthy* sink drains in seconds; against a hung sink (10s timeout → 3.2/s) it drains in ~10 min, which keeps worst-case event staleness at the SIEM acceptable. Memory cost of a full backlog is ~2–4 MB of suspended tasks. Overhead per audited request is one eager `create_task` (~µs), noise next to the settings lookups already on this path.

Still at-most-once — queue-full drops with the same warning log. Durable delivery remains ENG-1514's spool; this only stops us shedding load during ordinary bursts.

daryllimyt (changes requested):

tracecat/audit/service.py
jordan-umusu: agreed, caching added for the getters within `_build_delivery`

tracecat/audit/service.py
```diff
@@ -1,30 +1,243 @@
...
+            status_code=response.status_code if response is not None else None,
+            resource_type=delivery.resource_type,
+            action=delivery.action,
+            attempts=attempts,
+        )
+
+
+# Sentinel org id for the platform sink and for org-sink lookups with no org
+# identity; keeps those entries isolated from any real organization's cache key.
+_PLATFORM_ORG_SENTINEL = uuid.UUID(int=0)
```
daryllimyt: **nonblocking:** Using `UUID(int=0)` as the "no org identity" sentinel is a bit of a smell in this codebase specifically:

1. **Overloaded meaning.** The nil UUID historically was the *real* default-organization id here (seeded in `4a298374b126`, scrubbed out of production DBs by `4ef4d2e1d57b`). Reusing it to mean "no org" is the opposite semantic, and a reader who knows that history may wonder if this deliberately targets the legacy default org.
2. **Implicit invariant.** "UUID(0) can't collide with a real org" isn't a property of this code — it depends on that data migration having run. A restored pre-migration backup would make the real org's audit-setting lookups silently return `default` instead of its configured values.

Since `sink` is already part of the cache key, the sentinel isn't needed at all — `organization_id: OrganizationID | None` with `None` meaning "no org identity" says the same thing explicitly, and no value can ever collide with `None`:

```python
@alru_cache(ttl=30)
async def _get_audit_setting_cached(
    sink: AuditSink,
    organization_id: OrganizationID | None,
    key: str,
    default: Any = None,
) -> Any | None:
    ...
    if organization_id is None:
        # No org identity: preserve get_setting(role=None) semantics.
        return default
```

and at the call site, canonicalize platform-sink reads to `None`:

```python
organization_id = (
    self.role.organization_id
    if self.audit_sink == "organization" and isinstance(self.role, Role)
    else None
)
```

Small bonus: today a platform-sink read keyed with a caller's org id fragments the platform config into one cache entry per calling org (each with its own TTL + DB read); always passing `None` for the platform sink dedupes those into a single entry. Verified locally that ruff + basedpyright pass with this shape.

tracecat/middleware/request.py
```diff
@@ -1,26 +1,70 @@
...
+
+def build_request_audit_context(request: Request) -> RequestAuditContext:
+    """Derive audit attribution from an HTTP request.
+
+    Leftmost X-Forwarded-For hop, falling back to the socket peer.
+    Client-controlled: informational attribution, not a security control.
+    """
+    client_ip = None
+    if forwarded_for := request.headers.get("X-Forwarded-For"):
+        client_ip = _normalize_client_ip(forwarded_for.split(",", 1)[0].strip())
```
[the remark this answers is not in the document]
jordan-umusu: Keeping current behavior in this PR; filing a follow-up for a proper trusted-proxy chain. Per-deployment facts:

1. Compose: Caddy 2.10.2 overwrites XFF with {remote_host} on /api*, and its post-2.5 defaults drop untrusted inbound XFF on the other routes. Spoofed values don't reach the API through Caddy. Only bypass: dev compose publishes the API port directly.
2. Fargate: unspoofable, but attribution is already lossy — the ALB appends the real client IP, then Caddy overwrites the header with the ALB's private IP. ip_address records the ALB, not the client.
3. Helm/EKS: no shipped proxy config — plain Ingress to the API service. XFF semantics belong to the operator's controller. Your append scenario is real here.

The fix that covers all three is the trusted-proxy chain, not an inline tweak: uvicorn ProxyHeadersMiddleware + FORWARDED_ALLOW_IPS (app reads only request.client.host), Caddy trusted_proxies for the ALB, documented ingress expectations for Helm — plus the same leftmost-XFF reads in rate_limit.py and mcp/oidc/endpoints.py. Until then the field stays documented as informational attribution, not a security control.

daryllimyt (approved):
lgtm!

PR #3098, opened 2026-07-20
fix(redis): retry only transient errors and make stream TTL opt-in

daryllimyt:
## Summary

Hardens `RedisClient` so application-level Redis errors stop being treated as transport failures, and removes a chat-specific TTL default from the generic `xadd` API.

### Retry/reset policy (behavioral fix)

`ResponseError` is a subclass of `RedisError`, so every non-transient, application-level error (NOGROUP, BUSYGROUP, malformed commands) previously triggered:

- 3 retries with exponential backoff that could never succeed,
- a `_reset_connection()` that tears down the **shared** connection pool for all users of the singleton mid-flight, and
- a `tenacity.RetryError` wrapper (tenacity defaults to `reraise=False`), forcing callers like `CaseTriggerConsumer._is_nogroup_error` to unwrap `error.last_attempt` to find the real exception.

Now all methods share one policy: retry (and reset the connection) only on `redis.ConnectionError`, `redis.TimeoutError`, and `RuntimeError` (the event-loop-change recovery path in `_get_client` relies on `RuntimeError` retries), with `reraise=True` so callers always see the original exception, never `RetryError`.

Non-transient errors now propagate immediately on the first attempt: no retry, no error log, no pool reset. Existing `except (ResponseError, RetryError)` handlers in consumers remain compatible — the `RetryError` branch simply becomes dead defensive code.

### `xadd` TTL is now opt-in

`xadd` defaulted `expire_seconds` to `REDIS_CHAT_TTL_SECONDS`, so any stream author who forgot `expire_seconds=None` had their stream silently expired on the chat schedule. The default is now `None`; the agent stream connector passes the chat TTL explicitly (byte-for-byte identical behavior). Audit of all `.xadd(` call sites:

- `tracecat/agent/stream/connector.py` — now passes `REDIS_CHAT_TTL_SECONDS` explicitly (all markers route through `append`)
- `tracecat/cases/triggers/publisher.py` — already passed `expire_seconds=None` explicitly; unchanged

### Smaller fixes

- `xpending_range` return type: `list[Any]` → `list[dict[str, Any]]` (what redis-py actually returns with `decode_responses=True`), so callers can stop shape-sniffing entries
- `ping()` simplified; deleted the `_awaitable_to_bool` helper
- `close()` → `aclose()` (redis-py is pinned to 7.1.0 where `close()` is deprecated)

Out of scope, deliberately unchanged: singleton structure, event-loop-change handling in `_get_client`, and the synchronous Secrets Manager URL resolution.

## Testing

- `tests/unit/test_redis_client.py` updated: TTL default tests inverted; new tests assert a `ResponseError` propagates on the first call with no pool reset (and is the same exception instance, not a wrapper), and that exhausted transient retries re-raise the original `ConnectionError` after exactly 3 attempts, not `RetryError`.
- `uv run pytest tests/unit/test_redis_client.py tests/unit/test_case_trigger_consumer.py` — 22 passed (trigger consumer tests unchanged).
- `ruff check`, `ruff format --check`, `basedpyright --warnings` on touched paths — clean.

### DIFF
--- a/tests/unit/test_redis_client.py
+++ b/tests/unit/test_redis_client.py
@@ -6,6 +6,8 @@ from types import SimpleNamespace
 
 import pytest
 import tenacity
+from redis.exceptions import ConnectionError as RedisConnectionError
+from redis.exceptions import ResponseError
 
 from tracecat.redis.client import RedisClient
 
@@ -31,13 +33,15 @@ class DummyRedis:
     expire_call_count = 0
     _xadd_raised_once = False
     raise_on_xadd_once = False
+    xreadgroup_call_count = 0
+    xreadgroup_error: Exception | None = None
 
     def __init__(self, *, connection_pool: DummyConnectionPool) -> None:
         self.connection_pool = connection_pool
         self.closed = False
         DummyRedis.instances.append(self)
 
-    async def close(self) -> None:
+    async def aclose(self) -> None:
         self.closed = True
 
     async def ping(self) -> bool:
@@ -59,6 +63,14 @@ class DummyRedis:
     ) -> list[tuple[str, list[tuple[str, dict[str, str]]]]]:
         return []
 
+    async def xreadgroup(
+        self, *args, **kwargs
+    ) -> list[tuple[str, list[tuple[str, dict[str, str]]]]]:
+        DummyRedis.xreadgroup_call_count += 1
+        if DummyRedis.xreadgroup_error is not None:
+            raise DummyRedis.xreadgroup_error
+        return []
+
     async def xrange(self, *args, **kwargs) -> list[tuple[str, dict[str, str]]]:
         return []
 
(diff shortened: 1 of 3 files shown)


daryllimyt:
* **#2781** <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/2781?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3098** <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/3098?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/3098?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #3101, opened 2026-07-21
feat: add KSM opt-in and gc.freeze to standalone entrypoint

daryllimyt:
## Summary

Stacked on #3099. Implements the ENG-1549 memory items that must ship *before* VPS measurement can happen — KSM only merges pages from processes that explicitly opt in, so the baseline can never show its effect without this code.

- **KSM opt-in** — `TRACECAT__STANDALONE_MEMORY_MERGE` (`env_bool`, default `false`). When enabled on Linux ≥6.4, `tracecat.standalone` calls `prctl(PR_SET_MEMORY_MERGE)` at boot so the kernel can dedupe the ~400MB byte-identical imported module graph across instance containers. Graceful no-op with a single errno-coded warning on non-Linux / old kernel / missing capability.
- **`gc.freeze()`** after Temporal clients connect — moves import-era objects to the permanent generation so the GC stops touching (and dirtying, which would break KSM merges) those pages.
- Compose/env/README plumbing: flag exposed as `${VAR:-false}` in the instance compose, explicit `false` in `instance.env.example`, and a README "KSM" subsection (host prerequisites, verification via `/sys/kernel/mm/ksm/pages_sharing`, judge by PSS not RSS).

Notes:

- The third ENG-1549 candidate (shared SQLAlchemy engine) turned out to already hold: `get_async_engine()` is a module-level singleton (`tracecat/db/engine.py:205`), so the consolidated process shares one engine/pool by construction. No change needed.
- Expected effect (to be measured on the VPS, ENG-1549): marginal per-instance RSS ~620MiB → ~200-250MiB with N≥2 instances. Not measurable locally — OrbStack's kernel ships without `CONFIG_KSM`.

## Verification

- `uv run ruff check .` / `ruff format --check` — clean
- `uv run basedpyright --warnings` on touched Python files — 0 errors, 0 warnings
- `docker compose config -q` on both multi-dev compose files — pass
- `uv run pytest tests/unit/test_config.py` — 27 passed
- Forced-on sanity on macOS: warns "requires Linux; continuing without KSM opt-in" and returns cleanly
- Flag unset ⇒ byte-identical behavior to #3099 (opt-in path not entered; `gc.freeze` is the only unconditional addition)

Ref: ENG-1549

---
## Summary by cubic
Add KSM opt-in and `gc.freeze()` to the dev standalone entrypoint so Linux hosts can deduplicate identical pages and keep them mergeable. Supports ENG-1549 by making KSM savings measurable on the VPS.

- **New Features**
  - `TRACECAT__STANDALONE_MEMORY_MERGE` (default `false`): on Linux ≥6.4 calls `prctl(PR_SET_MEMORY_MERGE)`; otherwise warns and no-ops.
  - Run `gc.freeze()` after Temporal clients connect to keep import-era pages clean for KSM.
  - Compose/env/README updates to expose the flag and document host KSM setup and verification.

- **Migration**
  - Enable KSM on the host: `echo 1 | sudo tee /sys/kernel/mm/ksm/run`
  - Set `TRACECAT__STANDALONE_MEMORY_MERGE=true` for each instance.
  - Verify `/sys/kernel/mm/ksm/pages_sharing` grows; judge savings by PSS, not RSS.

<sup>Written for commit 59968645a86841fb5651377b54a769fbb5b9528a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


daryllimyt:
Superseded: replacing kernel-dependent KSM dedupe with deterministic fork-CoW (zygote) stacked on #3099 — works on any kernel and is measurable locally. gc.freeze returns as part of the zygote parent.

PR #1749, opened 2025-12-02
fix: move table sorting to server side

jordan-umusu:
## Summary
Adds server-side sorting support for the cases table, enabling sorting across all pages instead of just the current page's data.

## Description
- Add orderBy and sort query parameters to the cases list API endpoint
- Support sorting by created_at, updated_at, priority, severity, status, and tasks columns
- Add task count sorting via correlated subquery in the backend

---
## Summary by cubic
Moved table sorting to the server so sorting works across all pages with cursor pagination. Adds sorting to the Cases and Tables list endpoints.

- **Bug Fixes**
  - Server-side sorting for cases: created_at, updated_at, priority, severity, status, tasks (tasks via correlated subquery); also supports id and short_id (mapped to case_number).
  - New order_by and sort params in casesList and tablesListRows APIs; SDK types updated; invalid order_by returns 400.
  - DataTable enables manualSorting and emits sorting changes; CaseTable wires sorting into pagination hook.
  - Pagination hook tracks sorting, resets cursors on change, and includes orderBy/sort in queries.
  - Cursor pagination is sort-column-aware: encodes sort column/value, honors asc/desc and reverse, and uses id tie-breaker for stable cross-page sorting.

<sup>Written for commit 6d6fa5245348d268f166b23ea1262b173bbd6f5a. Summary will update automatically on new commits.</sup>

PR #3131, opened 2026-07-26
fix(api): fix reverse paging and reject stale-sort cursors

mcm:
## Problem

Reverse cursor pagination (`reverse=true`) returns the wrong page.

`prev_cursor` itself is computed correctly — it encodes the first row of the current page. The path that consumes it is what's broken. In `CasesService.search_cases`, the reverse branch inverted the cursor `WHERE` clause but never inverted the `ORDER BY`: the sort branched only on `sort_direction`, and the rows were never reversed back before returning. So a backward query selected *everything* preceding the cursor and then kept the top `limit` of it in forward order.

With the default `created_at desc, limit=20`:

| Request | Rows matched by WHERE | Returned | Expected |
| --- | --- | --- | --- |
| page 2 → back | 1–20 | 1–20 ✅ | 1–20 |
| page 3 → back | 1–40 | 1–20 ❌ | 21–40 |
| page 6 → back | 1–100 | 1–20 ❌ | 81–100 |

Going back from page 2 worked by accident — the preceding set is exactly one page. From page 3 on, you were teleported to the top of the list.

Two more defects in the same block:

- **`prev_cursor` was `None` on every reverse response.** The `if params.reverse:` branch assigned `next_cursor` and fell through the `else`, so `prev_cursor` was never set — you could step back once, then the Previous control was dead, while `has_previous` still reported `true`.
- **`has_more` / `has_previous` were not swapped.** In reverse mode `has_more` describes rows *behind* the cursor but was returned with forward semantics.

### Who was exposed

Not the cases table UI — `use-cursor-pagination.tsx` hardcodes `reverse: false` and keeps its own client-side cursor stack. But `reverse` is public API surface on the cases endpoints and is plumbed through the registry actions (`core/cases.py`) and the SDK (`sdk/cases.py`), so workflow authors and agents paginating backwards hit it silently: wrong rows, no error.

## Scope

The audit found six more listings with subsets of the same defect:

| Service | ORDER BY inverted | Page re-reversed | Cursors swapped | Flags swapped |
| --- | --- | --- | --- | --- |
| `cases/service.py` `search_cases` | ❌ | ❌ | ❌ | ❌ |
| `tables/service.py` `list_rows` | ❌ | ✅ | ✅ | ✅ |
| `cases/rows/service.py` | ❌ (no-op) | ✅ | ❌ (double) | ✅ |
| `agent/skill/service.py` (×2) | ✅ | ❌ | ❌ | ❌ |
| `agent/preset/service.py` | ✅ | ❌ | ❌ | ❌ |
| `workflow/management/management.py` | ✅ | ✅ | ✅ | ❌ |
| EE inbox `agent_runs.py` | ✅ | ✅ | ✅ | ❌ |

Two worth calling out:

- **`tables/service.py`** has the same page-1 teleport as cases — it reverses the page and swaps the cursors, but the `ORDER BY` only ever branched on `sort_direction`.
- **`cases/rows/service.py`** applied its reverse `ORDER BY` to a statement that already had one. `Select.order_by()` *appends*, so the first key won and the inversion was a no-op. It also derived the cursors from the already-reversed page and then swapped them a second time, pointing both the wrong way.

## Changes

**Shared helpers** in `tracecat/pagination.py` — `take_cursor_page()` trims the `limit + 1` over-fetch; `build_cursor_page()` puts the rows back into display order and swaps the cursors and flags. All seven call sites now route through them, so the trim/reverse/swap step has one definition and one set of tests. Sort-aware callers derive the scan direction with `scan_ascending = (sort_direction == "asc") != params.reverse`.

One behavioral refinement: on an **empty** reverse page, `has_more` is derived from `next_cursor is not None` rather than `cursor is not None`, so an empty backward page can't advertise a page there's no cursor to reach. This mirrors the reasoning already documented in the inbox provider's grouped path.

**Stale-sort cursors now 400 instead of silently rewinding.** The mismatch was already detected (`cursor_data.sort_column == sort_column`), but nothing acted on it: the cursor filter was dropped entirely and the query returned the first page while the response still advertised `prev_cursor` and `has_previous=true`. Tells that this was an unwritten `else` rather than a decision: `cursor_id` was decoded and then went unused on that path, and the sibling implementation in `workflow/management/management.py` *does* write the branch (ID-only fallback, with a comment). Both call sites now call `validate_cursor_sort_column()`, which raises `InvalidCursorError` — a `ValueError` subclass, which the cases and tables routers already map to 400, so no router changes were needed.

This is the one deliberately behavior-changing part of the PR. It's easy to drop if you'd rather keep the current leniency. Note it also rejects sortless cursors (`sort_column: None`), which only pre-#1749 code paths produced; softening that to an ID-only fallback would be a one-line change. The cases table UI is unaffected either way — `use-cursor-pagination.tsx:93-100` resets pagination state in the same render as any sort change, so it never submits a stale-sort cursor.

## Tests

`tests/unit/test_cursor_page.py` (new, 16 tests) covers the helpers directly: the page-3 → back → page-2 walk, cursor round-tripping in both directions, the boundary at page 1, the empty reverse page, and the six validator cases.

Service-level regression tests were added where none existed — **no test exercised `reverse=True` for cases at all** before this:

- `test_cases_service.py`: reverse paging under desc, asc, and enum (`priority`) sorts, plus stale-cursor rejection
- `test_tables_service.py`: multi-page backward walk, plus stale-cursor rejection
- `api/test_api_cases.py`: confirms `InvalidCursorError` surfaces as 400 end-to-end

### Verification

- `tests/unit -m "not (slow or integration or temporal)"`: **6547 passed, 24 skipped, 8 failed** — all 8 failures are `subprocess.TimeoutExpired` in nsjail sandbox tests that shell out to `docker compose run --build`, unrelated to this change and failing identically on a clean tree. -- these are almost certainly just from my local environment and should pass on the CI run.
- `ruff check` / `ruff format --check`: clean.
- `basedpyright --warnings`: 0 errors, 0 warnings.
- The `ORDER BY` inversion was additionally checked against compiled SQL for all four `sort` × `reverse` combinations.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Fixes reverse cursor pagination across listings so "Previous" returns the correct page, and returns 400 for cursors from a different sort. Also supports NULL sort cursors for table rows and unifies keyset logic for consistent behavior.

- **Bug Fixes**
  - Reverse paging now returns the preceding page with correct ORDER BY inversion, page re-reversal, and cursor/flag swaps. Affected: cases, tables, case rows, agent skills & versions, agent preset versions, workflow management, and EE inbox agent runs.
  - Cursors and flags are now consistent in both directions (e.g., `prev_cursor` set in reverse; empty reverse pages don’t advertise an unreachable next page).
  - Stale-sort cursors now raise `InvalidCursorError` (HTTP 400) instead of returning page 1.
  - Tables now follow `sort_value: null` cursors for nullable sort columns and handle `NULLS FIRST/LAST` via `keyset_filter`; EE inbox agent runs validates cross-sort cursors and only allows `created_at`/`updated_at`. Shared helpers `take_cursor_page`, `build_cursor_page`, and `validate_cursor_sort_column` unify trim/reverse/swap and validation across services.

- **Migration**
  - Clear/reset the cursor when changing sort. Clients will get 400 if they send a cursor created under a different sort.

<sup>Written for commit dced07ba5f2ce608f6cf5d258123e483b77f2f75. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3131?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
+++ b/packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
@@ -23,7 +23,12 @@ from tracecat.dsl.client import get_temporal_client
 from tracecat.inbox.schemas import InboxItemRead, UserSummary, WorkflowSummary
 from tracecat.inbox.types import InboxGroup, InboxItemStatus, InboxItemType
 from tracecat.logger import logger
-from tracecat.pagination import BaseCursorPaginator, CursorPaginatedResponse
+from tracecat.pagination import (
+    BaseCursorPaginator,
+    CursorPaginatedResponse,
+    build_cursor_page,
+    take_cursor_page,
+)
 from tracecat_ee.agent.types import AgentWorkflowID
 
 # The error signal is fully persisted (AgentSession.last_error), so Temporal is
@@ -379,69 +384,39 @@ class AgentRunsInboxProvider(BaseCursorPaginator):
         stmt = base_stmt.order_by(order_clause, id_order).limit(limit + 1)
 
         result = await self.session.execute(stmt)
-        sessions = list(result.scalars().all())
+        scanned_sessions, has_more = take_cursor_page(
+            list(result.scalars().all()), limit=limit
+        )
 
-        # Check if there are more items
-        has_more = len(sessions) > limit
-        if has_more:
-            sessions = sessions[:limit]
-
-        # The scan ran in scan_desc order; reverse back into display order so
-        # the page reads in the requested sort regardless of pagination
-        # direction.
-        if reverse:
-            sessions.reverse()
+        # The scan ran in scan_desc order; build_cursor_page reverses the page
+        # back into display order so it reads in the requested sort regardless
+        # of pagination direction, and swaps the cursors and flags accordingly.
+        page = build_cursor_page(
+            scanned_sessions,
+            cursor=cursor,
+            reverse=reverse,
+            has_more=has_more,
+            encode_cursor=lambda session: self.encode_cursor(
+                id=session.id,
+                sort_column=sort_col,
+                # Use the correct column value based on sort_col
+                sort_value=(
+                    session.updated_at
+                    if sort_col == "updated_at"
+                    else session.created_at
+                ),
+            ),
+        )
 
         # Enrich sessions
-        items = await self._enrich_sessions(sessions)
-
-        # Generate cursors
-        next_cursor = None
-        prev_cursor = None
-
-        if items:
-            if has_more:
-                last_item = items[-1]
-                # Get the session for this item to access the sort column value
-                last_session = next(
-                    (s for s in sessions if s.id == last_item.source_id), None
-                )
-                if last_session:
-                    # Use the correct column value based on sort_col
-                    sort_value = (
-                        last_session.updated_at
-                        if sort_col == "updated_at"
-                        else last_session.created_at
-                    )
-                    next_cursor = self.encode_cursor(
-                        id=last_item.id,
-                        sort_column=sort_col,
-                        sort_value=sort_value,
-                    )
-            if cursor:
-                first_item = items[0]
-                first_session = next(
-                    (s for s in sessions if s.id == first_item.source_id), None
-                )
-                if first_session:
-                    # Use the correct column value based on sort_col
-                    sort_value = (
-                        first_session.updated_at
-                        if sort_col == "updated_at"
-                        else first_session.created_at
-                    )
-                    prev_cursor = self.encode_cursor(
-                        id=first_item.id,
-                        sort_column=sort_col,
-                        sort_value=sort_value,
-                    )
+        items = await self._enrich_sessions(page.items)
 
         return CursorPaginatedResponse(
             items=items,
-            next_cursor=next_cursor,
-            prev_cursor=prev_cursor,
-            has_more=has_more,
-            has_previous=cursor is not None,
+            next_cursor=page.next_cursor,
+            prev_cursor=page.prev_cursor,
+            has_more=page.has_more,
+            has_previous=page.has_previous,
             total_estimate=None,
         )
 
--- a/packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
+++ b/packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
@@ -28,6 +28,7 @@ from tracecat.pagination import (
     CursorPaginatedResponse,
     build_cursor_page,
     take_cursor_page,
+    validate_cursor_sort_column,
 )
 from tracecat_ee.agent.types import AgentWorkflowID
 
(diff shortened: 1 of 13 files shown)


daryllimyt:
@codex review

tracecat/pagination.py
```diff
@@ -73,6 +163,43 @@ def parse_datetime_string(
...
+    Raises:
+        InvalidCursorError: The cursor belongs to a different sort.
+    """
+    if cursor.sort_column != sort_column:
+        raise InvalidCursorError(
+            f"Cursor was created for sort column {cursor.sort_column!r}, "
+            f"but this request sorts by {sort_column!r}. "
+            "Restart pagination without a cursor after changing the sort."
+        )
+    if cursor.sort_value is None:
```
daryllimyt, line 230: This treats every `null` sort value as a malformed or missing cursor, but `TablesService.list_rows()` can issue exactly that cursor: custom table columns default to `nullable=True`, and the cursor encoder passes `row.get(sort_column)` through unchanged. With more than `limit` rows sorted descending by a newly added nullable column, PostgreSQL puts the NULL rows first, so page 1 returns a `next_cursor` containing `"sort_value": null`; following it now returns 400 and makes all later rows unreachable.

Please distinguish an omitted legacy sort field from an explicit NULL and implement a NULL-aware keyset order and predicate, including the NULL/non-NULL transition, in both directions. Add a regression test for this server-issued cursor.

packages/tracecat-ee/tracecat_ee/inbox/providers/agent_runs.py
```diff
@@ -379,69 +384,39 @@ async def list_items(
...
-            sessions.reverse()
+        # The scan ran in scan_desc order; build_cursor_page reverses the page
+        # back into display order so it reads in the requested sort regardless
+        # of pagination direction, and swaps the cursors and flags accordingly.
+        page = build_cursor_page(
+            scanned_sessions,
+            cursor=cursor,
+            reverse=reverse,
+            has_more=has_more,
+            encode_cursor=lambda session: self.encode_cursor(
```
daryllimyt, line 407: This provider now uses the canonical page builder, but its cursor consumers still do not validate `cursor_data.sort_column` against the requested `sort_col` in either the ungrouped or grouped path. A `created_at` cursor submitted with `order_by=updated_at` is accepted; I reproduced compiled SQL applying that timestamp to `agent_session.updated_at < ...`, which can skip or duplicate rows instead of returning 400.

Please call `validate_cursor_sort_column(..., sort_column=sort_col, expected_type=datetime)` in both decode paths before applying the keyset predicate.

daryllimyt (changes requested):
Requesting changes for two cursor-contract gaps: server-issued NULL table-sort cursors cannot be consumed, and the touched inbox provider still accepts cross-sort cursors.

PR #3135, opened 2026-07-27
docs(benchmark): add postgres scatter load-test plan

daryllimyt:
## What

Design document for a PostgreSQL scatter/gather load test: `scripts/benchmark/postgres-scatter-load-test-plan.md`.

Establishes the decision the test supports (a safe aggregate DB-connection budget plus bounded, recoverable overload behavior), with a headline product target: **a burst of up to 10,000 concurrent actions, spread across many workflows, absorbed by the Temporal task queue and drained at a constant sustained rate on a small database**, given an explicitly capped executor.

## Contents

- Why current controls (per-workflow 64-task scheduler cap, per-process 10+60 pool, 100-activity executor default, unlimited self-hosted tier) don't compose into a deployment-wide bound.
- Scenario matrix phases 0–8: baseline → constrained characterization → pool budget → bounded executor → bulk-insert control → 10k burst absorption → optional PgBouncer → optional permit-leasing validation → subflow fan-out follow-on.
- Connection-budget invariant (`C_configured <= C_budget`), workload ramp, per-phase acceptance criteria, failure-mode classification, and post-test decision rules (including when a proxy is and is not justified).
- Safety/isolation rules: fresh numbered cluster only, no volume deletion, abort conditions.

## Review guide

The load-bearing sections are the scenario matrix, the connection budget, and the acceptance criteria. Organization permit leasing is deliberately excluded from the primary matrix (it is not activated in current deployments) and parked as optional phase 7.

PR #3141, opened 2026-07-27
fix(db): isolate bypass-RLS sessions on a dedicated pool

daryllimyt:
## What

Introduce a small, auth-specific SQLAlchemy connection pool so short authentication and authorization lookups do not contend with request-scoped database work.

- Add `get_async_auth_engine()` and `get_async_session_auth_context_manager()` for the auth bulkhead.
- Route workspace-to-organization resolution, effective-scope resolution, managed API-key authentication, first-user existence checks, user organization/domain lookups, and SAML policy reads through that pool.
- Keep general RLS-bypass work such as background jobs, registry sync, and administrative endpoints on the main engine.
- Enforce one-way acquisition: main-pool code may acquire auth capacity, but auth-pool code may not acquire the main pool or a nested auth session.
- Expose `TRACECAT__DB_AUTH_POOL_SIZE` and `TRACECAT__DB_AUTH_MAX_OVERFLOW`, both defaulting to `5`, through Docker Compose and Fargate.
- Attribute auth connections as `<service>-auth` in `pg_stat_activity`.
- Return a typed HTTP 503 with code `auth_database_unavailable` when auth-pool checkout times out.

The request-session path (`get_async_session` / `AsyncDBSession`) and general bypass-session binding remain unchanged.

## Why

Under main-pool saturation, requests can hold a request-scoped connection with an open transaction while the auth dependency chain waits for another connection from the same pool. Once every connection is held by a request waiting for one more, no request can progress: ordinary API requests, login, and executor Action Gateway calls fail until the service recovers.

This change removes that application-pool cycle. Auth work has independent admission capacity and a structurally enforced acquisition order:

```text
main -> auth   allowed
auth -> main   rejected
auth -> auth   rejected
```

Pool exhaustion remains possible, but it is bounded and retryable instead of becoming a same-pool hold-and-wait deadlock.

## Safety properties

- `AuthSession` exposes only the small capability surface required by current auth call sites.
- API-key workspace resolution reuses its already-held auth session, including on a cold cache.
- Main-engine checkout has a runtime guard in addition to accessor-level checks.
- Auth-pool checkout timeout is converted to `AuthPoolExhaustedError` and handled by both the API and Action Gateway.
- The auth pool is an operational bulkhead, not a privilege boundary: it uses the same database URI, role, and RLS bypass mechanism as the main pool.

## Verification

- Current GitHub unit, registry, Temporal, Ruff, Pyright, CodeQL, and security checks pass.
- Focused local suites: 105 passed.
- `uv run ruff check`, `uv run ruff format --check`, and targeted `uv run basedpyright`: clean.
- Terraform formatting and all three Docker Compose configurations validate.
- Real PostgreSQL smoke tests confirmed:
  - two requests holding every main-pool connection still completed auth lookups through a one-connection auth pool;
  - a main-pool checkout attempted while holding auth capacity was rejected.

## Scope and follow-ups

This is step 1 for the observed SQLAlchemy pool self-deadlock. It does not eliminate PostgreSQL lock deadlocks, global database connection exhaustion, or every remaining nested database-session pattern.

Follow-ups:

- Reuse the request session through the auth chain where safe so nested acquisition disappears entirely.
- Budget the additional auth capacity per process against deployment-wide PostgreSQL connection headroom and canary the rollout with auth-pool latency/503 monitoring.
- Use a separate PostgreSQL role for a true privilege boundary (ENG-1570).


daryllimyt:
* **#3141** <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/3141?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/3141?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3134** <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/3134?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3135](https://github.com/TracecatHQ/tracecat/pull/3135) <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/3135?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

tracecat/db/engine.py
```diff
@@ -249,6 +333,7 @@ async def get_async_session_bypass_rls() -> AsyncGenerator[AsyncSession, None]:
     WARNING: Use sparingly and only when necessary. Prefer get_async_session()
     with proper role context for most operations.
     """
+    _assert_main_pool_checkout_allowed()
     async with AsyncSession(get_async_engine(), expire_on_commit=False) as session:
```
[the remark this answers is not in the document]
daryllimyt: Fixed in 6c14865c8. The three short lookups in the login/registration path now run on the auth bulkhead via `get_async_session_auth_context_manager()`: `UserManager.validate_email` (first-user check), `_list_user_org_ids`, and `_get_org_id_for_email_domain`. `list_users` now takes the `SupportsExecute` protocol so both pools' sessions satisfy it.

Transactional paths (`_accept_pending_invitation`, `get_or_create_user`) intentionally stay on the main pool — they need the full `AsyncSession` surface that `AuthSession` deliberately withholds.

Regression tests in `tests/unit/test_rls_failure_repros.py` assert the three helpers execute on the auth engine bind, including one that models the production lifetime (main-pool `get_user_db` session held while the lookup runs, i.e. the allowed main→auth acquisition order). The misleading invitation-route ordering test was removed. 40f8c2645 additionally exposes `TRACECAT__DB_AUTH_POOL_SIZE` / `TRACECAT__DB_AUTH_MAX_OVERFLOW` in Fargate and Compose so the new pool's connection budget is deployable.

daryllimyt:
@codex review

tracecat/api/common.py
```diff
@@ -56,6 +57,37 @@ async def http_exception_handler(request: Request, exc: Exception) -> Response:
     return await default_http_handler(request, http_exc)
 
 
+def auth_pool_exhausted_exception_handler(
+    request: Request,
+    exc: Exception,
+) -> Response:
+    """Return a bounded, retryable response when auth DB capacity is exhausted."""
```
jordan-umusu: The comment here says retryable, but The SDK raises TracecatAPIError, and actions default to one attempt, so the action still fails. Could we clarify the wording or add a bounded retry for auth_database_unavailable?

tracecat/auth/users.py
```diff
@@ -220,30 +220,33 @@ async def _get_org_id_for_email_domain(self, email: str) -> OrganizationID | Non
...
         if AuthType.SAML not in config.TRACECAT__AUTH_TYPES:
             return False
 
-        saml_enabled = bool(
-            await get_setting(
-                "saml_enabled",
-                role=bootstrap_role(org_id),
-                default=True,
+        async with get_async_session_auth_context_manager() as session:
+            saml_enabled = bool(
```
jordan-umusu, line 232: ``/auth/discover`` still performs a nested main-pool checkout when reading saml_enabled. This predates the PR, but the new ``get_setting_from_bypass_session`` helper could let discovery reuse its existing bypass session and remove that nesting.

jordan-umusu (approved):
lgtm, two minor comments

jordan-umusu:
oh and we need these knobs on k8s as well? i didnt see these in the helm

PR #3165, opened 2026-07-31
fix(integrations): unblock catalog MCP OAuth setup

jordan-umusu:
## Summary by cubic
Unblocks MCP catalog OAuth setup by moving OAuth discovery/registration to the backend and strictly binding requests to a single catalog recipe. Adds one‑click Connect support with `connection_option_id` for predictable option selection and safer updates.

- **Bug Fixes**
  - Catalog resolver picks one recipe by server type, auth, and URI policy; rejects embedded creds, ambiguous matches, and off‑vendor hosts; supports stdio and adds stable option ids for bare-spec rows.
  - OAuth: uses catalog‑pinned endpoints when present (skips discovery); otherwise discovers with host allowlists; accepts BYO client creds and falls back to DCR.
  - Updates: changing an OAuth server URI requires reauth; leaving OAuth clears the grant; a grant can’t be reused or laundered across auth changes; only revalidates against the catalog when the connection changes (metadata edits allowed if recipes drift).
  - One‑click Connect binds via `connection_option_id` or the catalog default; rejects unknown option ids. Frontend passes the option id from catalog cards and removes client‑side provider creation/OAuth parsing, using backend discovery via `connectMcpIntegration`.

<sup>Written for commit 49ecc92ddb440b7e238d8fb221ea00ff5d254564. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3165?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tracecat/integrations/catalog/resolver.py
+++ b/tracecat/integrations/catalog/resolver.py
@@ -1,0 +1,168 @@
+"""Bind MCP connection requests to platform catalog connect recipes.
+
+The platform catalog is repo-owned data; a workspace request references it by
+``catalog_slug``. This module owns the binding between such a request and the
+one connect recipe it targets, so service code never re-infers recipe identity
+from transport/auth/URI heuristics.
+
+The server URI policy is credential-driven. A recipe that declares a
+``target: "server_uri"`` credential (or ships no URI at all) delegates the
+URI to the user; a recipe with a bare literal URI is a fixed vendor endpoint
+and must match exactly. The exact match is a security boundary: a trusted
+catalog row must not be bound to an arbitrary host that would then receive
+its tokens over the MCP transport.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from urllib.parse import urlparse
+
+from tracecat.integrations.catalog.loader import (
+    get_platform_mcp_catalog_entry_by_slug,
+)
+from tracecat.integrations.catalog.types import PlatformMCPCatalogEntry
+from tracecat.integrations.enums import MCPAuthType
+from tracecat.integrations.schemas import (
+    MCPConnectionOption,
+    MCPConnectionSpec,
+    MCPHTTPCustomConnectionSpec,
+    MCPHTTPNoneConnectionSpec,
+    MCPHTTPOAuth2ConnectionSpec,
+)
+from tracecat.integrations.types import MCPServerType
+
+MCPHTTPConnectionSpec = (
+    MCPHTTPOAuth2ConnectionSpec
+    | MCPHTTPCustomConnectionSpec
+    | MCPHTTPNoneConnectionSpec
+)
+
+
+class CatalogConnectionError(ValueError):
+    """Requested connection does not bind to a catalog connect recipe."""
+
+
+@dataclass(slots=True, frozen=True)
+class ResolvedCatalogConnection:
+    """One catalog connect recipe a request has been bound to."""
+
+    entry: PlatformMCPCatalogEntry
+    option: MCPConnectionOption
+
+    @property
+    def option_id(self) -> str:
+        return self.option.id
+
+    @property
+    def spec(self) -> MCPConnectionSpec:
+        return self.option.connection_spec
+
+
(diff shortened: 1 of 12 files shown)


tracecat/integrations/service.py
```diff
@@ -4108,9 +4209,44 @@ async def update_mcp_integration(
                 target_oauth_integration_id = params.oauth_integration_id
             elif server_type_changed:
                 target_oauth_integration_id = None
+            elif target_auth_type != MCPAuthType.OAUTH2:
```
daryllimyt: **P1 — Force-clear OAuth grants on non-OAuth targets**

An explicitly supplied oauth_integration_id wins before this non-OAuth clearing branch. Starting from OAUTH2, a caller can switch to CUSTOM and a new URI while resubmitting the old grant ID, then switch back to OAUTH2 while omitting the ID. The dormant grant is reused and the pre-save connectivity probe sends its bearer token to the new endpoint. Resolve target_oauth_integration_id as None whenever target_auth_type is not OAUTH2, regardless of field presence, and add a two-step regression test.

tracecat/integrations/catalog/resolver.py
```diff
@@ -0,0 +1,168 @@
...
+    return _server_uri_mismatch(spec, server_uri=server_uri)
+
+
+def _server_uri_mismatch(spec: MCPHTTPConnectionSpec, *, server_uri: str) -> str | None:
+    if server_uri_is_user_supplied(spec):
+        parsed = urlparse(server_uri)
+        if parsed.username is not None or parsed.password is not None:
+            return "server URI must not embed credentials"
+        return None
+    if server_uri != spec.server_uri:
```
daryllimyt, line 166: **P2 — Preserve documented regional and query endpoint variants**

This exact comparison treats every recipe without a server_uri credential as one immutable literal URL, but the shipped Datadog and Vanta metadata documents supported variants: Datadog regional hosts and toolsets query values (packages/tracecat-ee/tracecat_ee/mcp/catalog/mcp_catalog_private.json:416), and Vanta EU/AUS endpoints (the same file at line 2207). Catalog-backed creates and connection-changing updates now reject those valid URLs. Model the variants as constrained options/templates, or add an explicit host/path/query policy that preserves the intended token-destination boundary.

daryllimyt (changes requested):
lgtm besides thermos comments

tracecat/integrations/service.py
jordan-umusu: Resolved.

tracecat/integrations/catalog/resolver.py
jordan-umusu: Real catalog gap that is easily addressed by custom mcp. Can resolve in a followup if necessary. See my response to the codex comment below.

tracecat/integrations/catalog/resolver.py
```diff
@@ -0,0 +1,168 @@
...
+        return f"connection option uses {spec.auth_type} authentication"
+    return _server_uri_mismatch(spec, server_uri=server_uri)
+
+
+def _server_uri_mismatch(spec: MCPHTTPConnectionSpec, *, server_uri: str) -> str | None:
+    if server_uri_is_user_supplied(spec):
+        parsed = urlparse(server_uri)
+        if parsed.username is not None or parsed.password is not None:
+            return "server URI must not embed credentials"
+        return None
```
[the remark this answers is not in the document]
jordan-umusu: Valid, but out of scope: `main` rejects these too (it never compared `server_uri` at all), so this PR is strictly tighter. Fixing it is data — ~127 lines of `connection_options`, no code — but the hostnames need vendor-doc confirmation across 16 rows, and guessing wrong rejects valid endpoints. Custom MCP servers reach these today.

daryllimyt (approved):
lgtm

PR #3171, opened 2026-07-31
perf(cases): defer case number allocation

daryllimyt:
## Summary

- insert in-flight cases with a transaction-local sentinel that does not lock the workspace counter
- allocate and assign the real workspace case number only after field, event, and dropdown writes have flushed
- make workspace case-number uniqueness deferred so concurrent sentinels can coexist
- enforce at commit that no sentinel can persist
- retain the legacy trigger path for older application versions that insert without an explicit case number
- build the replacement unique index concurrently during migration

## Root cause

The existing `BEFORE INSERT` trigger incremented `workspace.last_case_number`. PostgreSQL held that row lock until the complete case transaction committed, so field, event, and dropdown work extended the critical section and made concurrent case inserts wait.

## Impact

The counter lock is held only across the final number assignment and commit. Numbering remains gapless on rollback, related case writes remain atomic, and rolling deployments remain compatible with older application versions.

## Validation

- `TRACECAT__DB_ENCRYPTION_KEY=<synthetic-test-key> uv run pytest tests/unit/test_cases_service.py -q` — 55 passed
- `uv run pytest tests/migration/test_deferred_case_number_allocation_migration.py -q` — 2 passed
- migration upgrade, downgrade, and re-upgrade completed successfully
- `uv run ruff check` and `uv run ruff format --check` on changed files
- targeted `uv run basedpyright` — 0 diagnostics
- full signed-commit hooks, including repository Python typechecking

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 29 | 2 |
| Tests | 629 | 12 |
| Migration | 180 | 0 |

## Tracking

[ENG-1575](https://linear.app/tracecat/issue/ENG-1575/perfdb-investigate-slow-registry-lookups-and-case-inserts)

---
## Summary by cubic
Defers workspace case-number allocation to commit time using a `case_number=0` sentinel to remove counter-row contention. Adds a deferred DB guard and trigger-side checks to keep numbering gapless and safe while staying compatible with older app versions. Addresses ENG-1575.

- **Refactors**
  - Insert with a sentinel; assign the real number at the end via `_assign_next_case_number`.
  - Hardened trigger behavior: keep legacy `NULL` path, accept explicit `case_number > 0` imports without locking, and validate workspace existence.
  - New tests cover concurrency without lock contention, rollback gaplessness, and guard enforcement.

- **Migration**
  - Replace the unique index concurrently and make it `DEFERRABLE INITIALLY DEFERRED`.
  - Install `assign_workspace_case_number()` and the `require_assigned_workspace_case_number()` constraint trigger to reject persisted sentinels.
  - Downgrade restores immediate allocation and a non-deferred constraint; ordering ensures the constraint is replaced before disabling the guard.

<sup>Written for commit fe2da1ac0edfb5d3b76a001fe03a1c94cbb6b78e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3171?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/alembic/versions/2792569cf359_defer_case_number_allocation.py
+++ b/alembic/versions/2792569cf359_defer_case_number_allocation.py
@@ -1,0 +1,180 @@
+"""Defer case-number allocation until the end of case creation.
+
+Revision ID: 2792569cf359
+Revises: 864d277bedfa
+Create Date: 2026-07-31 18:00:24.238167
+
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "2792569cf359"
+down_revision: str | None = "864d277bedfa"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+_CASE_NUMBER_CONSTRAINT = "uq_case_workspace_case_number"
+_REPLACEMENT_INDEX = "ix_case_workspace_case_number_replacement"
+
+
+def _create_replacement_index() -> None:
+    """Build a replacement index without blocking case writes."""
+    with op.get_context().autocommit_block():
+        # A failed migration can leave the concurrently-created index behind.
+        # Rebuild it so a retry never adopts an invalid partial index.
+        op.execute(f'DROP INDEX CONCURRENTLY IF EXISTS "{_REPLACEMENT_INDEX}"')
+        op.execute(
+            f"""
+            CREATE UNIQUE INDEX CONCURRENTLY "{_REPLACEMENT_INDEX}"
+            ON "case" (workspace_id, case_number)
+            """
+        )
+
+
+def _replace_case_number_constraint(*, deferred: bool) -> None:
+    _create_replacement_index()
+    op.drop_constraint(_CASE_NUMBER_CONSTRAINT, "case", type_="unique")
+    deferred_clause = " DEFERRABLE INITIALLY DEFERRED" if deferred else ""
+    op.execute(
+        f"""
+        ALTER TABLE "case"
+        ADD CONSTRAINT "{_CASE_NUMBER_CONSTRAINT}"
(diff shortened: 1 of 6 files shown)


daryllimyt:
@codex review

alembic/versions/2792569cf359_defer_case_number_allocation.py
```diff
@@ -0,0 +1,180 @@
...
+            f"""
+            CREATE UNIQUE INDEX CONCURRENTLY "{_REPLACEMENT_INDEX}"
+            ON "case" (workspace_id, case_number)
+            """
+        )
+
+
+def _replace_case_number_constraint(*, deferred: bool) -> None:
+    _create_replacement_index()
+    op.drop_constraint(_CASE_NUMBER_CONSTRAINT, "case", type_="unique")
```
jordan-umusu, line 40: From Thermos:

[P2] The constraint swap can indefinitely stall all case traffic. alembic/versions/2792569cf359_defer_case_number_allocation.py:38 build the index concurrently, but DROP CONSTRAINT and ADD CONSTRAINT still require an ACCESS EXCLUSIVE table lock. Production migrations run while existing API tasks may remain active, and no migration-level lock_timeout is configured. A long or idle transaction touching "case" can therefore hold the migration indefinitely; once the exclusive request queues, subsequent case reads and writes queue behind it. Bound lock acquisition after the concurrent build and abort safely for deployment retry.

jordan-umusu (approved):
LGTM, just thermos comment

aaron-tracecat:
lgtm

aaron-tracecat (approved):

PR #3201, opened 2026-08-04
fix(auth): mark session cookie Secure when public API uses HTTPS

Sirius1616:
## Summary

The auth session cookie's `Secure` flag was derived from `TRACECAT__API_URL`, the **internal** API URL that deployments set to plain HTTP (e.g. `http://api-service:8000` in the Fargate module at `deployments/fargate/modules/ecs/locals.tf:19`). This caused the session cookie to be sent **without** the `Secure` attribute even when the public API is served over HTTPS, so it could be transmitted over plaintext HTTP (CWE-614).

This PR derives the flag from `TRACECAT__PUBLIC_API_URL` instead — the browser-facing URL that already determines the cookie name (`tracecat/auth/users.py`), and which production sets to `https://<domain>/api`.

## Changes

- `tracecat/auth/users.py`: base `cookie_secure` on `TRACECAT__PUBLIC_API_URL` scheme
- `tests/unit/test_auth.py`: regression test covering both https/http public URLs, and asserting the internal API URL never gates the flag

## Verification

- `uv run pytest tests/unit/test_auth.py` — 9 passed (2 new)
- `uv run ruff check`, `uv run ruff format --check` — clean
- `uv run basedpyright --warnings --threads 4` — 0 errors

---
## Summary by cubic
Marks the auth session cookie as Secure when the public API is served over HTTPS, preventing it from being sent over plaintext HTTP.

- **Bug Fixes**
  - Derive `cookie_secure` from `TRACECAT__PUBLIC_API_URL` in `tracecat/auth/users.py` (not `TRACECAT__API_URL`).
  - Add regression tests to cover HTTPS/HTTP public URLs and ensure the internal API URL never gates the flag.

<sup>Written for commit 40b84ab5d97490cff4608286224c5b0e17c36b93. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/unit/test_auth.py
+++ b/tests/unit/test_auth.py
@@ -1,3 +1,4 @@
+import importlib
 import os
 import uuid
 
@@ -5,6 +6,7 @@ import pytest
 from httpx import AsyncClient
 
 from tracecat import config
+from tracecat.auth import users as users_mod
 from tracecat.auth.types import Role
 from tracecat.clients import AuthenticatedAPIClient, AuthenticatedServiceClient
 from tracecat.contexts import ctx_role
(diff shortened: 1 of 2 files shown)


topher-lo (approved):
@Sirius1616 thank you for finding this 🫡

PR #3210, opened 2026-08-07
fix(agent): scope uv cache to executor jobs

daryllimyt:
## Summary

- Isolate all writable UV runtime state beneath each per-turn agent job directory, mount it read-write at `/run/tracecat/uv-state` inside NSJail, and use the same job-scoped boundary in direct development mode.
- Pin UV cache, credentials, Python, tool, and link-mode settings at the agent transport boundary so runtime and MCP configuration cannot redirect writes into durable session storage.
- Sanitize protected environment keys and pre-command `uvx` options from legacy MCP configurations while preserving server arguments after the command boundary.

## Why

Agent executor session homes previously retained UV state across turns. UV and `uvx` package artifacts could therefore accumulate for the lifetime of stable session directories and consume enough pod-local storage to trigger ephemeral-storage eviction.

The job-owned state directory is created with restrictive permissions and removed by existing job cleanup. This intentionally accepts cold package downloads per turn as a short-term bridge to Firecracker VM snapshots, where prepared package state can be captured and resumed without a shared writable cache.

## Scope

This addresses agent-executor UV state ownership, override protection, and legacy MCP compatibility. It does not change the normal executor registry cache or worker concurrency and memory limits.

## Validation

At head `c42eb8018`, GitHub Actions passes:

- Ruff, Pyright, and repository quality checks
- Unit, Temporal, and registry test suites
- Python, JavaScript/TypeScript, and Actions analysis
- CodeQL and Socket Security checks

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 360 | 69 |
| Tests | 461 | 18 |

---
## Summary by cubic
Scopes all `uv` runtime state to each executor job and closes override paths in `uvx` and env to stop pod-local storage growth. Pins job-scoped `uv` settings (including `UV_PYTHON_CACHE_DIR`) at the transport boundary, enforces `UV_LINK_MODE=copy`, and sanitizes legacy MCP configs at resolution and runtime.

- **Bug Fixes**
  - Per-job state under `/run/tracecat/uv-state` (0700, RW) in NSJail; direct mode uses `<job>/uv-state`. Env is job-scoped: `UV_CACHE_DIR`, `UV_CREDENTIALS_DIR`, `UV_PYTHON_INSTALL_DIR`, `UV_PYTHON_BIN_DIR`, `UV_PYTHON_CACHE_DIR`, `UV_TOOL_DIR`, `UV_TOOL_BIN_DIR`, and `UV_LINK_MODE=copy`. Job dirs are auto-allocated and cleaned on failure or via `cleanup_spawned_runtime`.
  - Protects overrides: `AgentSandboxConfig`/`build_agent_env_map` reserve the above vars and `UV_LINK_MODE`. Preset resolution and runtime strip protected `uvx` options (`--cache-dir`, `--link-mode`) and protected env keys from legacy configs; `mcp_validation` rejects those overrides. Sanitization stops at the `uvx` command or `--` so tool args are preserved.
  - Transport pins settings at the boundary: injects the job-scoped env via `options.settings`, using `/run/tracecat/uv-state` in NSJail or the job dir in direct mode.

<sup>Written for commit c42eb801881862450e9a0c5dbcd38a102e614ade. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3210?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


daryllimyt:
@codex review

tracecat/agent/common/config.py
```diff
@@ -27,6 +27,35 @@
 ).lower() in ("true", "1")
 """Disable nsjail sandbox and use the unsafe PID executor instead."""
 
+_AGENT_RUNTIME_UV_PATH_ENV_VARS = (
```
jordan-umusu, line 30: non-blocking: can this be a dict instead?

jordan-umusu (approved):
LGTM, but we probably need a patch for npm/npx as well

PR #3211, opened 2026-08-07
feat(ui): agent mention autocomplete in case comment composer

aaron-tracecat:
## Screen Recording

https://github.com/user-attachments/assets/c719fa82-1856-4cd3-98bb-90035e0589f5

## Summary

Adds `@`-autocomplete to the case comment composer so users can mention agent presets. The composer reads like Linear's: typing `@` opens a suggestion menu, and picking one drops a highlighted `@Preset name` chip-like run into the text. What the API receives is still the shared wire token `[@<Preset name>](mention://agent/<preset_id>)`.

- Applies to both the top-level comment composer and the inline reply composer. The edit composer is intentionally excluded — mentions only fire on create, and edits can still hand-type tokens.
- Suggestions come from `useAgentPresets(workspaceId)`, filtered by preset name as you type and capped at 8 results.
- Trigger rules match the chat composer: `@` at the start of the input or after whitespace opens the popover, whitespace in the query dismisses it, `Escape` dismisses, `ArrowUp`/`ArrowDown` navigate with wraparound, `Enter` or `Tab` selects, and a mouse click selects.
- `Enter` and `Tab` are consumed by the popover while it is open, so they select instead of submitting the comment, inserting a newline, or moving focus out of the composer. `Shift+Tab` is left alone and keeps native focus behaviour.
- Gated on the `agent:execute` **and** `agent:read` scopes **and** both the `agent_addons` and `case_addons` entitlements. When gated, the popover never opens, the preset query is never enabled, and `@` behaves exactly as it does today. `agent:read` is required because the suggestion list comes from the preset-list endpoint, which `tracecat/agent/preset/router.py` guards with that scope — a role holding execute but not read would otherwise get a 403 and see a misleading "No agents found". Mirrors the workspace chat gate.

The composer stays a plain `<textarea>`. A rich-text editor was considered and rejected as far too invasive for this surface.

### Display-value mapping

The textarea's value holds *display* text — a mention reads as `@Triage agent` — and the composer keeps a side list of `MentionRange` offsets (`{ start, end, label, targetId }`) pointing into it. `serializeMentions` converts display text plus ranges into the wire format on submit, so the shared token contract is unchanged. The 25k length limit is validated against the **serialized** string rather than the shorter display text.

Keeping those offsets honest across editing is the bulk of the work:

- **Offset maintenance.** Every controlled change derives a single-splice diff (`start`, `deleted`, `inserted`) from the old value, the new value and the caret, then shifts the ranges. The common suffix is measured before the prefix and capped at the caret, because everything after the caret is untouched by definition; measuring the prefix first biases the splice to the right and shifts the wrong mentions when the edit sits among repeated characters.
- **Dissolve on edit.** Any edit that reaches into a mention drops it from the list. The characters stay in the text as ordinary prose — the run simply stops being a mention, which is what makes partial edits predictable.
- **Atomic delete.** Backspace with a collapsed caret immediately after a mention removes the whole mention in one keystroke, Linear-style, instead of nibbling one character at a time.
- **Image paste.** `useCommentImagePaste` reports its own splice through the same remapping path, so pasting a screenshot above a mention keeps that mention aligned.

### Overlay styling

Mentions are highlighted with the standard backdrop technique: a div behind the textarea renders the same display text with mention runs wrapped in tinted spans, and the textarea renders its own text transparent while keeping `caret-color` so the native caret stays visible.

Metric parity between the two layers is what makes this safe, and it is guaranteed structurally rather than by eye: the textarea and the overlay are handed the *same* typography and padding class string from a single `textMetricsClassName` constant, the overlay sets `white-space: pre-wrap` and `overflow-wrap: break-word` to match textarea wrapping, and Tailwind's preflight makes the textarea inherit the page font so both boxes resolve the same family, size and line height. The overlay is `absolute inset-0` inside the wrapper the textarea sizes, so it tracks the auto-grow height for free.

Mentions are tinted but deliberately **not** bolded. A weight change alters glyph widths, which shifts every following character and desynchronises the native caret from the text the user sees. That constraint is documented at the overlay and asserted in a test.

The composer auto-grows and pins `overflow-y: hidden`, so the textarea never scrolls and the overlay needs no scroll synchronisation.

### Popover behaviour

Modelled on Linear's `@`-mention menu.

- **Pinned to the `@` trigger.** A textarea exposes no caret geometry, so `frontend/src/lib/textarea-caret.ts` measures it with the standard mirror technique: a hidden div copies every style that affects text layout (box sizing, width, padding, borders, the full font shorthand, letter/word spacing, tab size, text align/indent/transform), receives the text up to the offset, and a marker span is measured for its position. The popover anchors to a 1px marker placed at those coordinates inside a relative wrapper, so it sits just above the trigger's line rather than above the whole composer. The measurement is taken at the **token start** — the `@` itself — and held for the duration of the mention session, so the menu stays pinned where the `@` is while the query grows instead of drifting with the caret. It is re-measured only when the token start offset changes, i.e. when a new `@` session begins.
- **Compact.** Fixed `w-64` content (not anchor width), `max-h-52` scroll region of roughly six rows, single-line `text-xs` items. It is portaled, so it floats over the comment body instead of pushing layout, and Radix's collision detection flips it below the caret when there is no room above.
- **Sectioned.** Suggestions are modelled as `Array<{ section: "agents" | "users" | "cases" | "tables"; label: string; items: MentionSuggestion[] }>` with a Linear-style uppercase muted section header. Only the `Agents` section is populated and rendered in this PR; the shape exists so later sources drop in without reshaping the model. Keyboard navigation runs over the flattened item list, so it already moves across sections seamlessly and skips headers.
- **Empty state.** A query that matches nothing keeps the popover open on a single muted "No agents found" row. `Enter` and `Tab` are swallowed there without selecting anything, and `Escape` or whitespace still dismisses.

Pasting a raw `[@x](mention://agent/...)` token as text is out of scope: it stays visible as raw text and still resolves server-side. The submitted comment renders mentions as markdown links in the viewer; mention chips in the rendered comment are a separate PR.

### Files

| File | Change |
| --- | --- |
| `frontend/src/lib/comment-mentions.ts` | New. All pure mention logic: token detection (`getAgentMentionToken`), display/wire mapping (`serializeMentions`, `diffTextSplice`, `remapMentions`), pure edits (`applyMentionInsertion`, `applyMentionRemoval`, `findMentionEndingAt`) and overlay segmentation (`buildMentionSegments`). |
| `frontend/src/lib/textarea-caret.ts` | New. Mirror-div caret measurement, split into a DOM step and a pure `resolveCaretCoordinates` geometry step. |
| `frontend/src/hooks/use-comment-mentions.ts` | New. The single mention hook: gating, the sectioned suggestion model, filtering, keyboard nav (popover keys + atomic backspace), caret pinning, mention ranges, splice application and submit-time serialization. |
| `frontend/src/components/agents/agent-mention-popover.tsx` | New. Caret-anchored, portaled, sectioned suggestion list. |
| `frontend/src/components/cases/comment-mention-overlay.tsx` | New. Highlight backdrop behind the composer textarea. |
| `frontend/src/components/cases/case-comments-section.tsx` | Thin wiring: hands the hook `getText`/`setText`, routes `onChange`/`onSelect`/`onKeyDown` through it, renders the overlay and serializes on submit. |
| `frontend/tests/comment-mentions.test.ts` | New. Token detection, serializer, diff, remap, pure-edit and segment unit tests. |
| `frontend/tests/textarea-caret.test.ts` | New. Caret geometry unit tests. |
| `frontend/tests/case-comments-section.test.tsx` | Composer behaviour tests plus mocks for `useAgentPresets`, `useScopeCheck` and the caret util. |

### Internals

After the feature settled, mention state that had accumulated across the UX iterations was consolidated: the autocomplete hook and the composer's inline range plumbing merged into one `useCommentMentions` hook, all pure logic (token detection, splice math, pure insert/remove edits) moved into `lib/comment-mentions.ts` where it is unit-tested, and vestigial API surface from earlier iterations (unused `enabled` option, unused `query` field, the `onSelect` indirection) was deleted. The composer now touches mentions through a single hook API.

Closes ENG-1600
https://linear.app/tracecat/issue/ENG-1600

## Test plan

**`frontend/tests/comment-mentions.test.ts`**

- **Token detection** — `@` at the start, `@` after whitespace, and the negative cases (no `@`, `@` after a non-space character such as in an email address, whitespace inside the query, text after the caret).
- **Serializer** — no mentions, a mention surrounded by text, a mention at the start, a mention at the end, multiple adjacent mentions, and ranges that fall outside the text.
- **Splice diff** — insertion at the caret, single-character deletion, and the repeated-character case where only the caret disambiguates the offset.
- **Remap** — insert before (shifts), insert after (untouched), insert exactly at the start boundary (shifts), edit inside (dissolves), deletion overlapping an edge (dissolves), and multiple mentions remapped independently.
- **Pure edits** — `applyMentionInsertion` replaces the `@query`, registers the range and shifts later mentions; `applyMentionRemoval` removes the mention and lands the caret at its start.
- **Atomic delete + segments** — `findMentionEndingAt` matches only the exact end offset; `buildMentionSegments` splits text into plain and mention runs.

**`frontend/tests/textarea-caret.test.ts`** — `resolveCaretCoordinates` adds the textarea's border widths back to the mirror measurements, subtracts scroll offsets, and passes line height through. `getTextareaCaretCoordinates` returns finite coordinates against a controlled style fixture, falls back to a font-derived height when `line-height: normal` does not parse, and removes its measurement mirror from the document.

**`frontend/tests/case-comments-section.test.tsx`**

- **Display value and overlay** — selecting a suggestion puts `@Triage agent ` in the textarea (not the wire token); the overlay renders a highlighted span carrying the target id, with the accent class and explicitly without `font-bold`.
- **Submit** — a composed comment reaches `createComment` as `Ping [@Triage agent](mention://agent/preset-1)`.
- **Atomic delete** — Backspace with the caret directly after a mention clears the whole mention.
- **Positioning** — the content resolves to `data-side="top"`, is portaled outside the composer's `<form>` so the thread's `overflow-hidden` cannot clip it, and anchors to a marker rather than the composer. The measurement util is spied on to assert it is called once per mention session at the `@` offset and that the marker position does not move as the query grows, plus a second test that a fresh `@` session re-measures at the new offset.
- **Sections and empty state** — suggestions render under an `Agents` header; a non-matching query keeps the popover open on "No agents found" with no section header, and `Escape` still dismisses it.
- **Popover behaviour** — opens on `@` and lists presets, filters by query, ignores `@` that does not follow whitespace, dismisses on whitespace in the query and on `Escape`.
- **Keyboard selection** — `ArrowDown` then `Enter` inserts the second preset, asserts the composer value, asserts `createComment` was **not** called, asserts the popover closed, and asserts the caret landed after the insertion. `ArrowUp` wraps to the last suggestion. `Tab` inserts the highlighted preset while keeping focus on the textarea and without submitting, and `Shift+Tab` in the same test leaves the text untouched.
- **Mouse selection** — clicking a suggestion in the inline reply composer inserts into the reply composer only, leaving the top-level composer untouched.
- **Gating** — no popover without the agent scopes (and `@` remains as typed); no popover when only one of `case_addons` / `agent_addons` is present; with `agent:execute` but not `agent:read` the popover stays closed and `useAgentPresets` is asserted to be called with `enabled: false`, so the 403 can never fire; and with both scopes granted the popover opens and the query is enabled. A test helper mirrors real `useScopeCheck` semantics (single scope, scope list, and the `all` option) rather than stubbing a blanket boolean.

Verification run locally:

```
pnpm -C frontend exec biome check --write .   # Checked 949 files. No fixes applied.
pnpm -C frontend check                        # Checked 949 files. No fixes applied.
pnpm -C frontend run typecheck                # clean
pnpm -C frontend test                         # 100 suites, 605 tests passed
```

## LOC breakdown

| Category | + | - |
| --- | --- | --- |
| Logic | 1018 | 26 |
| Tests | 775 | 7 |


daryllimyt:
@codex review

daryllimyt (approved):
lgtm

PR #3212, opened 2026-08-07
feat(ui): render agent mention chips in case comment viewer

aaron-tracecat:
## Summary

Case comments are plain markdown rendered through a read-only TipTap viewer, and agent mentions are encoded inline as ordinary markdown links: `[@Label](mention://agent/<preset_uuid>)`. Until now they rendered as plain links. This PR renders them as inert inline mention chips.

The displayed name always comes from the preset **resolved by id**, never from the label embedded in the markdown. That is deliberate: a hand-typed token can carry a label that disagrees with the id it points at, and presets get renamed after the comment is written. The embedded label is used only as a fallback when the id does not resolve, and is then given a visually distinct "unavailable" treatment (dashed border, muted foreground, `title="Agent preset unavailable"`).

Malformed hrefs (non-UUID id, unknown path segment, missing/extra segments) and every non-mention link are handed straight back to the Link mark and render exactly as they do today.

### How the interception works

`parseMentionHref` (`frontend/src/lib/mentions.ts`) parses a `mention://` href into a discriminated `MentionTarget` union, dispatched with a `switch` on the path segment, so a future `mention://user/<uuid>` is one extra variant and one extra `case`. It never throws.

A new inline atom node (`frontend/src/components/tiptap-node/mention-node/mention-node.tsx`) claims anchors whose href parses as a mention. No raw comment text is ever re-parsed, and `@tiptap/extension-mention` is not installed — this is href interception, not editor mention machinery.

Two things worth flagging for review:

1. **`@tiptap/markdown@3.7.1` does not route markdown through HTML.** It runs `marked`'s lexer and dispatches tokens to extension `parseMarkdown` handlers keyed by `markdownTokenName`; `parseHTML` rules are only consulted for literal HTML embedded in the markdown. So an `a[href]` parse rule alone never fires for a case comment. The node therefore intercepts on the already-parsed href in **both** pipelines: the markdown `link` token (how comments actually load) and the `a[href]` HTML rule (inline HTML and `setContent` with HTML). Because only the first registered handler per token type is consulted, the markdown handler reproduces `@tiptap/extension-link`'s own `parseMarkdown` verbatim (including `title`) for ordinary links.
2. **`Mention` must be registered before `StarterKit`.** `MarkdownManager` is built from the raw, declaration-ordered extension array rather than the priority-sorted one, so extension `priority` does not decide markdown handler order. There is a comment at the call site explaining this, and the integration test fails loudly if it regresses. Relatedly, TipTap does not propagate extension `priority` into ProseMirror parse rules, so the HTML rule carries its own `priority: 1000` to outrank the Link mark's default of 50.

`renderHTML` / `renderText` / `renderMarkdown` round-trip the node back to `[@Label](mention://agent/<id>)` unchanged.

## Rendering scope decision: read-only viewers only

Chips render **only where a consumer opts in**, via a new `enableMentions` prop on `SimpleEditor` (default `false`). `CaseCommentViewer` is the only consumer that sets it.

The alternative — chips everywhere a `mention://` link appears — was rejected because `SimpleEditor` is also the editable case description editor and comment composer. Swapping a link for an atom node in an editable surface would make the token uneditable in place: an author could no longer fix a typo in the label or retarget the mention without deleting the whole chip. Editable surfaces keep the raw markdown link, which stays selectable and editable; read-only viewers get the chip. This also keeps the blast radius of the change to exactly one call site.

## Styling

The chip reuses the existing badge/chip visual language (`inline-flex`, `rounded-md`, `border`, `px-1.5`, `text-xs font-medium`) with `align-baseline` and `items-baseline` so it does not disturb line height inside a comment paragraph or a thread reply. Colors are semantic tokens only (`bg-muted`, `text-foreground`, `border-border`, `text-muted-foreground`), so light and dark both work with no hardcoded values. The "unavailable" treatment reuses the dashed-border/muted language already used by `AttachmentImagePlaceholder`. No icon was added: the app has no consistent agent-preset icon (the sidebar uses `MousePointerClickIcon`, the canvas uses `BotIcon`), so inventing one here would have been the inconsistent choice.

Preset data loads async. While it is in flight the chip shows the embedded label in the normal treatment (no skeleton flash), then swaps to the resolved name.

## Test plan

Jest + Testing Library. Note the issue said vitest; this repo's frontend runner is Jest (`frontend/jest.config.cjs`, `pnpm -C frontend test` → `jest`).

- `frontend/src/lib/mentions.test.ts` — `parseMentionHref` across valid lower/uppercase UUIDs, https and relative urls, non-UUID and numeric ids, `mention://user/<uuid>`, `mention://foo/<uuid>`, missing/empty id, extra segments, missing segment, bare `mention://`, wrong scheme, empty string, `null`, `undefined`, plus a never-throws sweep over hostile inputs (path traversal, `%%%`, NUL, 10k chars, leading whitespace, uppercase segment).
- `frontend/src/components/tiptap-node/mention-node/mention-node.test.tsx` — chip states with the presets hook mocked. Resolved renders `@<current name>` and the stale label is asserted **absent**. Unknown id and query error both produce the unavailable treatment with `title`. Loading (and `presets === undefined`) renders the label with `data-state="loading"` and no `title`.
- `frontend/src/components/cases/case-description-editor.test.tsx` — real `CaseCommentViewer` over real markdown: a valid mention renders a chip with the resolved name and no stale label; `mention://agent/not-a-uuid` and `mention://user/<uuid>` both stay plain `<a>` elements with their original text and no chip; an ordinary `https://` link is the only anchor in the doc and keeps its href. Every case asserts `console.error` was never called.

Verification, all clean:

```
pnpm -C frontend exec biome check --write .   # 947 files, no fixes applied
pnpm -C frontend check                        # 947 files, no fixes applied
pnpm -C frontend run typecheck                # clean
pnpm -C frontend test                         # 101 suites, 581 tests passed
```

### `jest.config.cjs` changes

The integration test is the first test in the repo to mount `SimpleEditor`, which surfaced three latent config problems. All three are fixes, not workarounds, and the full suite passes with them:

- `marked` and `react-hotkeys-hook` (both ESM-only, reached transitively) now get transformed. The `transformIgnorePatterns` regex gained an optional `.pnpm/` hop — under pnpm every real package path contains a second `node_modules/` segment, so the previous pattern could never un-ignore anything.
- The `.ts/.tsx` and `.js/.mjs` transforms are merged into a single `ts-jest` entry. Two separate `ts-jest` entries make ts-jest apply one options set globally, which silently broke JSX compilation suite-wide.
- `moduleNameMapper` is reordered so the stylesheet mock precedes the `^@/(.*)$` alias. Only the first match applies, so `@/…/x.scss` imports were resolving to real `.scss` files.

No dependency or lockfile changes.

## Out of scope

Chips are inert in v1 — no click behavior, no navigation, no hover card. The comment composer and case description editor are behaviorally unchanged.

Closes ENG-1601
https://linear.app/tracecat/issue/ENG-1601

## LOC breakdown

| Category | + | - |
| --- | --- | --- |
| Logic | 295 | 1 |
| Tests | 337 | 0 |
| Infra/config | 16 | 6 |
| **Total** | **648** | **7** |

### DIFF
--- a/frontend/jest.config.cjs
+++ b/frontend/jest.config.cjs
@@ -3,25 +3,35 @@ module.exports = {
   setupFilesAfterEnv: ["<rootDir>/jest.setup.js"],
   testEnvironment: "jsdom", // For React components
   transform: {
-    "^.+\\.tsx?$": [
+    // `.js`/`.mjs` are included because `marked` (pulled in by
+    // @tiptap/markdown) ships ESM only and has to be downleveled to CommonJS
+    // before Jest can require it. ts-jest only honours one set of options per
+    // run, so this must stay a single entry.
+    "^.+\\.m?[jt]sx?$": [
       "ts-jest",
       {
         tsconfig: {
           jsx: "react-jsx",
+          allowJs: true,
         },
       },
     ],
   },
   testPathIgnorePatterns: ["/node_modules/", "/.next/", "/tests/smoke/"],
-  transformIgnorePatterns: ["node_modules/(?!(yaml)/)"],
+  // The optional `.pnpm/` hop makes this work with pnpm's nested store layout,
+  // where every real package path contains `node_modules/.pnpm/<name>@<ver>/`.
+  transformIgnorePatterns: [
+    "node_modules/(?!(\\.pnpm/)?(yaml|marked|react-hotkeys-hook)([@/]|$))",
+  ],
   moduleNameMapper: {
-    "^@/(.*)$": "<rootDir>/src/$1",
-    "^lucide-react/dynamicIconImports$":
-      "<rootDir>/tests/mocks/lucide-dynamic-icon-imports.js",
-    // Mock CSS and other non-JS files
+    // Only the first matching pattern applies, so the asset mocks have to come
+    // before the `@/` alias: stylesheets are imported through it too.
     "\\.(css|less|scss|sass)$": "identity-obj-proxy",
     "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$":
       "jest-transform-stub",
+    "^@/(.*)$": "<rootDir>/src/$1",
+    "^lucide-react/dynamicIconImports$":
+      "<rootDir>/tests/mocks/lucide-dynamic-icon-imports.js",
     // Mock yaml module for Jest
     "^yaml$": "identity-obj-proxy",
   },
--- a/frontend/src/components/cases/case-description-editor.test.tsx
+++ b/frontend/src/components/cases/case-description-editor.test.tsx
@@ -86,6 +86,17 @@ describe("CaseCommentViewer mentions", () => {
     expect(consoleErrorSpy).not.toHaveBeenCalled()
   })
 
+  it("resolves an uppercase uuid href against the lowercase preset id", async () => {
+    renderViewer(
+      `Please review [@Stale Label](mention://agent/${PRESET_ID.toUpperCase()}).`
+    )
+
+    const chip = await screen.findByTestId("mention-chip")
+    expect(chip).toHaveTextContent("@Triage Agent")
+    expect(chip).toHaveAttribute("data-state", "resolved")
+    expect(consoleErrorSpy).not.toHaveBeenCalled()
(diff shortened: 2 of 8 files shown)


aaron-tracecat:
@codex review

PR #3213, opened 2026-08-07
feat(cases): parse and persist case comment mentions

aaron-tracecat:
## Summary

Implements the "fact half" of the agent-comment-mentions feature: parses `@mention` tokens out of case comment content and persists them as immutable rows at comment-creation time. A later task (ENG-1603) will add dispatch/notification/gating logic on top of these persisted facts — none of that is included here.

- New `case_comment_mention` table (`tracecat/db/models.py`), registered in the RLS-managed workspace table list (`tracecat/db/tenant_rls.py`).
- New pure parser module `tracecat/cases/mentions.py` for the token syntax `[@<label>](mention://agent/<uuid>)`. **This is the only place in the codebase that understands the token encoding** — it is intentionally quarantined there because the encoding is pending team review and must stay swappable.
- `CaseCommentsService.create_comment` parses mentions, deduplicates by `(target_type, target_id)`, batch-validates `agent` targets against live (non-soft-deleted) `AgentPreset` rows in the workspace, and inserts mention rows in the same transaction as the comment. Unknown target types and unresolved/soft-deleted agent targets are inert (no row, no error).
- `update_comment` is untouched — mentions are create-only, immutable event records; edits never add/remove/update mention rows.
- `CaseCommentRead` gains a `mentions: list[CaseCommentMentionRead]` field, populated via a batched (not per-comment) query across `list_comments`, `list_comment_threads`, `get_comment_thread`, and the internal router's comment create/update responses.

## Migration

Autogenerated with `uv run alembic revision --autogenerate` against the local dev database, then hand-tuned: stripped unrelated schema drift picked up by autogenerate, and added the `enable_workspace_table_rls` / `disable_workspace_table_rls` calls (matching the pattern used by other workspace-scoped table migrations) since RLS enablement isn't captured by autogenerate. Verified with a full upgrade/downgrade/upgrade round-trip against the dev database, and confirmed a follow-up autogenerate produces no further diff for this table.

## Test plan

- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run basedpyright --warnings --threads 4` — 0 errors, 0 warnings
- `uv run pytest tests/unit -k "mention or comment" -x` — 81 passed
- `uv run pytest tests/unit -k "mention or comment" -n 2` — 81 passed, stable distributed collection
- `pnpm -C frontend exec biome check --write .`
- `pnpm -C frontend run typecheck`
- Manual: `alembic upgrade head` / `downgrade -1` / `upgrade head` round-trip against the local dev database; confirmed RLS policy present via `pg_class.relrowsecurity` and `\d case_comment_mention`.

New/extended coverage:
- `tests/unit/test_case_comment_mentions.py`: parser tests (valid token, malformed variants, duplicate tokens preserved, multiple distinct targets, token adjacent to surrounding text, unicode labels). All test inputs use hardcoded, named UUID constants (no `uuid.uuid4()`) so distributed (pytest-xdist) collection is deterministic.
- `tests/unit/test_case_comments_service.py`: valid mention persistence (incl. batched hydration across list/thread/tombstone reads), dedupe, multiple targets, inert-on-unknown/soft-deleted target, immutability across edits, and FK cascade on hard comment delete. Uses a deterministic per-preset slug and a hardcoded unresolved-target UUID constant instead of `uuid.uuid4()`.

## LOC breakdown

| Category | + | - |
|---|---|---|
| Logic | 229 | 6 |
| Tests | 354 | 2 |
| Generated | 149 | 0 |

Closes ENG-1602
https://linear.app/tracecat/issue/ENG-1602

### DIFF
--- a/tracecat/cases/service.py
+++ b/tracecat/cases/service.py
@@ -42,8 +42,9 @@ from tracecat.cases.enums import (
     CaseSeverity,
     CaseStatus,
     CaseTaskStatus,
+    MentionTargetType,
 )
-from tracecat.cases.mentions import parse_mentions
+from tracecat.cases.mentions import MentionToken, parse_mentions
 from tracecat.cases.schemas import (
     AssigneeChangedEvent,
     CaseBatchItemResult,
@@ -2494,8 +2495,8 @@ class CaseCommentsService(BaseWorkspaceService):
         content: str,
     ) -> None:
         """Persist live, workspace-local mention targets for a new comment."""
-        unique_tokens = []
-        seen_targets: set[tuple[str, uuid.UUID]] = set()
+        unique_tokens: list[MentionToken] = []
+        seen_targets: set[tuple[MentionTargetType, uuid.UUID]] = set()
         for token in parse_mentions(content):
             target_key = (token.target_type, token.target_id)
             if target_key in seen_targets:
@@ -2503,8 +2504,10 @@ class CaseCommentsService(BaseWorkspaceService):
             seen_targets.add(target_key)
             unique_tokens.append(token)
 
-        agent_ids = {
-            token.target_id for token in unique_tokens if token.target_type == "agent"
+        agent_ids: set[uuid.UUID] = {
+            token.target_id
+            for token in unique_tokens
+            if token.target_type == MentionTargetType.AGENT
         }
         if not agent_ids:
             return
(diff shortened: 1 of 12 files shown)


daryllimyt:
@codex review

tracecat/cases/service.py
```diff
@@ -2414,6 +2487,50 @@ async def _comment_has_replies(self, comment_id: uuid.UUID) -> bool:
...
         return bool(result.scalar())
 
+    async def _persist_comment_mentions(
+        self,
+        *,
+        comment: CaseComment,
+        content: str,
+    ) -> None:
+        """Persist live, workspace-local mention targets for a new comment."""
+        unique_tokens = []
```
daryllimyt: basedpyright infers this as list[Unknown], explicitly type

daryllimyt (changes requested):
lgtm just one typing change

tracecat/cases/service.py
aaron-tracecat: done

daryllimyt (approved):

PR #3215, opened 2026-08-09
fix(engine): collapse S3 I/O onto one event loop

daryllimyt:
## Summary

- Route worker blob-storage I/O through one process-owned `AppAsyncRuntime` running uvloop.
- Reuse one loop-affine aiobotocore client and connector pool while keeping serialization, hashing, parsing, and expression work on Temporal activity threads.
- Preserve cancellation and context propagation across sync and async bridges, drain accepted submissions before S3 cleanup, and shut the worker down if the application loop dies.
- Add focused lifecycle, concurrency, storage-cache, and Linux Temporal/MinIO load coverage.

## Root cause

Synchronous Temporal activities previously created persistent thread-local asyncio runners. Each runner owned a separate event loop and therefore a separate loop-bound S3 client and connector pool. At worker concurrency, this multiplied threads, file descriptors, and retained resources, while recycled threads could leave clients bound to closed loops.

The new worker-owned application runtime gives loop-affine I/O resources one explicit home without moving CPU-heavy payload work onto the Temporal event loop.

## Impact

Workers now share one S3 I/O loop and bounded connector pool across activity threads. Normal shutdown drains Temporal work and the activity executor before closing loop-owned storage resources. If the application loop stops unexpectedly, the watchdog requests graceful worker shutdown so the process can be restarted instead of consuming task slots while storage activities fail indefinitely.

## Validation

- `uv run ruff check .`
- `uv run ruff format --check` on changed runtime and test files
- `uv run basedpyright --warnings` on changed runtime and test files: 0 errors, 0 warnings
- Focused storage/runtime unit suite: 140 passed
- Runtime-death race tests repeated 20 times each: 40 passed
- Pre-commit hooks passed
- Linux-only Temporal/MinIO load test added; skipped locally on macOS

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 1,375 | 225 |
| Tests | 1,652 | 48 |

## Related

Resolves [ENG-1606](https://linear.app/tracecat/issue/ENG-1606/fixworker-collapse-s3-access-onto-one-app-event-loop).

---
## Summary by cubic
Collapse all worker blob-storage I/O onto one process-owned async runtime using `uvloop`, reusing a single loop-bound `aiobotocore` S3 client and a bounded connector pool. Improves performance, reduces FD/thread churn, and hardens startup, shutdown, and cancellation behavior. Resolves ENG-1606.

- **Refactors**
[...]

PR #3217, opened 2026-08-09
fix(expressions): cache jsonpath parser to fix import-lock deadlock

daryllimyt:
## Tracking

- [ENG-1608: Prevent JSONPath parser races under multithreaded evaluation](https://linear.app/tracecat/issue/ENG-1608/prevent-jsonpath-parser-races-under-multithreaded-evaluation)

## Problem

`jsonpath_ng.ext.parse(expr)` constructs a brand-new PLY parser on **every call**. jsonpath_ng ships no pre-generated parsetab module and passes `write_tables=0`, so each call:

1. attempts `import parser_jsonpath_parsetab`, which always fails and therefore goes through `importlib._find_and_load` — taking Python's per-module import lock — on every single expression evaluation, and
2. falls back to regenerating the full LALR parse table (~7 ms of pure-Python CPU per call).

Since `eval_jsonpath` sits on the expression-evaluation hot path, every workflow expression pays this tax. Worse, under many concurrent Temporal sync-activity threads the import-lock traffic is a deadlock hazard: threads queued on the import lock cannot heartbeat, activities hit their heartbeat timeout, and the SDK cancels them by injecting exceptions into the threads. If an injection lands while a thread is inside the import critical section, the import lock leaks permanently — threads blocked in C-level `lock.acquire()` can never receive injected exceptions — and the worker wedges irrecoverably.

This is not theoretical: the temporal CI job for #3215 deadlocked exactly this way (61 of 64 activity threads parked on the leaked import lock in the faulthandler dump; the job hung until its 60-minute timeout).

## Fix

Build one `ExtentedJsonPathParser` per parser module at module import time, removing parser construction and parse-table generation from the expression hot path. Serialize cache misses with a parser lock because PLY's `LRParser` mutates instance state during parsing and is not thread-safe. Current callers treat cached expression trees as read-only, so they can safely share parsed results.

The process-local cache deliberately uses a small manual LRU rather than `functools.lru_cache`:

- Cache hits take only the cache lock, so they do not wait behind an unrelated slow parse.
- A miss acquires the parser lock and rechecks the cache, coalescing concurrent cacheable misses for the same expression into one parse.
- Each module retains at most 1024 expressions, and only expressions up to 256 characters are admitted.
- Longer expressions remain valid and parse normally without being retained; the limit does not change JSONPath validation semantics.

Applied in three places:

- `tracecat/expressions/common.py` — shared `parse_jsonpath()`, used by `eval_jsonpath`
- `tracecat/expressions/validator/base.py` — validator reuses `parse_jsonpath()`
- `packages/tracecat-registry/tracecat_registry/_internal/jsonpath.py` — standalone copy of the same parser and cache behavior because the registry package cannot import `tracecat.*`

## Verification

- Parser-cache regression suite covering both implementations: 10 passed
- Focused core expression and parsing tests: 18 passed
- Focused registry JSONPath tests: 27 passed
- Focused validator JSONPath test: 1 passed
- Concurrent cold-miss coverage verifies one parse across eight simultaneous callers
- Oversized-expression coverage verifies valid paths still parse without entering the cache
- `uv run ruff check`, `uv run ruff format --check`, and targeted `uv run basedpyright`: clean
- Signed commit hooks: passed

## LOC breakdown

| Category | + | - |
|---|---:|---:|
| Logic | 130 | 7 |
| Tests | 189 | 0 |

### DIFF
--- a/packages/tracecat-registry/tracecat_registry/_internal/jsonpath.py
+++ b/packages/tracecat-registry/tracecat_registry/_internal/jsonpath.py
@@ -4,14 +4,34 @@ This module provides a lightweight JSONPath evaluation function that avoids
 importing heavy tracecat modules during SDK-style invocation.
 """
 
+import threading
+from functools import lru_cache
 from typing import Any
 
-import jsonpath_ng.ext
 import jsonpath_ng.jsonpath as jsonpath_nodes
 from jsonpath_ng.exceptions import JsonPathParserError
+from jsonpath_ng.ext.parser import ExtentedJsonPathParser
 
 from tracecat_registry._internal.exceptions import TracecatExpressionError
 from tracecat_registry._internal.logger import logger
+
+# jsonpath_ng builds a new PLY parser per parse() call, and without a shipped
+# parsetab module every call re-imports the missing table module and regenerates
+# the LALR table. That import-lock traffic can deadlock the process if a Temporal
+# activity thread is cancelled while holding the import lock. Build one parser at
+# import time (before any worker threads exist) and serialize access: PLY's
+# LRParser mutates shared state during parse and is not thread-safe. Parsed
+# expression trees are immutable, so caching and sharing them across threads is
+# safe.
+_JSONPATH_PARSER = ExtentedJsonPathParser()
+_JSONPATH_PARSER_LOCK = threading.Lock()
+
+
+@lru_cache(maxsize=4096)
+def _parse_jsonpath(expr: str) -> jsonpath_nodes.JSONPath:
+    """Parse a jsonpath expression using the shared process-wide parser."""
+    with _JSONPATH_PARSER_LOCK:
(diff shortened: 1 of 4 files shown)


daryllimyt:
* **#3215** <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/3215?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3217** <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/3217?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TracecatHQ/tracecat/3217?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aaron-tracecat (approved):

PR #3231, opened 2026-08-11
feat(cases): show agent activity in comment threads

aaron-tracecat:
## Summary

https://github.com/user-attachments/assets/51f66390-0b1c-4f67-a7d8-f047c4a9fc24

- Show a separate pending, running, or failed state for every agent mentioned in a case comment.
- Poll comments only while an invocation is active, then show the finished reply with the saved agent name and avatar.
- Let users open the invocation's existing agent session from the comment thread through the case chat sidebar.
- Keep attribution stable even if the live agent preset is later renamed or deleted.

## Why

Mentioning an agent previously gave very little feedback. Users could not easily tell whether the agent was starting, thinking, finished, or failed, and completed replies looked like generic system comments.

This makes the entire lifecycle visible without adding a second transcript UI.

## Design

1. Comment reads include each mention's invocation state and completed replies include snapshotted agent attribution.
2. The frontend polls once per second only while at least one invocation is pending or running.
3. Pending, running, and failed invocations render directly below the source comment; succeeded status rows disappear when the attributed reply arrives.
4. “View session” selects the existing agent session with `chatId=<session_id>` and opens the case chat sidebar.
5. Shared query keys keep flat comments and threaded comments synchronized.

## Steps to QA

1. Mention one or more agents in a case comment.
2. Confirm each invocation independently shows preparing and thinking states.
3. Open “View session” and confirm the matching case chat session opens.
4. Confirm success removes the temporary state and shows an agent-attributed reply.
5. Confirm failure stays visible with the server error and retry guidance.

## Validation

- Focused backend service coverage for invocation state and attribution across comment read surfaces.
- Frontend coverage for lifecycle rendering, conditional polling, query invalidation, attribution, and session navigation.
- Commit hooks, including Ruff, Python type checking, generated-client verification, Biome, and frontend type checking.

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 451 | 74 |
| Tests | 855 | 2 |
| Generated | 180 | 0 |
## Stack

- Base: #3227
- This PR: #3231
- Follow-up: #3232

Linear: [ENG-1604](https://linear.app/tracecat/issue/ENG-1604/featui-agent-invocation-status-and-attribution-in-comment-threads)

### DIFF
--- a/frontend/src/client/schemas.gen.ts
+++ b/frontend/src/client/schemas.gen.ts
@@ -6464,6 +6464,63 @@ export const $CaseBatchUpdate = {
   description: "Request body for updating multiple cases.",
 } as const
 
+export const $CaseCommentAgentInvocationRead = {
+  properties: {
+    id: {
+      type: "string",
+      format: "uuid",
+      title: "Id",
+    },
+    preset_name: {
+      type: "string",
+      title: "Preset Name",
+    },
+    preset_slug: {
+      type: "string",
+      title: "Preset Slug",
+    },
+    status: {
+      $ref: "#/components/schemas/CaseCommentAgentInvocationStatus",
+    },
+    session_id: {
+      anyOf: [
+        {
+          type: "string",
+          format: "uuid",
+        },
+        {
+          type: "null",
+        },
+      ],
+      title: "Session Id",
+    },
+    error: {
+      anyOf: [
+        {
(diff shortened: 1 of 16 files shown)


daryllimyt (approved):
lgtm

PR #3252, opened 2026-08-14
fix(sandbox): raise NSTUN UDP capacity to 2048

daryllimyt:
## What changed

- Raise the shared parent UDP/ICMP socket budget from 1,024 to 2,048 per NSTUN sandbox.
- Give UDP its own 2,048-flow capacity while keeping TCP and ICMP flow tables at 1,024.
- Raise the FD lookup and monitor-handler ceilings to 4,096 so they do not reject flows before the socket budget does.
- Extend the exact-limit Docker regression to open 2,064 UDP destinations and verify exactly 2,048 are accepted.
- Add a production-path agent regression that starts 12 separate cold `uvx` stdio MCP servers, drives 128 acknowledged startup mappings per server (1,536 total), and requires every server to complete MCP tool discovery.
- After MCP initialization, make Claude invoke its real Bash tool inside the same jail. Bash opens and acknowledges another 256 UDP mappings, and the test requires its tool result to reach the second model turn (1,792 combined flows).
- Verify Bash is exposed locally and WebFetch/WebSearch are available as deferred provider-side tools without counting them as jailed NSTUN traffic.
- Let the agent Docker fallback create its Linux TUN device in a privileged test container instead of skipping when the macOS host has no `/dev/net/tun`.

## Why

MCP startup can create a short, repeatable burst of DNS/UDP flows. NSTUN retains those parent mappings long enough for the burst to fill the current reservation budget, causing subsequent resolver attempts and MCP connections to fail.

Raising only the reservation counter is insufficient: the UDP flow arrays, numeric FD lookup, and monitor handler table are independent ceilings. This changes them together while preserving hard overflow behavior at 2,048 parent UDP/ICMP sockets per active sandbox.

The synthetic limit test proves the hard boundary. The agent scenario covers the missing production chain: agent executor, broker shim, Claude CLI, concurrent cold `uvx` processes, MCP initialize/tool discovery, a real Bash tool round trip under retained MCP load, and the shared NSTUN parent.

WebFetch and WebSearch execute through the model provider rather than inside NsJail, so they do not consume this sandbox's NSTUN reservations. The regression verifies that they remain selectable but does not manufacture fake NSTUN flows for them.

## Capacity and risk

- Each active sandbox may now consume up to 2,048 host-side UDP/ICMP sockets; host exposure still scales with concurrent sandboxes.
- The combined agent scenario reaches 1,792 mappings and leaves 256 mappings of headroom below the new hard cap.
- TCP and ICMP fixed flow-table counts remain unchanged.
- The bounded-memory patch still enforces the 8 MiB NSTUN context assertion.
- Attempts beyond the UDP flow cap are still dropped rather than growing without bound.

## Validation

- `uv run pytest --confcutdir=tests/unit tests/unit/test_executor_sandbox_nsjail.py::test_nstun_bounds_parent_udp_sockets -x -vv`
- `uv run pytest --confcutdir=tests/unit tests/unit/test_agent_sandbox_litellm.py::test_agent_nsjail_handles_stdio_mcp_startup_and_bash_network_burst -x -vv`
- `uv run pytest --confcutdir=tests/unit tests/unit/test_agent_sandbox_litellm.py::test_run_agent_activity_spawns_full_claude_harness_runtime_in_each_sandbox_mode -k direct -x -vv`
- Red/green control: the 1,536-flow MCP phase failed against `origin/main` at the old 1,024 UDP-flow ceiling. This branch passes all 12 MCP initializations plus the 256-flow Bash call (1,792 total).
- `uv run ruff check tests/unit/test_agent_sandbox_litellm.py tests/unit/test_executor_sandbox_nsjail.py`
- `uv run ruff format --check tests/unit/test_agent_sandbox_litellm.py tests/unit/test_executor_sandbox_nsjail.py`
- `uv run basedpyright tests/unit/test_agent_sandbox_litellm.py tests/unit/test_executor_sandbox_nsjail.py`
- Applied the patch to the pinned NsJail source and built the Docker image successfully.

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Infra/config | 114 | 12 |
| Tests | 532 | 13 |

---
## Summary by cubic
Raises NSTUN UDP capacity to 2048 to prevent DNS/UDP bursts during MCP startup from exhausting the parent socket budget and causing resolver/MCP failures. Old: UDP/ICMP parents capped at 1024 with lower FD/handler ceilings; New: UDP has its own 2048-flow cap with `NSTUN_MAX_FDS` and monitor handler ceilings at 4096; overflow still hard-drops.

- Introduces `NSTUN_MAX_UDP_FLOWS=2048`; TCP/ICMP tables remain 1024; `UDP_ICMP_SOCKET_BUDGET=2048`.
- Raises `MAX_HANDLERS` and `NSTUN_MAX_FDS` to 4096 so the socket budget governs rejection.
- Extends Docker NSTUN limit test to open 2,064 UDP destinations, assert exactly 2,048 accepted, and set container `ulimits.nofile` to 4096; also asserts no monitor handler overflow.
- Adds a production-path agent test: starts 12 cold `uvx` stdio MCP servers (1,536 UDP startup probes), then executes a Bash probe for 256 more flows (total 1,792 < 2,048); verifies all servers initialize and Bash completes; confirms web tools are listed without inflating jailed flow counts. New CLI: `--run-nsjail-stdio-mcp-burst-smoke`.
- Docker fallback now runs privileged to create `/dev/net/tun` and can set `nofile=4096`; tests allow UDP egress to the parent via a targeted network policy override.
- Memory bounds unchanged; excess flows beyond the UDP cap still drop without unbounded growth.

**Review/rollout**
- Confirm limits align in `monitor.cc`, `nstun/core.h`, and `nstun/udp.cc`: `MAX_HANDLERS=4096`, `NSTUN_MAX_FDS=4096`, `NSTUN_MAX_UDP_FLOWS=2048`, `UDP_ICMP_SOCKET_BUDGET=2048`.
- Ensure environments running NSTUN or Docker fallbacks allow privileged containers, have `/dev/net/tun`, and set `nofile` ≥ 4096.

<sup>Written for commit 4c4193483bd2fe1ce8641d08fdeec16811a3090b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3252?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docker/nsjail/nstun-bounded-memory.patch
+++ b/docker/nsjail/nstun-bounded-memory.patch
@@ -1,5 +1,17 @@
+diff --git a/monitor.cc b/monitor.cc
+--- a/monitor.cc
++++ b/monitor.cc
+@@ -83,7 +83,7 @@ struct fdHandler_t {
+ /* --- per-thread context -------------------------------- */
+ 
+ constexpr size_t MAX_EVENTS = 16;
+-constexpr size_t MAX_HANDLERS = 2048;
++constexpr size_t MAX_HANDLERS = 4096;
+ constexpr size_t MAX_PERIODICS = 16;
+ constexpr int kSetupTimeoutSec = 10;
+ constexpr int kShutdownEscalationSec = 2;
 diff --git a/nstun/core.h b/nstun/core.h
-index 0e033ff..3be3bdf 100644
+index 0e033ff..4999e14 100644
 --- a/nstun/core.h
 +++ b/nstun/core.h
 @@ -11,6 +11,7 @@
@@ -10,7 +22,15 @@ index 0e033ff..3be3bdf 100644
  
  #include "net_defs.h"
  #include "nstun.h"
-@@ -25,11 +26,24 @@ constexpr size_t UDP_QUEUE_PACKET_MAX = 1500;
+@@ -19,17 +20,31 @@
+ namespace nstun {
+ 
+ constexpr size_t NSTUN_MAX_FLOWS = 1024;
+-constexpr size_t NSTUN_MAX_FDS = 2048;
++constexpr size_t NSTUN_MAX_UDP_FLOWS = 2048;
++constexpr size_t NSTUN_MAX_FDS = 4096;
+ constexpr size_t NSTUN_MAX_RULES = 128;
+ constexpr size_t UDP_QUEUE_PACKET_MAX = 1500;
  constexpr size_t UDP_QUEUE_MAX_PACKETS = 50;
  constexpr int VLEN = 64;
  
@@ -33,12 +53,12 @@ index 0e033ff..3be3bdf 100644
 +/* UDP and ICMP sockets are owned by the parent NSTUN process and are not
 + * constrained by the jailed child's RLIMIT_NOFILE. Bound their aggregate
 + * kernel socket memory separately while leaving headroom for DNS bursts. */
-+constexpr size_t UDP_ICMP_SOCKET_BUDGET = 1024;
++constexpr size_t UDP_ICMP_SOCKET_BUDGET = 2048;
 +static_assert(UDP_ICMP_SOCKET_BUDGET < NSTUN_MAX_FDS);
  
  
  /* Removed MemcmpLess in favor of C++20 operator<=> */
-@@ -62,6 +75,22 @@ struct __attribute__((packed)) IcmpFlowKey6 {
+@@ -62,6 +77,22 @@ struct __attribute__((packed)) IcmpFlowKey6 {
  
  struct Context;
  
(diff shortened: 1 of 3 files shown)


daryllimyt:
@codex review

jordan-umusu (approved):
lgtm

PR #3255, opened 2026-08-17
feat(cases): use display names for custom fields

aaron-tracecat:
## Summary

Updates the frontend to treat a case field's `display_name` as its user-facing label while keeping `name` / `id` as the machine reference.

- reads `display_name` across case configuration, case details, lists, activity, durations, and workflow-trigger suggestions
- writes both the derived reference and display name on creation
- lets users edit the display name and reference independently
- preserves raw references as a fallback when metadata is unavailable, such as for a deleted field in historical activity

Linear: [ENG-1646](https://linear.app/tracecat/issue/ENG-1646/update-frontend-to-read-and-write-display-name-via-the-new-schema)

## Why

The new schema separates two concepts that the frontend previously treated as one:

- **Display name:** mutable, human-readable text shown in the UI
- **Reference:** technical identifier used as the database column, API key, workflow input, and expression reference

Using the reference as visible text exposed implementation details such as `analyst_verdict_v2`. Automatically deriving and submitting a new reference during a display-name edit was more serious: it could rename the underlying column and invalidate existing consumers without the user explicitly choosing to do so.

### Creation and editing

- The create dialog accepts a friendly name, derives a snake_case reference, previews it, and submits both `name` and `display_name`.
- References beginning with a number receive a deterministic `field_` prefix; for example, `2FA status` becomes `field_2fa_status`.
- Generated reference collisions use the first available `_2`, `_3`, … suffix while preserving the API length limit.
- Names that cannot produce a usable reference receive a targeted validation error.
- The edit dialog exposes separate **Name** and **Reference** inputs:
  - changing only **Name** updates `display_name` and omits `name` from the request
  - changing **Reference** explicitly sends `name` and uses the existing backend rename path
  - reference edits use the API's identifier and length constraints and include a warning about workflow/API compatibility
- The custom-fields table is ordered as **Field name → Data type → Reference**.

### Display surfaces

Case field labels now prefer `display_name` in:

- custom-field settings, delete confirmations, and closure requirements
- case tables, column selectors, accordions, and case cards
- case detail fields, URL fields, and closure dialogs
- duration definitions and activity-feed field-change events
- workflow-trigger case-value suggestions, including friendly-name search

Duration field selectors show the stable reference as secondary text, keeping fields distinguishable when they share a display name. Historical or stale references fall back to the raw identifier so the UI remains readable if current field metadata no longer exists.

### Data plumbing

- Carries display names alongside references in case-list field metadata.
- Adds a shared reference-to-display-name lookup helper for surfaces that store only references.
- Centralizes the case-field cache invalidations shared by create, edit, and delete.
- Extends the registry case-field type with optional `display_name` support.

No database migration or backend behavior change is included.

## Test strategy

Coverage is intentionally concentrated in a few deeper, behavior-level flows rather than one unit test per display surface:

- create a field through the rendered dialog and verify collision-safe reference derivation plus the API payload
- edit through the rendered dialog, proving display-name changes do not rename the reference and explicit reference changes are validated before submission
- render the settings table and representative case surfaces to verify friendly labels, column order, and stable references
- render the duration dialog with duplicate display names and verify both references remain visible

Existing surface tests use display-name-aware fixtures so their normal behavior continues to exercise the shared metadata path without duplicating the same assertion across every component.

## Review guide

1. Start with `case-field-display.ts` and the add/edit dialogs to verify the display-name/reference contract.
2. In `custom-field-metadata-dialogs.test.tsx`, compare the display-name-only request (no `name`) with the validated explicit-reference request (includes `name`).
3. Review the case-list metadata flow in `cases-layout.tsx`, `case-item.tsx`, and `cases-accordion.tsx`.
4. Check the reference-only display surfaces: activity feed, duration summaries, and workflow-trigger suggestions.
5. Review `case-duration-dialog.test.tsx` and `custom-fields-table.test.tsx` for duplicate-name disambiguation and the visible Name/Data type/Reference contract.

## Validation

- `pnpm -C frontend test` — 128 suites, 936 tests passed
- frontend TypeScript check passed
- Biome check passed across 994 files
- full pre-commit suite passed, including secret scanning, frontend formatting, and frontend typing
- `git diff --check` passed

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 307 | 117 |
| Tests | 287 | 7 |


frontend/src/components/cases/case-duration-dialog.tsx
```diff
@@ -402,7 +402,7 @@ export function CaseDurationDialog({
       .filter((f) => !f.reserved)
       .map((f) => ({
         value: f.id,
-        label: f.id,
+        label: f.display_name,
```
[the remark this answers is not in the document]
aaron-tracecat: Fixed in `429c82c25`. Duration field options now keep `display_name` as the primary label and show the stable field reference as secondary text, so fields with duplicate display names remain distinguishable. Added a focused rendered-dialog regression test with two same-name fields; the full frontend suite passes (128 suites, 936 tests).

frontend/src/components/cases/case-panel-fields-group.tsx
```diff
@@ -55,7 +55,7 @@ export function CasePanelFieldsGroup({
             group in `case-panel-view.tsx`. */}
         <div className="flex flex-col gap-1">
           {visibleCustomFields.map((field) => {
-            const label = field.id
+            const label = field.display_name
```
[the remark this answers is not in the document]
aaron-tracecat: users will have to know the impact of creating two fields with duplicate display names

frontend/src/components/cases/closure-requirements-view.tsx
```diff
@@ -217,7 +217,7 @@ export function ClosureRequirementsView() {
     ...customFields.map(
       (f): ClosureRow => ({
         id: f.id,
-        name: f.id,
+        name: f.display_name,
```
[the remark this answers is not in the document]
aaron-tracecat: users will have to know the impact of creating two fields with duplicate display names

frontend/src/components/cases/edit-custom-field-dialog.tsx
```diff
@@ -391,16 +405,33 @@ export function EditCustomFieldDialog({
...
+                    Human-readable label shown throughout the product.
+                  </FormDescription>
+                  <FormMessage />
+                </FormItem>
+              )}
+            />
+
+            <FormField
+              control={form.control}
+              name="reference"
```
jordan-umusu, line 420: Can we add collision handling when generating the reference on create? Create and edit currently have different effective contracts: I can’t create Risk score and Risk-score because both generate risk_score, but I can create two fields with different references and later edit their display names to those same values.

frontend/src/components/cases/add-custom-field-dialog.tsx
```diff
@@ -305,9 +310,13 @@ export function AddCustomFieldDialog({
         },
       })
 
-      queryClient.invalidateQueries({
-        queryKey: ["case-fields", workspaceId],
-      })
+      await Promise.all([
```
jordan-umusu: nit: this invalidation logic is duplicated in `edit-custom-fields-dialog.tsx` and `custom-fields-view.tsx`, can we consolidate this?

jordan-umusu (changes requested):
mostly LGTM, minor changes

frontend/src/components/cases/edit-custom-field-dialog.tsx
aaron-tracecat: yea that's reasonable, will think about how to handle this case

aaron-tracecat: I handled this by automatically appending a counter to the end, so like `risk_score` -> `risk_score_2`

I think in practice this shouldn't happen that much but at least it allows the user to think about the naming more clearly.

frontend/src/components/cases/add-custom-field-dialog.tsx
aaron-tracecat: Done

jordan-umusu (approved):

PR #3261, opened 2026-08-17
fix(executor): recover stale loop device nodes

daryllimyt:
## Summary

Repair executor containers whose `/dev` view is missing loop devices created by the shared node kernel after the container started, without exposing node-global block devices to action subprocesses.

## Root cause

Kubernetes containers have a private `/dev` populated at startup, while loop devices are allocated in the node kernel. A later `mount -o loop` can select a kernel loop device visible in `/sys/block` whose `/dev/loopN` node is absent inside an older executor pod, causing `failed to setup loop device` even though the kernel device exists.

## What changed

- Add a fixed-purpose C++ helper that mirrors validated kernel loop devices from `/sys/block/loopN/dev` into the container's `/dev`.
- Create repaired nodes as root-owned mode `0600` from birth, so UID-1001 action subprocesses cannot read, write, or chmod node-global loop devices.
- Keep the Python executor unprivileged; only the fixed-purpose helper receives `CAP_MKNOD`, `CAP_DAC_OVERRIDE`, and `CAP_SETUID`.
- Preserve the normal first mount attempt, then synchronize stale loop-device nodes and retry up to three times with short jitter after failure.
- Isolate mount/recovery policy in `registry_artifact_mounts.py` and cover helper protocol, retry, and failure behavior with fast unit tests.
- Run one xdist-safe privileged Docker regression against an explicitly supplied image resolved to its immutable ID. It reproduces the exact missing-node failure, mounts through the executor path, asserts UID 1001 cannot open the repaired node, and verifies cleanup.
- Reuse the exact executor image already built by scheduled integration CI rather than a possibly stale Compose image.

## Impact

No configuration or manifest changes are required. Existing executor images gain the helper in every environment, including `tracecat-ci`, and normal `mount -o loop` allocation remains kernel-managed.

## Validation

- `uv run pytest --confcutdir=tests/unit tests/unit/test_registry_artifact_mounts.py tests/unit/test_registry_artifacts.py -q` (`136 passed`)
- `TRACECAT__REGISTRY_LOOP_HOTPLUG_IMAGE=sha256:710e5c066b7baac344c2f0828a41321108effa2b3d2c9c713f8b5d171d4e4e39 uv run pytest --confcutdir=tests/integration tests/integration/test_registry_artifact_loop_hotplug.py -m integration -q -s` (`1 passed`)
- `uv run ruff check ...` and `uv run ruff format --check ...`
- `uv run pyright ...` (`0 errors, 0 warnings`)
- Workflow YAML parsed successfully.
- The C++ helper and full executor image built successfully with warnings treated as errors.

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 407 | 40 |
| Tests | 765 | 2 |
| Infra/config | 21 | 7 |

## Stack

PR 1 of 4. Followed by #3262, #3264, and #3265.

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -250,6 +250,8 @@ jobs:
           # Tests run on the host, so Temporal must be reachable via localhost
           TEMPORAL__CLUSTER_URL: "localhost:7233"
           TRACECAT__DISABLE_NSJAIL: "true"
+          # Privileged loop-device regression reuses the exact image built above.
+          TRACECAT__REGISTRY_LOOP_HOTPLUG_IMAGE: tracecat-ci:integration-${{ github.run_id }}-${{ github.run_attempt }}
         run: |
           # Skip pool integration tests for now.
           # Worker-specific task queues in conftest.py ensure isolation
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,15 @@ RUN git clone https://github.com/google/nsjail.git /tmp/nsjail && \
     install -m 0755 nsjail /usr/local/bin/nsjail && \
     rm -rf /tmp/nsjail
 
+COPY docker/loop-device-sync.cc /tmp/loop-device-sync.cc
+
+RUN g++ -std=c++20 -O2 -Wall -Wextra -Werror -Wformat=2 \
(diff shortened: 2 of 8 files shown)


daryllimyt:
@codex review

jordan-umusu (approved):

PR #3266, opened 2026-08-18
fix: retry actions after heartbeat timeouts

daryllimyt:
## Summary

- retry an action once when Temporal reports a typed activity heartbeat timeout
- preserve existing action retry budgets and propagate application and non-heartbeat failures unchanged
- guard the new command-producing path with a Temporal patch marker for replay compatibility
- add focused regression coverage for the retry and non-retry boundaries

## Root cause

An executor pod can disappear while running an activity, including during node or ephemeral-storage eviction. Temporal detects the lost worker through the heartbeat timeout and fails the workflow even when another healthy executor is available.

The heartbeat timeout is an infrastructure interruption with unknown completion state. This change therefore adds one bounded redispatch for default single-attempt actions instead of broadly increasing every action's retry policy.

## Impact

Workflows can recover from a single executor loss without retrying application failures or multiplying explicit action retry budgets. A second heartbeat timeout still fails normally.

Linear: [ENG-1668](https://linear.app/tracecat/issue/ENG-1668/retry-actions-once-after-a-temporal-heartbeat-timeout)

## Validation

- `uv run pytest --confcutdir=tests/unit tests/unit/test_dsl_workflow_concurrency.py tests/unit/test_dsl_workflow_activity_retry.py -q` — 26 passed
- `uv run ruff check tracecat/dsl/workflow.py tests/unit/test_dsl_workflow_activity_retry.py`
- `uv run pyright tracecat/dsl/workflow.py tests/unit/test_dsl_workflow_activity_retry.py`
- signed commit verified after rebasing onto current `origin/main`

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 36 | 2 |
| Tests | 167 | 0 |

---
## Summary by cubic
Retry actions once after a Temporal activity heartbeat timeout so workflows can recover from a lost executor. Previously, a heartbeat timeout failed the workflow; now single‑attempt actions get one bounded redispatch, while other failures and explicit retry budgets remain unchanged.

- Retries only when the cause is a heartbeat timeout, the task has `max_attempts == 1`, and the `dsl-action-heartbeat-timeout-retry-v1` patch is active; logs a warning and still fails on a second heartbeat timeout.
- Preserves application error propagation and non‑heartbeat timeouts; tasks with `max_attempts > 1` do not get an extra retry.
- Extracts `_execute_action_activity` and guards the new path with `workflow.patched` for replay compatibility; adds focused unit tests for retry and non‑retry boundaries.
- Addresses Linear ENG-1668.

<sup>Written for commit b379e52b4ce751e9a8dd398818094f12c0dfc4f9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3266?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tracecat/dsl/workflow.py
+++ b/tracecat/dsl/workflow.py
@@ -18,7 +18,11 @@ from temporalio.exceptions import (
     ApplicationError,
     ChildWorkflowError,
     FailureError,
+    TimeoutType,
     is_cancelled_exception,
+)
+from temporalio.exceptions import (
+    TimeoutError as TemporalTimeoutError,
 )
 
 with workflow.unsafe.imports_passed_through():
@@ -163,6 +167,7 @@ with workflow.unsafe.imports_passed_through():
 
 
 _CHILD_RUN_ARG_PREP_YIELD_EVERY = 8
+ACTION_HEARTBEAT_TIMEOUT_RETRY_PATCH = "dsl-action-heartbeat-timeout-retry-v1"
 
 
 def _inherit_search_attributes_with_alias(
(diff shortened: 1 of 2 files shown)


tracecat/dsl/workflow.py
```diff
@@ -1795,9 +1800,39 @@ async def _run_action(self, task: ActionStatement) -> StoredObject:
...
+        """Dispatch an action, retrying one executor heartbeat interruption."""
+        try:
+            stored = await self._execute_action_activity_once(task, arg)
+        except ActivityError as exc:
+            cause = exc.cause
+            should_retry_heartbeat = (
+                task.retry_policy.max_attempts == 1
+                and isinstance(cause, TemporalTimeoutError)
+                and cause.type is TimeoutType.HEARTBEAT
+                and workflow.patched(ACTION_HEARTBEAT_TIMEOUT_RETRY_PATCH)
```
aaron-tracecat, line 1817: should we embed the task id in the patch id? 
`f"{ACTION_HEARTBEAT_TIMEOUT_RETRY_PATCH}-{task.id}"`? 

otherwise I'm not sure how this is tied to the specific task

aaron-tracecat: Daryl answered offline: this code is running on a single worker, which is already running a workflow instance. So the patch only applies to this specific temporal worker

aaron-tracecat (approved):

PR #3272, opened 2026-08-19
feat(cases): add title and description versioning

aaron-tracecat:
## Description

Implements the persistence foundation for case title and description versioning from ENG-1674.

- Adds append-only, per-field `CaseVersion` rows for case summaries and descriptions.
- Creates version 1 on case creation and appends only when persisted content changes; whitespace-only edits remain meaningful changes.
- Serializes version allocation with a row lock so concurrent writes receive unique, consecutive numbers.
- Records the persisted user when available and clears attribution safely if that user is deleted.
- Registers case versions for tenant RLS.
- Keeps case activity behavior unchanged: description edits persist a new description version without emitting a `case_updated` event.

The migration creates one summary baseline and one description baseline for every existing case. It intentionally does not reconstruct historical content from case events.

## Related Issues

- [ENG-1674](https://linear.app/tracecat/issue/ENG-1674)
- Parent design: [ENG-1637](https://linear.app/tracecat/issue/ENG-1637/case-description-versioning)

## User / developer impact

Existing cases receive a reliable version-1 baseline during rollout. New case writes maintain independent, append-only summary and description histories for the follow-up version APIs and UI. No public version-history API or frontend behavior is added in this PR.

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 188 | 2 |
| Tests | 390 | 0 |
| Generated | 161 | 0 |

Generated consists of the Alembic revision.

## Steps to QA

1. Run `uv run pytest tests/integration/test_case_versions.py tests/migration/test_case_versions_migration.py tests/unit/test_cases_service.py -q`.
2. Run `uv run basedpyright --warnings --threads 4`.

The focused test run passed 59 tests. Commit-time hooks also passed Ruff, frontend client generation, Biome, backend typing, frontend typing, and secret scanning.

### DIFF
--- a/alembic/versions/598b32358ec5_add_case_versions.py
+++ b/alembic/versions/598b32358ec5_add_case_versions.py
@@ -141,8 +141,8 @@ def upgrade() -> None:
                 baseline.content,
                 NULL,
                 case_row.workspace_id,
-                transaction_timestamp(),
-                transaction_timestamp()
+                case_row.updated_at,
+                case_row.updated_at
             FROM "case" AS case_row
             CROSS JOIN LATERAL (
                 VALUES
--- a/alembic/versions/598b32358ec5_add_case_versions.py
+++ b/alembic/versions/598b32358ec5_add_case_versions.py
@@ -1,0 +1,161 @@
+"""Add case versions.
+
+Revision ID: 598b32358ec5
+Revises: 35fa9fc71258
+Create Date: 2026-08-19 17:33:35.727844
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+from alembic import op
+from tracecat.db.tenant_rls import (
+    disable_workspace_table_rls,
+    enable_workspace_table_rls,
+)
+
+# revision identifiers, used by Alembic.
+revision: str = "598b32358ec5"
+down_revision: str | None = "35fa9fc71258"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    sa.Enum("SUMMARY", "DESCRIPTION", name="caseversionfield").create(op.get_bind())
+    op.create_table(
+        "case_version",
+        sa.Column("id", sa.UUID(), nullable=False),
+        sa.Column("case_id", sa.UUID(), nullable=False),
+        sa.Column(
+            "field",
+            postgresql.ENUM(
+                "SUMMARY",
+                "DESCRIPTION",
+                name="caseversionfield",
+                create_type=False,
+            ),
+            nullable=False,
+        ),
+        sa.Column("version", sa.Integer(), nullable=False),
+        sa.Column("content", sa.Text(), nullable=False),
+        sa.Column("user_id", sa.UUID(), nullable=True),
+        sa.Column("workspace_id", sa.UUID(), nullable=False),
+        sa.Column("surrogate_id", sa.Integer(), nullable=False),
+        sa.Column(
+            "created_at",
+            sa.TIMESTAMP(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.Column(
+            "updated_at",
+            sa.TIMESTAMP(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.CheckConstraint(
+            "version > 0",
+            name=op.f("ck_case_version_version_positive"),
+        ),
+        sa.ForeignKeyConstraint(
+            ["case_id"],
+            ["case.id"],
+            name=op.f("fk_case_version_case_id_case"),
+            ondelete="CASCADE",
+        ),
+        sa.ForeignKeyConstraint(
+            ["user_id"],
+            ["user.id"],
+            name=op.f("fk_case_version_user_id_user"),
+            ondelete="SET NULL",
+        ),
+        sa.ForeignKeyConstraint(
+            ["workspace_id"],
+            ["workspace.id"],
+            name=op.f("fk_case_version_workspace_id_workspace"),
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("surrogate_id", name=op.f("pk_case_version")),
+        sa.UniqueConstraint(
+            "workspace_id",
+            "case_id",
+            "field",
+            "version",
+            name="uq_case_version_workspace_case_field_version",
+        ),
+    )
+    op.create_index(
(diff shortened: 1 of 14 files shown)


tracecat/cases/service.py
```diff
@@ -1049,12 +1069,19 @@ async def _apply_case_update(self, case: Case, params: CaseUpdate) -> None:
...
+                        content=value,
+                    )
                     events.append(
                         UpdatedEvent(
-                            field="summary", old=old, new=value, wf_exec_id=wf_exec_id
+                            field=typing_cast(Literal["summary", "description"], key),
+                            old=old,
+                            new=value,
+                            wf_exec_id=wf_exec_id,
                         )
```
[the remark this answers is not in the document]
aaron-tracecat: Removed the description activity event in 873154c; description updates now only persist versions.

tracecat/cases/versions/service.py
```diff
@@ -0,0 +1,80 @@
...
+
+    async def append_version(
+        self,
+        *,
+        case_id: uuid.UUID,
+        field: CaseVersionField,
+        content: str,
+    ) -> CaseVersion:
+        """Append and flush the next version without committing."""
+        await self.lock_case(case_id)
```
jordan-umusu: **Batch update depends on this inner lock.** `batch_update_cases` locks via `_lock_cases(key_share=True)` (`service.py:1147`). `FOR KEY SHARE` doesn't serialize two concurrent batches on the same case — the `FOR UPDATE` taken here is what actually prevents both reading `max=1` and one failing on `uq_case_version_workspace_case_field_version`.

If the inner lock is removed (see comment on `update_case`), batch must upgrade to `FOR UPDATE` when versioned fields are in the params, e.g. `key_share = not ({"summary", "description"} & params.model_fields_set)`. Either way, worth documenting the dependency.

tracecat/cases/service.py
```diff
@@ -927,6 +941,12 @@ async def update_case(self, case: Case, params: CaseUpdate) -> Case:
         """
 
         try:
+            if {"summary", "description"} & params.model_fields_set:
+                await self.versions.lock_case(case.id)
```
jordan-umusu, line 940: Since ``update_case`` already holds FOR UPDATE here, can we skip locking inside append_version?

tests/integration/test_case_versions.py
```diff
@@ -0,0 +1,205 @@
...
+            await session.execute(
+                select(CaseVersion.id).where(CaseVersion.case_id == case.id)
+            )
+        )
+        .scalars()
+        .all()
+    )
+
+
+async def test_concurrent_case_version_allocation(svc_role: Role) -> None:
```
jordan-umusu: From Codex Thermos P3: This exercises `CaseVersionsService.append_version` directly, so it proves the allocator serializes — not that `update_case` / `batch_update_cases` do. Suggest driving two concurrent `CasesService.update_case(summary=...)` calls (plus one batch case) and asserting `[1, 2, 3]` with the expected content per version. That's the only test that would catch the lock-ownership change regressing.

tracecat/cases/service.py
```diff
@@ -784,6 +787,17 @@ async def create_case(self, params: CaseCreate) -> Case:
             # Generate the case ID without locking the workspace counter.
             await self.session.flush()
 
+            await self.versions.append_version(
```
jordan-umusu: Both calls here lock + query `max(version)` that can only be null. Can we skip locking here and insert the two `CaseVersion(version=1, ...)` rows directly?

alembic/versions/598b32358ec5_add_case_versions.py
```diff
@@ -0,0 +1,161 @@
...
+                updated_at
+            )
+            SELECT
+                gen_random_uuid(),
+                case_row.id,
+                baseline.field::caseversionfield,
+                1,
+                baseline.content,
+                NULL,
+                case_row.workspace_id,
```
jordan-umusu, line 143: nit: `case_row.updated_at` is probably more faithful for `ix_case_version_case_timeline` vs `transaction_timestamp()`

tracecat/cases/versions/service.py
```diff
@@ -0,0 +1,80 @@
...
+    """Create immutable versions for case titles and descriptions."""
+
+    service_name = "case_versions"
+
+    def __init__(self, session: AsyncSession, role: Role | None = None):
+        super().__init__(session, role)
+        self._user_id_resolved = False
+        self._user_id: uuid.UUID | None = None
+
+    async def _resolve_user_id(self) -> uuid.UUID | None:
```
jordan-umusu, line 27: Do we need this? The role is guaranteed to be valid by the time it reaches a service and if a dangling user id is possible it'd fail on the FK anyways.

jordan-umusu (changes requested):
mostly LGTM, small changes/refactors

tracecat/cases/service.py
aaron-tracecat: Removed the redundant lock from append_version; update_case and batch_update_cases now own the case lock.

tests/integration/test_case_versions.py
aaron-tracecat: Updated the integration test to exercise concurrent update_case and batch_update_cases writes, asserting versions [1, 2, 3] and their contents.

tracecat/cases/service.py
aaron-tracecat: Added a direct initial-version path so case creation inserts both version-1 rows without locking or querying max(version).

alembic/versions/598b32358ec5_add_case_versions.py
aaron-tracecat: Updated the baseline rows to use `case_row.updated_at` and added a migration assertion for it.

tracecat/cases/versions/service.py
aaron-tracecat: Yes we need this. Version persistence should not fail if an attributed user has already been deleted: the resolver stores NULL instead of violating the FK, while ON DELETE SET NULL handles deletion after the version is created.

alembic/versions/598b32358ec5_add_case_versions.py
```diff
@@ -0,0 +1,161 @@
...
+        op.f("ix_case_version_user_id"),
+        "case_version",
+        ["user_id"],
+        unique=False,
+    )
+
+    op.execute(
+        sa.text(
+            """
+            INSERT INTO case_version (
```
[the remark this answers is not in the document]
aaron-tracecat: I think that this is actually completely ok, as we don't need the existing versions to be completely comprehensive or perfect. We just need a starting off point for the upcoming version timeline.

jordan-umusu (approved):
LGTM

PR #3282, opened 2026-08-20
feat(cases): add case version history and restore APIs

aaron-tracecat:
## Summary

- Add cursor-paginated history for case `summary` and `description` versions, with actor and latest-version metadata.
- Add comparison APIs that return the selected version, its immediate same-field predecessor, and a render-ready word-level textual diff.
- Add append-only restore APIs routed through the existing locked case-update transaction, with workspace/case-scoped `404` behavior.

Depends on #3272. Implements [ENG-1676](https://linear.app/tracecat/issue/ENG-1676/featcases-add-case-version-history-and-restore-apis).

## Review guide

1. **API contract:** `tracecat/cases/versions/schemas.py` and the three routes in `tracecat/cases/router.py`.
2. **History reads:** `CaseVersionsService` projects metadata only for lists, uses stable `(created_at, surrogate_id)` cursors, and fetches the immediate predecessor for comparison.
3. **Textual diff:** `tracecat/cases/versions/diff.py` produces ordered, exact-text `equal`, `delete`, and `insert` segments. Whitespace and punctuation are preserved, replacements are `delete` + `insert`, `changed` handles identical restores, and `diff` is `null` when no predecessor exists.
4. **Restore semantics:** `CasesService.restore_version` validates scope, locks the case, and forces a new immutable version even when restored content matches the current head.
5. **Coverage/client:** focused unit and integration tests cover pagination, auth, cursor validation, mismatches, diff reconstruction, compare, restore, rollback, and concurrency; the generated TypeScript client exposes the complete contract.

| Method | Endpoint | Purpose |
| --- | --- | --- |
| `GET` | `/cases/{case_id}/versions` | List version metadata, optionally filtered by field |
| `GET` | `/cases/{case_id}/versions/{version_id}/compare` | Return the selected version, predecessor, and structured textual diff |
| `POST` | `/cases/{case_id}/versions/{version_id}/restore` | Restore through the normal case-update transaction |

The frontend can render `diff.segments` in order: show `equal` normally, `delete` as removed content, and `insert` as added content. Ignoring inserts reconstructs the predecessor; ignoring deletes reconstructs the selected version.

## Verification

- `uv run pytest tests/unit/test_case_versions_service.py tests/unit/test_cases_service.py tests/unit/api/test_api_cases.py tests/unit/test_route_scope_guards.py tests/integration/test_case_versions.py -x` — 232 passed
- `uv run ruff check .` and `uv run ruff format --check .`
- `uv run basedpyright --warnings --threads 4`
- `pnpm -C frontend check` and `pnpm -C frontend run typecheck`
- Signed commit hooks, including secret scanning and generated-client validation

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 487 | 10 |
| Tests | 695 | 2 |
| Generated | 565 | 0 |

### DIFF
--- a/tracecat/cases/router.py
+++ b/tracecat/cases/router.py
@@ -27,7 +27,6 @@ from tracecat.cases.enums import (
     CasePriority,
     CaseSeverity,
     CaseStatus,
-    CaseVersionField,
 )
 from tracecat.cases.filters import parse_assignee_filter
 from tracecat.cases.rows.service import CaseTableRowsService
@@ -64,11 +63,7 @@ from tracecat.cases.service import (
 )
 from tracecat.cases.tags.schemas import CaseTagRead
 from tracecat.cases.tags.service import CaseTagsService
-from tracecat.cases.versions.schemas import (
-    CaseVersionCompareRead,
-    CaseVersionReadMinimal,
-    CaseVersionRestoreRead,
-)
+from tracecat.cases.versions.router import router as case_versions_router
 from tracecat.db.dependencies import AsyncDBSession
 from tracecat.exceptions import (
     TracecatAuthorizationError,
@@ -81,7 +76,6 @@ from tracecat.logger import logger
 from tracecat.pagination import (
     CursorPaginatedResponse,
     CursorPaginationParams,
-    PageParams,
 )
 from tracecat.tiers.enums import Entitlement
 
@@ -557,101 +551,7 @@ async def batch_delete_cases(
         ) from exc
 
 
-@cases_router.get(
-    "/{case_id}/versions",
-    response_model=CursorPaginatedResponse[CaseVersionReadMinimal],
-)
-@require_scope("case:read")
-async def list_case_versions(
-    *,
-    role: WorkspaceActorRouteRole,
-    session: AsyncDBSession,
-    case_id: uuid.UUID,
-    limit: int = Query(
-        config.TRACECAT__LIMIT_DEFAULT,
-        ge=config.TRACECAT__LIMIT_MIN,
-        le=config.TRACECAT__LIMIT_CURSOR_MAX,
-        description="Maximum items per page",
-    ),
-    cursor: str | None = Query(
-        None,
-        max_length=8192,
-        description="Cursor for pagination",
-    ),
-    field: CaseVersionField | None = Query(
-        None,
-        description="Optionally include only summary or description versions",
-    ),
-) -> CursorPaginatedResponse[CaseVersionReadMinimal]:
-    """List immutable case field versions newest-first."""
-    service = CasesService(session, role)
-    if await service.get_case(case_id) is None:
-        raise HTTPException(
-            status_code=HTTP_404_NOT_FOUND,
-            detail=f"Case with ID {case_id} not found",
-        )
-    try:
-        return await service.versions.list_versions(
-            case_id=case_id,
-            page=PageParams(limit=limit, cursor=cursor),
-            field=field,
-        )
-    except TracecatValidationError as exc:
-        raise HTTPException(
-            status_code=HTTP_400_BAD_REQUEST,
-            detail=exc.detail or str(exc),
-        ) from exc
-
-
-@cases_router.get(
-    "/{case_id}/versions/{version_id}/compare",
-    response_model=CaseVersionCompareRead,
-)
-@require_scope("case:read")
-async def compare_case_version(
-    *,
-    role: WorkspaceActorRouteRole,
-    session: AsyncDBSession,
-    case_id: uuid.UUID,
-    version_id: uuid.UUID,
-) -> CaseVersionCompareRead:
-    """Compare a case field version with its immediate predecessor."""
-    service = CasesService(session, role)
-    comparison = await service.versions.compare_with_predecessor(
-        case_id=case_id,
-        version_id=version_id,
-    )
-    if comparison is None:
-        raise HTTPException(
-            status_code=HTTP_404_NOT_FOUND,
-            detail=f"Case version '{version_id}' not found",
-        )
-    return comparison
-
-
-@cases_router.post(
-    "/{case_id}/versions/{version_id}/restore",
-    response_model=CaseVersionRestoreRead,
-)
-@require_scope("case:update")
-async def restore_case_version(
-    *,
-    role: WorkspaceActorRouteRole,
-    session: AsyncDBSession,
-    case_id: uuid.UUID,
-    version_id: uuid.UUID,
-) -> CaseVersionRestoreRead:
-    """Restore one historical case field version atomically."""
-    try:
-        return await CasesService(session, role).restore_version(
-            case_id=case_id,
-            version_id=version_id,
-        )
-    except TracecatNotFoundError as exc:
-        raise HTTPException(
-            status_code=HTTP_404_NOT_FOUND,
-            detail=str(exc),
-        ) from exc
+cases_router.include_router(case_versions_router)
 
 
 @cases_router.get("/{case_id}")
--- a/tracecat/cases/router.py
+++ b/tracecat/cases/router.py
@@ -23,7 +23,12 @@ from tracecat.auth.schemas import UserRead
 from tracecat.auth.users import search_users
 from tracecat.authz.controls import require_scope
 from tracecat.cases.dropdowns.service import CaseDropdownValuesService
-from tracecat.cases.enums import CasePriority, CaseSeverity, CaseStatus
+from tracecat.cases.enums import (
+    CasePriority,
+    CaseSeverity,
+    CaseStatus,
+    CaseVersionField,
+)
 from tracecat.cases.filters import parse_assignee_filter
 from tracecat.cases.rows.service import CaseTableRowsService
 from tracecat.cases.schemas import (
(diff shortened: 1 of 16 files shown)


tracecat/cases/router.py
```diff
@@ -546,6 +557,99 @@ async def batch_delete_cases(
...
+    role: WorkspaceActorRouteRole,
+    session: AsyncDBSession,
+    case_id: uuid.UUID,
+    limit: int = Query(
+        config.TRACECAT__LIMIT_DEFAULT,
+        ge=config.TRACECAT__LIMIT_MIN,
+        le=config.TRACECAT__LIMIT_CURSOR_MAX,
+        description="Maximum items per page",
+    ),
+    cursor: str | None = Query(None, description="Cursor for pagination"),
```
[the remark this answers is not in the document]
aaron-tracecat: Addressed in 7ce5edd60. The route now enforces the same 8192-character maximum at the query boundary, and a focused API regression test verifies oversized cursors return 422 before CasesService is constructed.

tracecat/cases/service.py
```diff
@@ -951,7 +963,46 @@ async def update_case(self, case: Case, params: CaseUpdate) -> Case:
...
+    ) -> CaseVersionRestoreRead:
+        """Restore one scoped field version through the normal update transaction."""
+        case = await self.get_case(case_id, for_update=True)
+        if case is None:
+            raise TracecatNotFoundError(f"Case '{case_id}' not found")
+
+        version = await self.versions.get_version(
+            case_id=case_id,
+            version_id=version_id,
+        )
```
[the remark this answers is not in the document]
aaron-tracecat: Thanks for flagging this. No code change is needed because Tracecat scope implication explicitly makes case:update satisfy case:read in tracecat/authz/controls.py, and tests/unit/test_rbac_scopes.py::test_update_implies_read covers that exact pair. An update-only role therefore passes the case:read guard on get_version.

tracecat/cases/router.py
```diff
@@ -546,6 +557,103 @@ async def batch_delete_cases(
...
+        description="Cursor for pagination",
+    ),
+    field: CaseVersionField | None = Query(
+        None,
+        description="Optionally include only summary or description versions",
+    ),
+) -> CursorPaginatedResponse[CaseVersionReadMinimal]:
+    """List immutable case field versions newest-first."""
+    service = CasesService(session, role)
+    if await service.get_case(case_id) is None:
```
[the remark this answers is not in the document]
aaron-tracecat: Thanks for the suggestion. The eager relationships are existing CasesService.get_case behavior. This endpoint performs one workspace-scoped lookup to preserve 404 semantics before the projection-only, cursor-paginated version query. We are keeping this PR focused on the history and restore API contract; a projected-ID existence helper can be handled as a separate optimization if needed.

tracecat/cases/versions/diff.py
```diff
@@ -0,0 +1,68 @@
...
+    predecessor: str,
+    selected: str,
+) -> CaseVersionDiffRead:
+    """Build a word-level edit script from predecessor to selected content."""
+    predecessor_tokens = _WORD_DIFF_TOKEN_PATTERN.findall(predecessor)
+    selected_tokens = _WORD_DIFF_TOKEN_PATTERN.findall(selected)
+    matcher = SequenceMatcher(
+        a=predecessor_tokens,
+        b=selected_tokens,
+        autojunk=False,
```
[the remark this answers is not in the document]
aaron-tracecat: Addressed in dc63829cd by restoring SequenceMatcher’s standard autojunk heuristic. A synthetic 4,998-character repeated-punctuation comparison that previously ran for over a minute now completes in about 1 ms, while the focused diff tests and exact reconstruction checks still pass.

tracecat/cases/router.py
```diff
@@ -546,6 +557,103 @@ async def batch_delete_cases(
         ) from exc
 
 
+@cases_router.get(
+    "/{case_id}/versions",
```
jordan-umusu: refactor: move these routes under cases/versions

tracecat/cases/router.py
```diff
@@ -546,6 +557,103 @@ async def batch_delete_cases(
...
+        max_length=8192,
+        description="Cursor for pagination",
+    ),
+    field: CaseVersionField | None = Query(
+        None,
+        description="Optionally include only summary or description versions",
+    ),
+) -> CursorPaginatedResponse[CaseVersionReadMinimal]:
+    """List immutable case field versions newest-first."""
+    service = CasesService(session, role)
```
jordan-umusu: `get_case` here is just an existence check but it does `select(Case)` and pulls in `table_rows`, `assignee`, `tags`, `dropdown_values` via selectin. ~5 queries per list/compare/restore for nothing. `lock_case` in the versions service already has the right query minus the `for_update`. add a `case_exists` next to it and use that in all three handlers.

tracecat/cases/router.py
jordan-umusu: refactor: Could we move the version routes into `cases/versions/router.py`? `cases/attachments/router.py` and `cases/rows/router.py` are good references. Totally fair that you followed what's already in `cases/router.py`, but I think the god case router pattern is a smell so I'd rather we stop adding to it.

tracecat/cases/router.py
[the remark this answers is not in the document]
jordan-umusu: Can we actually throw this in too? It should under 20 lines and a quick benchmark on a case with 200 linked rows/tags/dropdowns: get_case probe ~13.5ms vs 0.3ms for select(Case.id), effectively 10x slowerwith the paginated versions query itself at ~1.2ms. Just less headache as things scale.

tracecat/cases/versions/service.py
```diff
@@ -108,3 +124,150 @@ async def append_version(
...
+            )
+        )
+        row = (await self.session.execute(statement)).one_or_none()
+        if row is None:
+            return None
+        selected_version = cast(CaseVersion, row[0])
+        predecessor_version = cast(CaseVersion | None, row[1])
+        selected_read = CaseVersionContentRead.model_validate(selected_version)
+        predecessor_read = (
+            CaseVersionContentRead.model_validate(predecessor_version)
```
jordan-umusu, line 259: Wrap `compute_case_version_diff` in `asyncio.to_thread` so it does not block the API event loop.

jordan-umusu (changes requested):

tracecat/cases/versions/service.py
aaron-tracecat: Implemented in fb0bb56f5. Diff computation now runs through asyncio.to_thread when a predecessor exists. The integration test verifies the offloaded function and inputs while retaining the response assertions.

tracecat/cases/router.py
aaron-tracecat: Implemented in fb0bb56f5. The three version endpoints now live in cases/versions/router.py and are included by the parent cases router. Route behavior and the generated client remain unchanged.

tracecat/cases/router.py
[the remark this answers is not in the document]
aaron-tracecat: Implemented in fb0bb56f5. The history endpoint now uses a workspace-scoped case_exists probe that selects only Case.id. Tests cover existing and missing cases and verify that get_case is not called.

tracecat/cases/service.py
```diff
@@ -951,7 +971,46 @@ async def update_case(self, case: Case, params: CaseUpdate) -> Case:
...
-    async def _apply_case_update(self, case: Case, params: CaseUpdate) -> None:
+    @require_scope("case:update")
+    async def restore_version(
+        self,
+        *,
+        case_id: uuid.UUID,
+        version_id: uuid.UUID,
+    ) -> CaseVersionRestoreRead:
+        """Restore one scoped field version through the normal update transaction."""
+        case = await self.get_case(case_id, for_update=True)
```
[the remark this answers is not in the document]
aaron-tracecat: Implemented in 41a606d70. Restore now reuses the workspace-scoped locked-case query, which applies SELECT FOR UPDATE while suppressing relationship loading. The restore integration test also verifies that get_case is not called.

tracecat/cases/router.py
```diff
@@ -546,6 +551,9 @@ async def batch_delete_cases(
         ) from exc
 
 
+cases_router.include_router(case_versions_router)
```
jordan-umusu: super-nit: could we mount case_versions_router directly in tracecat/api/app.py alongside the other case-related routers?

tracecat/cases/versions/diff.py
```diff
@@ -0,0 +1,67 @@
...
+    predecessor: str,
+    selected: str,
+) -> CaseVersionDiffRead:
+    """Build a word-level edit script from predecessor to selected content."""
+    predecessor_tokens = _WORD_DIFF_TOKEN_PATTERN.findall(predecessor)
+    selected_tokens = _WORD_DIFF_TOKEN_PATTERN.findall(selected)
+    matcher = SequenceMatcher(
+        a=predecessor_tokens,
+        b=selected_tokens,
+    )
```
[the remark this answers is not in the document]
aaron-tracecat: Implemented in 71fd20a7e. Comparisons now skip SequenceMatcher when the old-token and new-token pair count exceeds one million, returning an exact coarse delete/insert diff instead. Identical inputs also bypass matching, and the regression test verifies SequenceMatcher is not called above the budget.

jordan-umusu (approved):
lgtm

tracecat/cases/router.py
aaron-tracecat: good call good call

aaron-tracecat: Done in 235e3771d. case_versions_router is now mounted directly in tracecat/api/app.py alongside the other case routers. Paths and API types are unchanged; the generated client changes are ordering only.

PR #3303, opened 2026-08-25
fix(docker): use dynamic ports for service healthchecks

### LINKED ISSUE #3237: docker-compose.dev.yml uses hardcoded ports for service healthchecks
**Describe the bug**
Trying to run a local dev instance, my port 8000 is already in use, so set a different port for the API using `export PORT 8001` (or something like that). Then after

`docker compose -f docker-compose.dev.yml -f docker-compose.sandbox.yml up` (or similar) the health check in docker-compose.dev.yml never passes

(Issue is pretty straight-forward if you look at the docker-compose.dev.yml file)

sundeep8967:
## Problem
When running Tracecat locally, if a user exports a custom `PORT` (e.g. `export PORT=8001`) to avoid conflicts, the healthchecks in the `docker-compose` files fail because they are hardcoded to query `http://localhost:8000/health`.

## Solution
Modified the `api` and `litellm` healthchecks across all docker-compose files (`dev`, `local`, `prod`) to use `CMD-SHELL` instead of `CMD`. This allows the healthcheck to dynamically evaluate the container's internal environment variables (`$PORT` and `$TRACECAT__LITELLM_PORT`) when curling the endpoints. 

If the port is overridden by the user, the healthcheck will now correctly query the overridden port instead of failing.

Fixes #3237

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -98,7 +98,7 @@ services:
       redis:
         condition: service_healthy
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8000/health"]
+      test: ["CMD-SHELL", "curl -f http://localhost:$${PORT:-8000}/health || exit 1"]
       interval: 5s
       timeout: 5s
       retries: 30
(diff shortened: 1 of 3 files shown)


topher-lo:
@codex review

topher-lo:
Makes sense but we have a separate PR fixing this. Thank you for opening though!

PR #3328, opened 2026-08-27
feat(ee): vendor copilot skills from the public plugin repo

topher-lo:
> **Stacked on #3326.** Base is `agent/mcp-prompt-corrections`, so this diff shows only the vendoring change. Merge #3326 first, then retarget this to `main`.
>
> **Draft** until [tracecat-plugins#2](https://github.com/TracecatHQ/tracecat-plugins/pull/2) merges — the pinned SHA points at that PR's head and must be re-pinned to the merge commit.

## Summary

`TracecatHQ/tracecat-plugins` is now public and is the canonical home for Tracecat user-facing agent skills. The workspace-chat copilot's built-in skills were a divergent second copy: one commit, never revised, no eval coverage, and wrong in places (it claimed a `datetime` typecast that does not exist, and carried two navigation tables that disagreed with each other).

This vendors three skills from the plugin at image build time and deletes the duplicates.

## Where the files land, and why it differs per stage

This is the part worth reviewing carefully. The two app stages resolve `importlib.resources` differently:

- `development-app` runs `uv sync --frozen --no-dev` — **editable**, so `tracecat_ee` resolves into the source tree.
- `production-app` runs `uv sync --locked --no-dev --no-editable` — a **copied** install into `site-packages`. Anything added to `/app/packages/...` after that sync has no effect.

So the `COPY --from=plugin-skills` must precede each stage's sync. Get it wrong and production silently ships a copilot with no skills — no error, just missing guidance. The fetch stage fails loudly if any of the three skills is absent rather than copying an empty tree.

Verified by building the `tracecat-ee` wheel and confirming the vendored trees are inside it:

```
tracecat_ee/workspace_chat/skills/tracecat-automation-best-practices/SKILL.md
tracecat_ee/workspace_chat/skills/tracecat-automation-best-practices/references/*.md   (9 files)
tracecat_ee/workspace_chat/skills/tracecat-slackbot-best-practices/...
tracecat_ee/workspace_chat/skills/tracecat-platform-guide/SKILL.md
```

The pinned fetch was exercised against the real public repo — the archive resolves and contains all three skills.

## Local development

`just sync-copilot-skills` does the same thing locally, preferring a sibling `../tracecat-plugins` checkout and falling back to the pinned archive. The vendored directories are gitignored, so a fresh checkout has none until you run it.

## Tests

`test_builtin_skills.py` previously asserted only that `SKILL.md` existed — which is exactly how a wrong typecast claim and two divergent navigation tables survived in the copies this PR deletes. It now also parses frontmatter, checks the name matches its directory, and resolves every reference link. Those content assertions skip with a message naming `just sync-copilot-skills` when the vendored tree is absent; the plumbing tests (payload threading, entitlement gating, staging, collision) stay unconditional since they use fixtures.

## Deletions

- `packages/tracecat-ee/.../workspace_chat/skills/tracecat-manage-workflows/` and `tracecat-platform-guide/`
- `.agents/skills/tracecat-automation-best-practices/` and `tracecat-slackbot-best-practices/` — now published in the plugin

`.agents` keeps `gh-prerelease`, `docs-authoring`, `test-tracecat-prompts` and `tracecat-qa`, which are repo-engineering skills, not user-facing. Note `.claude/skills` symlinks to `.agents/skills`, so contributors who relied on the two deleted skills locally should install the plugin instead.

## Validation

```
just sync-copilot-skills                       # copies all three
run_local.py --static-only                     # 48 checks, 0 failing
pytest tests/unit/test_builtin_skills.py tests/unit/test_mcp_server.py   # 268 passed
```

The static count moves 49 → 48 because the checks are per-fence and the plugin's `SKILL.md` has one YAML fence instead of two — the `model:` block moved into a reference. I diffed the check names across both branches to confirm no guard was dropped; only `best_practices_skill:fence_1_yaml:yaml_parse` disappears.

---
## Summary by cubic
Vendors the workspace-chat copilot's built-in skills from the public `TracecatHQ/tracecat-plugins` repo at image build time, replacing the in-repo copies that had drifted — including a nonexistent `datetime` typecast claim and two conflicting navigation tables.

**Build and rollout**
- A new `plugin-skills` build stage fetches a pinned archive and fails loudly when a skill is missing; the copy must land before each stage's `uv sync` because production installs non-editable and would otherwise ship silently without skills.
- `just sync-copilot-skills` populates the same vendored directories locally (sibling checkout first, then the pinned archive); the directories are gitignored.
- `TRACECAT_PLUGINS_REF` is pinned to the merge commit of `tracecat-plugins` PR #2.
- Removes the old EE skills and the two `.agents` skills now published in the plugin.

**Tests**
- `test_builtin_skills.py` now checks frontmatter, directory-name match, and reference links; those content assertions skip with a hint when the vendored tree is absent.
- Static check count drops 49 → 48 because the plugin's `SKILL.md` has one YAML fence instead of two; no guard is dropped.

<sup>Written for commit dc1833808aadc88aa5d4439ad17b605bf08f7d9b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3328?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


topher-lo:
Combined into #3326, which now carries the MCP prompt fixes, the eval-harness removal, and this vendoring work as three commits against `main`. Closing to keep it to one PR.

PR #3412, opened 2026-09-07
fix(engine): cap retry_until iterations with loop-limit error

### LINKED ISSUE #3411: Unsatisfiable retry_until conditions loop forever until history limits
A workflow action with a `retry_until` condition that never evaluates true currently retries forever.

For such an action, the DSL workflow loops with no iteration bound. `DSLConfig.timeout=0` (the default) means no workflow timeout, and the execution timeout resolves to `None` when unset, so the only backstop is Temporal's history size or count limit. When that limit is hit, the execution fails with an opaque history-limit error instead of a clear, user-attributed message.

Expected behavior: the loop should stop at a bounded iteration count and fail with a clear, non-retryable error that names the task and the limit.

Observed on the current main branch. Reproduced with a single `core.transform.reshape` action whose `retry_until` condition is a constant false expression under the time-skipping test environment.

warun7:
## Summary

- Adds `MAX_RETRY_UNTIL_ITERATIONS = 2048` to `tracecat/dsl/constants.py`, reusing the do-while cap value under a distinct name so the intent stays explicit
- Bounds the `retry_until` retry loop in `_execute_task_until_condition` with a deterministic local counter, which is safe for replay
- When the cap is reached, raises a user-attributed, non-retryable `workflow.loop.limit_exceeded` application error via `raise_application_error_from_classification`, mirroring the scheduler's do-while loop-limit error
- Fixes #3411

## Why

A `retry_until` condition that never evaluates true loops with no bound today. With the default config (no workflow timeout, no execution timeout, and no tier limits in the OSS default), the only backstop is Temporal's history limits, which surface as an opaque error. The new cap fails cleanly at the limit with a message that names the task and suggests fixing the condition. Workflows whose conditions succeed normally are unaffected.

## Testing

- New unit tests in `tests/unit/test_dsl_workflow_retry_until.py`: cap exhaustion raises a non-retryable `ApplicationError` whose classification is `USER` / `WORKFLOW_LOOP_LIMIT_EXCEEDED`; a condition satisfied within the cap returns that iteration's result with exact call counts
- New temporal test `test_workflow_retry_until_rejects_over_iteration_cap` in `tests/temporal/test_workflow_timers.py`: with the cap monkeypatched to 3, the action runs exactly 3 times, the failure cause is a non-retryable `ApplicationError` with the limit message, and `extract_error_classification` yields user ownership with `WORKFLOW_LOOP_LIMIT_EXCEEDED`
- All 31 unit tests across the touched suites pass, and the temporal test passes against the time-skipping environment
- `ruff check`, `ruff format --check`, and `basedpyright --warnings` are clean on all touched files

## LOC breakdown

| Category | Added | Removed |
| --- | --- | --- |
| Logic | 21 | 1 |
| Tests | 260 | 0 |

---
## Summary by cubic
Caps `retry_until` loops at 2048 iterations so a condition that never evaluates true fails with a clear, non-retryable error instead of looping until Temporal's history limits surface an opaque failure. Fixes #3411.

- Adds `MAX_RETRY_UNTIL_ITERATIONS = 2048` in `tracecat/dsl/constants.py`, reusing the do-while cap value under a distinct name.
- Bounds the loop with a deterministic local counter that is safe for replay.
- On cap exhaustion, raises a user-attributed `workflow.loop.limit_exceeded` error that names the task and suggests fixing the condition.
- Workflows whose conditions succeed within the cap are unaffected.
- Adds unit and Temporal tests covering cap exhaustion and in-cap success, with the Temporal test asserting exactly the capped number of executions.

<sup>Written for commit c5837c4c6d7d28be5c66f0bfd5310ddb8fa1c9da. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3412?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/temporal/test_workflow_timers.py
+++ b/tests/temporal/test_workflow_timers.py
@@ -21,6 +21,8 @@ pytestmark = [
 
 import temporalio.api.enums.v1
 from temporalio import activity
+from temporalio.api.operatorservice.v1 import AddSearchAttributesRequest
+from temporalio.client import WorkflowFailureError
 from temporalio.exceptions import ApplicationError
 from temporalio.testing import WorkflowEnvironment
 
@@ -34,7 +36,14 @@ from tracecat.dsl.worker import get_activities
 from tracecat.dsl.workflow import DSLWorkflow
 from tracecat.executor.activities import ExecutorActivities
 from tracecat.logger import logger
+from tracecat.runtime.errors import (
+    RetryDisposition,
+    RuntimeErrorKind,
+    RuntimeErrorOwner,
+)
 from tracecat.storage.object import InlineObject, StoredObject
+from tracecat.temporal.errors import extract_error_classification
+from tracecat.workflow.executions.enums import TemporalSearchAttr
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


topher-lo:
No. Temporal best practice IS infinite retries
.

PR #3414, opened 2026-09-07
fix(api): reject oversized webhook bodies with 413

### LINKED ISSUE #3413: Webhook request bodies are parsed without any size limit
# Webhook request bodies are parsed without any size limit

`parse_webhook_payload` in `tracecat/webhooks/dependencies.py` calls `await request.body()` and passes the result to `orjson.loads` without checking how large the body is. A client can POST an arbitrarily large payload to a webhook endpoint and the API will buffer all of it in memory before parsing.

## Actual behavior

- The full body is read into memory regardless of size.
- Only the MCP layer has an input size limit (`MCPInputSizeLimitMiddleware`, default 4 MiB). Webhook endpoints have none.

## Expected behavior

- Webhook payloads above a configured byte limit are rejected with `413` before being buffered or parsed.
- The limit should be configurable for operators and default to the same 4 MiB the MCP middleware uses.

## Impact

- Memory pressure on the API service from a single request (no authentication is required to reach a webhook URL).
- `orjson` then parses the whole buffer, adding CPU cost on top.

## Proposed fix

- Add `TRACECAT__WEBHOOK_MAX_BODY_BYTES` (bounded, default 4 MiB) to `tracecat/config.py`.
- In `parse_webhook_payload`, check the `content-length` header for a fast 413, then read the stream incrementally and abort with 413 as soon as the running total exceeds the cap.
- Cache the read body on the request the same way Starlette's `body()` does, so downstream raw-body envelopes and form parsing keep working unchanged.

## Environment

- Current `main` branch

warun7:
# Reject oversized webhook bodies with 413

Fixes #3413

## Summary

Webhook payloads are now size capped. `parse_webhook_payload` used to call `await request.body()` unbounded, so a single POST could buffer an arbitrarily large body in API memory before `orjson.loads` parsed it. The parser now checks the `content-length` header for a fast rejection, then reads the stream incrementally and aborts with `413 Content Too Large` the moment the running total exceeds the configured cap.

## Changes

- `tracecat/config.py`: add `TRACECAT__WEBHOOK_MAX_BODY_BYTES` via `bound_env`, defaulting to 4 MiB to match the existing MCP input size middleware. Minimum 1 byte.
- `tracecat/webhooks/dependencies.py`: add `_read_capped_body` and use it in `parse_webhook_payload`. The read body is cached on the request the same way Starlette's `body()` caches, so downstream consumers keep working unchanged: raw-body webhook envelopes (`router.py` reads `request.body()` after the payload dependency resolves) and form parsing (`request.form()` consumes the same cache). NDJSON, form, and JSON branches are untouched.
- `tests/unit/test_webhook.py`: migrate the payload parser fixtures from `MagicMock(spec=Request)` to real Starlette requests backed by ASGI receive channels, since the parser now consumes the stream. Add `TestWebhookBodySizeCap` covering: under-cap JSON, NDJSON, form, and Slack interaction payloads parse as before; over-cap bodies reject with 413 via the content-length fast path (asserting no stream bytes were pulled) and via a chunked stream that grows past the cap mid-read.

## Notes

- Out of scope, noted for reviewers: `dict(form_data)` in the form branch collapses repeated keys (last value wins). Pre-existing behavior, unchanged here.
- No `.env.example` change: the default works for ordinary deployments and the knob is operator-level tuning, per the repo rule on keeping `.env.example` focused.

## Test plan

- `uv run pytest tests/unit/test_webhook.py`: 33 passed (7 new cap tests plus migrated parser fixtures).
- `uv run ruff check`, `uv run ruff format --check`, `uv run basedpyright --warnings` on the three changed files: clean.

## LOC breakdown

| Category | Added | Removed |
| --- | --- | --- |
| Logic | +64 | -1 |
| Tests | +183 | -36 |

---
## Summary by cubic
Rejects oversized webhook bodies with HTTP 413 so a single POST can't buffer unbounded data in API memory. Previously `parse_webhook_payload` read the whole body without a cap; now it checks `content-length` for fast rejection and stream-reads, aborting as soon as the running total exceeds the limit.

- Adds `TRACECAT__WEBHOOK_MAX_BODY_BYTES` (default 4 MiB) to set the cap.
- Caches the read body on the request, so downstream `request.body()` and `request.form()` calls keep working unchanged.

<sup>Written for commit 105267143cfda2d3c09614d21ac7f4d7cf1de392. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/TracecatHQ/tracecat/pull/3414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/unit/test_webhook.py
+++ b/tests/unit/test_webhook.py
@@ -1,14 +1,14 @@
 import hashlib
+import urllib.parse
 import uuid
 from datetime import UTC, datetime
 from types import SimpleNamespace
-from typing import cast
+from typing import Any, cast
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import orjson
 import pytest
 from fastapi import HTTPException, Request
-from fastapi.datastructures import FormData
 from sqlalchemy.exc import NoResultFound
 
 from tracecat import config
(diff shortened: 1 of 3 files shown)


topher-lo:
This should be a WAF / ALB layer thing.

PR #3435, opened 2026-09-09
fix(skills): lock MCP references before publication

daryllimyt:
An MCP integration deleted between validation and publication left the published skill version pointing at a missing row (the FK is `ON DELETE SET NULL`, so the grant silently vanished). Take key-share locks on the projected MCP integrations inside the publish transaction and fail validation when any of them are gone.

Layer 3 of 5 in the split of #3325.

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 39 | 0 |
| Tests | 122 | 3 |

https://claude.ai/code/session_01LHvvWytf7Uz8nfcycVAbF8

---

### DIFF
--- a/tests/integration/test_skill_mcp_deletion.py
+++ b/tests/integration/test_skill_mcp_deletion.py
@@ -4,11 +4,11 @@ import asyncio
 import uuid
 
 import pytest
-from sqlalchemy import select, text
+from sqlalchemy import delete, select, text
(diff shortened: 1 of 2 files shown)


daryllimyt:
Closing: the original #3325 will merge instead. This stack was a byte-identical split of it.

PR #3437, opened 2026-09-09
feat(skills): add a tools picker to the skill frontmatter editor

daryllimyt:
Skill frontmatter tools had to be typed by hand. Add a searchable picker that lists registry actions and MCP tools, wire it into the SKILL.md editor panel, and keep malformed declarations out of the draft tool error codes so the raw YAML editor can repair them.

- `frontend/src/lib/skill-tools.ts` reads and rewrites `metadata.tools` while preserving the rest of the frontmatter source, and builds the option list from the registry and MCP catalogs.
- `skill-tools-dropdown.tsx` renders the picker on top of `MultiTagCommandInput`, which gains an accessible input id, labelled remove buttons, and closes when suggestions are disabled.
- Backend: `SKILL_DRAFT_TOOL_ERROR_CODES` no longer includes `invalid_skill_tool_declaration`, so drafts with malformed IDs save but cannot publish.

Layer 5 of 5 in the split of #3325. This is the feature; layers 1 to 4 are the backend fixes that were previously bundled with it. Supersedes #3325.

## LOC breakdown

| Category | + | - |
| --- | ---: | ---: |
| Logic | 497 | 4 |
| Tests | 715 | 7 |

https://claude.ai/code/session_01LHvvWytf7Uz8nfcycVAbF8

---

### DIFF
--- a/frontend/src/components/skills/editor-panel.tsx
+++ b/frontend/src/components/skills/editor-panel.tsx
@@ -18,6 +18,7 @@ import { CenteredSpinner } from "@/components/loading/spinner"
 import { SkillFileTree } from "@/components/skills/file-tree"
 import { RenameDialog } from "@/components/skills/rename-dialog"
 import { SkillFileIcon } from "@/components/skills/skill-file-icon"
+import { SkillToolsDropdown } from "@/components/skills/skill-tools-dropdown"
 
 const CodeEditor = dynamic(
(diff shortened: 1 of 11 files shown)


daryllimyt:
Closing: the original #3325 will merge instead. This stack was a byte-identical split of it.

PR #3443, opened 2026-09-09
feat(cases): aggregate cases by dropdowns and tags

aaron-tracecat:
## Description

Case aggregation can now filter and group by `dropdowns.<ref>` and `tags`. Dropdowns return option references and preserve unset or deleted selections in the null group. Tag grouping counts each case once per tag group; sums, means, and medians are rejected when tag grouping would repeat cases.

The query explicitly scopes dropdowns and tags to the workspace. Dropdown access and definition lookup run only when the request references a dropdown, including inside nested filters, so ordinary case and tag reports remain independent of Case Addons configuration. The request documentation explains the tag counting behavior for reuse by the registry action.

## Related issues

Closes ENG-1751
https://linear.app/tracecat/issue/ENG-1751

Builds on ENG-1750 / #3419. The registry action is tracked separately in ENG-1753.

## Validation

- 248 focused tests passed, including real PostgreSQL requests through the internal endpoint: `tests/unit/test_cases_query.py`, `tests/unit/test_case_aggregation.py`, and `tests/unit/query/test_aggregations.py`.
- Covers dropdown deletion and missing values, multiple dropdowns, tag counts and minimum group sizes, filter/group interactions, workspace isolation, entitlement rejection, mixed grouping, and the ungated-report regression found during adversarial review.
- Repository Ruff lint and formatting checks passed.
- Full Python type check passed with zero errors or warnings.

## LOC breakdown

| Kind | + | - |
| --- | ---: | ---: |
| Logic | 165 | 20 |
| Tests | 436 | 2 |

### DIFF
--- a/tests/unit/test_case_aggregation.py
+++ b/tests/unit/test_case_aggregation.py
@@ -19,7 +19,20 @@ from tracecat.cases.enums import CasePriority, CaseSeverity, CaseStatus
 from tracecat.cases.service import CasesService
 from tracecat.contexts import ctx_role
 from tracecat.db.engine import get_async_session
-from tracecat.db.models import Case, CaseFields, Organization, User, Workspace
+from tracecat.db.models import (
+    Case,
+    CaseDropdownDefinition,
+    CaseDropdownOption,
+    CaseDropdownValue,
+    CaseFields,
+    CaseTag,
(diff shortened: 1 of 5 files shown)


jordan-umusu (approved):
lgtm

PR #3528, opened 2026-09-16
feat(rbac): add SCIM connection credentials

jordan-umusu:
## What changes

Adds the credential an identity provider uses to call Tracecat's SCIM endpoints, and the role those calls run as. The endpoints themselves land in the next PR.

- **One connection per organization.** `scim_connection` stores a hashed token (`tc_scim_sk_` prefix, same scheme as managed API keys), a preview, `last_used_at`, and `revoked_at`. A unique index on `organization_id` enforces one row; issuing again rotates it.
- **Management API** under `/scim/connection`, gated by the RBAC add-on entitlement: `POST` issues or rotates and returns the raw token once, `GET` reads status without the token, `DELETE` revokes. Scopes: `org:rbac:create`, `org:rbac:read`, `org:rbac:delete`.
- **`scim` role type.** A bearer token resolves to `Role(type="scim")` carrying `organization_id`, `scim_connection_id`, and a fixed scope set of `org:member:remove`, the only scope the SCIM paths need. The role's authority is fixed in code, not stored per connection. Revoked tokens, unknown key ids, and unentitled organizations all get 401.
- **Audit attribution.** New `AuditEventActor.SCIM`; the connection id is the actor label. Existing service-account attribution is unchanged.

## Migration

`63bb4d4466aa` creates `scim_connection` with RLS. Revises `a667d946cca1` from the previous PR.

## Not in this PR

- SCIM 2.0 `/Users` and `/Groups` endpoints.
- Group mapping API and organization settings UI.

## Validation

- Ruff, `basedpyright --warnings`, Biome, and frontend typecheck clean.
- New tests: token issue/rotate/revoke, bearer authentication paths, audit actor chain, migration round-trip against a live database.

## LOC

| Category | + | - |
| --- | ---: | ---: |
| Logic | 397 | 8 |
| Tests | 1,015 | 0 |
| Generated | 279 | 3 |


jordan-umusu:
Superseded: stack resliced into data model (#3527), provisioning API, and UI PRs.

PR #3530, opened 2026-09-16
feat(rbac): add SCIM provisioning API

jordan-umusu:
Tracecat becomes a native SCIM 2.0 server (EE). The IdP pushes users and groups; admins map provider groups onto Tracecat groups; existing group role assignments stay the authority on what access a group grants.

Design, invariants, and the decision log live in [ENG-1728](https://linear.app/tracecat/issue/ENG-1728/scim).

## What changes

- **Connection tokens** (`/scim/connection`): issue, rotate, read, revoke a `tc_scim_sk_` token. Gated by the RBAC add-on and `org:rbac:*`.
- **`scim` role type.** A bearer token resolves to `Role(type="scim")` with a fixed `org:member:remove` scope. Audit events record a `SCIM` actor.
- **SCIM 2.0 protocol** (`/scim/v2`): Okta's three discovery endpoints, plus `Users` and `Groups` CRUD with `PATCH` and filter-by-`userName`.
- **Provisioning is create-or-link.** An existing email links through an `external_user` row rather than failing. No role is granted: admission is the `organization_membership` row, and any pending invitation is revoked at admission.
- **Pending, review, activate.** A new connection admits nobody until an admin reviews what arrived and applies mappings in one transaction. After that every push applies on arrival.
- **Deprovisioning removes, it does not suspend.** `active=false` and `DELETE` run the org-scoped `delete_member` path. The global `is_active` flag is never written; the `external_user` row survives so reactivation relinks the same identity.
- **A mapped group is IdP-owned.** First mapping purges manual members (audited), manual edits 409, last unmap freezes IdP members as manual rows.
- **Login policy.** A provisioned user's generated password cannot be used; discovery still routes invited users to basic auth.

## Reviewing this PR

Logic is ~28% of the diff; the rest is tests and the generated client, which GitHub collapses.

Read `packages/tracecat-ee/tracecat_ee/scim/` in this order: `protocol.py` (parsing only, no database) → `provisioning.py` → `service.py`.

`tracecat/db/models.py` is unchanged here — the role-path union and `effective_group_members` belong to #3527.

## Not in this PR

Organization settings UI (#3531). The `trace_member_access` IdP arm does not join organization admission, so it can over-report access for a deprovisioned-but-relinkable user; tracked separately.

## Validation

Targeted unit suites on the rebased branch (SCIM protocol, mappings, grant resolver, organization service, workspace membership): 119 passed, 0 failed. Okta SPEC suite against a live organization not yet run.

## LOC

| Category | + | - |
| --- | ---: | ---: |
| Logic | 3,416 | 58 |
| Tests | 6,309 | 12 |
| Infra/config | 0 | 0 |
| Docs | 0 | 0 |
| Generated | 2,436 | 54 |


tracecat/auth/discovery.py
```diff
@@ -126,16 +137,79 @@ async def _resolve_organization_by_slug(
...
+        self, org_id: OrganizationID, email: str | None = None
     ) -> AuthDiscoveryMethod:
         if await self._org_saml_enabled(org_id):
+            # A live invitation is an admin-created non-IdP path: the IdP does
+            # not know this invitee, so offer basic auth when it is enabled.
+            if (
+                email is not None
+                and await self._org_basic_enabled(org_id)
+                and (
+                    await self._has_live_invitation(org_id, email)
```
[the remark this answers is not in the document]
jordan-umusu: Confirmed and tracked separately, not fixed in this PR.

The asymmetry is real: `_has_live_invitation` checks only status and expiry, while `_is_manual_invitee` ends in `return not saml_enforced`. Since the two are `or`-ed, a pending invitation routes to BASIC even under enforcement.

Worth recording for whoever picks this up: discovery is only the UI hint, so fixing it here would not close the hole. The account-level gate is `_is_local_password_login_allowed`, which derives `org_ids` from `organization_membership`; a pending invitee has no membership row yet, so `org_ids` is empty and local login is permitted regardless of what discovery returned. The fix therefore belongs in `users.py`, and it has to consider invitations rather than membership alone.

Leaving this open deliberately pending that change.

tracecat/api/app.py
```diff
@@ -571,6 +647,21 @@ def create_app(**kwargs) -> FastAPI:
...
     app.include_router(rbac_assignments_router)
+
+    # EE-only SCIM connection and mapping administration - gated by RBAC entitlement
+    from tracecat_ee.scim.protocol import router as scim_protocol_router
+    from tracecat_ee.scim.router import (
+        connections_router as scim_connections_router,
+    )
+    from tracecat_ee.scim.router import (
+        mappings_router as scim_mappings_router,
+    )
```
daryllimyt, line 658: nit: just import at the top since we already import tracecat_ee

packages/tracecat-ee/tracecat_ee/scim/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+            session, connection.organization_id, Entitlement.RBAC_ADDONS
+        ):
+            raise UNAUTHORIZED_EXCEPTION
+
+        connection.last_used_at = datetime.now(UTC)
+        session.add(connection)
+        await session.commit()
+
+        role = Role(
+            type="scim",
```
daryllimyt, line 78: Flagging a deliberate gap so it's recorded: scim roles bypass `enforce_org_ip_allowlist` (`tracecat/auth/credentials.py:862` only gates `user` and `service_account`, and this dependency never reaches `_role_dependency` anyway). That's the right call since the IdP pushes from its own egress IPs, but an org with an allowlist will expect it to apply. Can we note it in the SCIM docs and in this module's docstring?

tracecat/auth/credentials.py
```diff
@@ -442,7 +442,7 @@ async def _authenticate_api_key(
 
 @contextmanager
 def TemporaryRole(
-    type: Literal["user", "service", "service_account"] = "service",
+    type: Literal["user", "service", "service_account", "scim"] = "service",
```
daryllimyt, line 445: Nit: `"scim"` in this Literal is dead. The body builds a Role with no `organization_id` or `scim_connection_id`, so the validator in `Role` rejects it. Either drop `"scim"` from the Literal or thread the two ids through.

daryllimyt, line 445: Re the `x-tracecat-role-type` parser at `tracecat/auth/credentials.py:332` (outside the diff, so anchoring here):

`Role.to_headers()` now emits `x-tracecat-role-type: scim` and `x-tracecat-role-scim-connection-id`, but this parser only recognises `service_account` and defaults everything else to `service`. A scim role forwarded over an internal hop would arrive as a `service` principal with header-supplied scopes and no connection id. Nothing forwards one today, so this is latent, but it's ~10 lines to add a `scim` branch mirroring the `service_account` one and it closes the gap before something does. Happy for this to be a follow-up if you'd rather keep this PR tight.

packages/tracecat-ee/tracecat_ee/scim/schemas.py
```diff
@@ -0,0 +1,309 @@
...
+
+USER_SCHEMA = "urn:ietf:params:scim:schemas:core:2.0:User"
+GROUP_SCHEMA = "urn:ietf:params:scim:schemas:core:2.0:Group"
+LIST_RESPONSE_SCHEMA = "urn:ietf:params:scim:api:messages:2.0:ListResponse"
+PATCH_OP_SCHEMA = "urn:ietf:params:scim:api:messages:2.0:PatchOp"
+ERROR_SCHEMA = "urn:ietf:params:scim:api:messages:2.0:Error"
+SERVICE_PROVIDER_CONFIG_SCHEMA = (
+    "urn:ietf:params:scim:schemas:core:2.0:ServiceProviderConfig"
+)
+RESOURCE_TYPE_SCHEMA = "urn:ietf:params:scim:schemas:core:2.0:ResourceType"
```
daryllimyt, line 118: this could be a StrEnum

packages/tracecat-ee/tracecat_ee/scim/service.py
```diff
@@ -0,0 +1,916 @@
...
+        """The users the provider has pushed into this organization."""
+        rows = (
+            await self.session.execute(
+                select(  # pyright: ignore[reportCallIssue]
+                    ExternalUser.id,
+                    User.email,  # pyright: ignore[reportArgumentType]
+                    ExternalUser.external_id,
+                    ExternalUser.active,
+                )
+                .join(User, User.id == ExternalUser.user_id)  # pyright: ignore[reportArgumentType]
```
daryllimyt, line 588: nit: cast instead of typeignore

daryllimyt (approved):
lgtm, flagging some items

PR #3572, opened 2026-09-18
feat(tables): track semantic search indexing lifecycle

aaron-tracecat:
Persist semantic-search column selections and invalidate stale row embeddings in the same transaction as source edits. Changing selected text invalidates the whole published row revision immediately; changing unselected columns leaves it intact. Selections can be saved before an embedding provider is configured.

Covers single/batch writes, upserts, CSV imports, managed-table editing, schema sync, and deletion while search is disabled. Shared storage owns invalidation for normal writes and backfill. Adds bounded source reads and the worker handoff contract. The migration preserves existing bindings and allows an absent provider reference; its guarded downgrade preserves unconfigured selections and gives an explicit recovery error directing operators to restore a backup/snapshot or remove incompatible selections before retrying.

Status requires both `table:read` and `workspace:read`. Selection and retry validate all requested changes before committing, and unexpected internal failures remain server errors. The generated client documents both domain and request-validation 422 responses. Progress uses the shared `PageParams`/`paginate` implementation with a 20-document default and 100-document maximum. Opaque next/previous cursors are bound to the collection and generation; chunk sampling stays bounded to 1,001 states per document. Invalid cursors return a safe 400 response.

Closes [ENG-1819](https://linear.app/tracecat/issue/ENG-1819). Stacked on #3518 (`alee/eng-1817`); background execution and ranking are implemented by later PRs. Enable search only after compatible writers are deployed.

Validation: 25 lifecycle integration tests and 17 route tests passed in the latest focused runs, including transaction rollback, scope checks, reverse pagination, invalid/cross-table/version cursors, page bounds, and the generated HTTP/OpenAPI contract. The live migration regression passed again with assertions for the recovery message, preserved data/schema after rejection, and successful downgrade after incompatible selections are removed. Broader table/storage regressions passed earlier. API client regeneration and Python/frontend checks passed. The test expression that exposed a Ruff-version difference now passes both the commit hook and CI's pinned formatter.

| Category | + | - |
| --- | ---: | ---: |
| Logic | 1004 | 48 |
| Tests | 1353 | 0 |
| Generated | 801 | 0 |


tracecat/tables/search_router.py
```diff
@@ -0,0 +1,128 @@
+"""Public table selection, status and retry; query routes belong to search."""
```
jordan-umusu: nit: move search_* into tables/search/*

tracecat/tables/search_router.py
```diff
@@ -0,0 +1,128 @@
...
+        yield
+        return
+    except SearchError as exc:
+        status = 404 if exc.code == SearchErrorCode.NOT_FOUND else 409
+        error = HTTPException(status, detail={"code": exc.code.value})
+    except TracecatNotFoundError:
+        error = HTTPException(404, detail={"code": "NOT_FOUND"})
+    except ValueError:
+        error = HTTPException(422, detail={"code": "INVALID_SELECTION"})
+    raise error
```
jordan-umusu: Wrapping the whole body hides which call yields which status — in `get_table_search` the 404 could come from `service.table()`, `service.configuration()`, or the availability lookup. Sibling routes in `tables/router.py` wrap just the failing call, and doing the same here drops the four `raise AssertionError("unreachable")` lines:

```python
try:
    table = await service.get_table(table_id)
except TracecatNotFoundError as e:
    raise HTTPException(status_code=404, detail=str(e)) from e
```

tracecat/tables/search_router.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def table_search(
+    session: AsyncDBSession, role: WorkspaceActorRouteRole
+) -> TableSearchService:
+    assert role.organization_id is not None and role.workspace_id is not None
+    return TableSearchService(
+        session, SearchScope(role.organization_id, role.workspace_id)
+    )
```
jordan-umusu: All 18 routes in `tables/router.py` construct the service inline, and `Depends` is used once in that module — for a query param, never a service. Since this takes exactly the two args every route already declares, inlining it at the four call sites removes a hop:

```python
service = TableSearchService(session, SearchScope(role.organization_id, role.workspace_id))
```

tracecat/tables/search_router.py
```diff
@@ -0,0 +1,128 @@
...
+    params: TableSearchSelection,
+    role: WorkspaceActorRouteRole,
+    session: AsyncDBSession,
+) -> TableSearchConfiguration:
+    """Persist selection and backfill marker together, even without a provider."""
+    async with search_errors():
+        service = table_search(session, role)
+        await service.select_column(table_id, params)
+        result = await service.configuration(table_id)
+        await session.commit()
```
jordan-umusu: `session.commit()` here means the caller has to know `select_column` leaves an open transaction and that `configuration()` must be read before the commit (same on line 107). Worth checking whether `TableSearchService` should commit internally like the other services, unless the single transaction is deliberate.

jordan-umusu (changes requested):

tracecat/tables/search_router.py
aaron-tracecat: Moved the related files into tables/search/ and updated the imports in abe269de2. The folder now contains router.py, schemas.py, service.py, and source.py. The API URLs and request/response formats are unchanged.

tracecat/tables/search_router.py
aaron-tracecat: Agreed. In abe269de2, each call now handles its expected errors directly, and the broad wrapper and unreachable assertions are gone. This also fixes a misleading case: a failure while building the response or saving changes no longer gets reported as an invalid column selection. Added tests for those cases while keeping the existing public error codes.

tracecat/tables/search_router.py
aaron-tracecat: Done in abe269de2. Each of the four routes now creates TableSearchService directly, matching the surrounding table routes. Removed the helper so the setup is visible where it is used.

tracecat/tables/search_router.py
aaron-tracecat: Yes, the single transaction is deliberate. These shared methods leave the final save to the caller so a larger operation can undo all its changes if something fails. For selection, we change the selected columns, record the indexing work, and read the resulting status while the same lock is held, then save everything together. Retry also waits until the whole requested batch has been checked before saving. I kept the commits here and added comments explaining the order, plus a test proving that a caller can roll back a retry (abe269de2). Leaving this thread open for your take on that choice.

tracecat/tables/search/router.py
```diff
@@ -0,0 +1,151 @@
...
+    try:
+        await service.table(table_id)
+    except TracecatNotFoundError as exc:
+        raise HTTPException(404, detail={"code": "NOT_FOUND"}) from exc
+    # Provider resolution is a separate read-only transaction, never a source
+    # mutation. Authorize the table before inspecting workspace availability.
+    availability = None
+    provider_error = False
+    try:
+        availability = await WorkspaceEmbeddingService(role).get()
```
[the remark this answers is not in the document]
aaron-tracecat: Fixed in a878759d5. This endpoint includes provider availability, which already requires workspace-read permission. The route now explicitly requires both table:read and workspace:read, and its API description says so. It checks both permissions before reading table data. Added tests for roles with either permission alone, neither permission, and both; document progress keeps its existing table-read requirement.

tracecat/tables/search/router.py
```diff
@@ -0,0 +1,151 @@
...
+    TableSearchProgressPage,
+    TableSearchRetry,
+    TableSearchSelection,
+)
+from tracecat.tables.search.service import TableSearchService
+
+router = APIRouter(
+    prefix="/{table_id}/search",
+    tags=["tables"],
+    responses={code: {"model": TableSearchErrorResponse} for code in (404, 409)},
```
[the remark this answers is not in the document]
aaron-tracecat: Fixed in a878759d5. A 422 response can contain either an invalid-selection code or the normal list of request validation errors. The selection endpoint now documents both formats, and I regenerated the frontend client so its types match. Tests check both real HTTP responses and the published schema; other endpoints keep their existing error contracts.

tracecat/tables/search/router.py
```diff
@@ -0,0 +1,154 @@
...
+
+@router.get("/documents")
+@require_scope("table:read")
+async def get_table_search_progress(
+    table_id: TableID,
+    role: WorkspaceActorRouteRole,
+    session: AsyncDBSession,
+    generation: int = Query(ge=1),
+    cursor: UUID | None = None,
+    limit: int = Query(default=20, ge=1, le=100),
```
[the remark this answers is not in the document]
aaron-tracecat: Fixed in 31eb65f1d. Progress now uses the repository’s current PageParams/paginate implementation (the older CursorPaginationParams is marked for existing callers). It returns opaque next and previous cursors, bound to this table and index generation, while preserving the 20-row default, 100-row maximum, and bounded chunk sampling. Added forward/backward, invalid/cross-table/version cursor, bounds, and HTTP/OpenAPI tests; regenerated the client.

alembic/versions/8c0e18190001_allow_unconfigured_search_selection.py
```diff
@@ -0,0 +1,37 @@
...
+        existing_type=sa.BigInteger(),
+        nullable=True,
+    )
+
+
+def downgrade() -> None:
+    # Fail rather than discard saved selections that have no provider binding.
+    op.alter_column(
+        "search_collection",
+        "config_version",
```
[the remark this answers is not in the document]
aaron-tracecat: Fixed in 260348132. The downgrade now checks for saved selections without a provider before changing the column. If it finds any, it stops with an explicit explanation: restore a database backup/snapshot before rolling the app back, or remove the incompatible selections before retrying. It does not delete selections itself. The live migration test verifies the message, confirms that the data and nullable column are preserved after rejection, and confirms that downgrade succeeds once the incompatible selections are removed.

jordan-umusu (approved):

PR #3575, opened 2026-09-18
feat(engine): index table text in durable background batches

aaron-tracecat:
Index selected table text in bounded background batches. Long rows resume from durable chunk checkpoints, and a row becomes searchable only after every chunk of its current revision has been embedded.

Adds a recurring Temporal dispatcher, fenced writes that reject superseded workers, retry/backoff, stale-data cleanup, and shared Redis capacity limits that reserve room for foreground queries. Provider calls run outside database transactions. The worker consumes the lifecycle and provider contracts from the preceding PRs. See `tracecat/search/INDEXING.md` for operational details.

Implements [ENG-1820](https://linear.app/tracecat/issue/ENG-1820). Stacked on #3572 (`alee/eng-1819`).

Validation: 23 focused tests passed using live PostgreSQL/Redis, a controlled HTTP provider, and a real local Temporal server. Coverage includes crash/resume, stale edits, retries, fairness, capacity, schedule reconciliation, and dispatch. Storage/lifecycle/chunker regressions and signed-commit hooks passed.

| Category | + | - |
| --- | ---: | ---: |
| Logic | 773 | 9 |
| Tests | 383 | 0 |
| Docs | 67 | 0 |


tracecat/search/indexing.py
```diff
@@ -0,0 +1,341 @@
...
+                    inputs,
+                )
+            )
+            if inputs
+            else None
+        )
+        await _finish_batch(work, claim, result, progress)
+    except (EmbeddingError, ChunkingError, SearchError) as exc:
+        if claim is not None:
+            await record_failure(work, claim, exc)
```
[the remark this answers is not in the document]
aaron-tracecat: Fixed in cf39280d4. A damaged checkpoint now marks the row as failed with MANIFEST_CONFLICT, clears its lease, and stops automatic retries. The decoder discards the original validation error so its contents cannot leak. Rebuilding the collection replaces the invalid checkpoint. Tests cover both malformed checkpoints and conflicting old-format positions, and confirm that no provider call or repeated claim occurs.

tracecat/search/indexing_types.py
```diff
@@ -0,0 +1,35 @@
...
+        return SearchScope(self.organization_id, self.workspace_id)
+
+
+class DispatchPage(BaseModel):
+    collections: list[CollectionWork]
+    next_cursor: UUID | None = None
+
+
+class IndexingProgress(BaseModel):
+    outcome: str
```
daryllimyt, line 26: could be enum

tracecat/search/indexing_schedule.py
```diff
@@ -0,0 +1,45 @@
...
+    ScheduleUpdate,
+    ScheduleUpdateInput,
+)
+
+from tracecat.search.indexing_workflow import SearchIndexDispatcher
+
+SCHEDULE_ID = "semantic-search-dispatch-v1"
+
+
+async def ensure_search_schedule(client: Client, task_queue: str) -> None:
```
daryllimyt, line 23: i think this should fail open because it runs blocking in the Worker's startup path, otherwise run it as a background task outside of the startup path

daryllimyt:
**From thermos — [P2, non-blocking] Reuse freshly prepared chunk inputs**

Location: [tracecat/search/indexing.py:260–266](https://github.com/TracecatHQ/tracecat/blob/cf39280d4c356710551908ecb592314cf7d5d217/tracecat/search/indexing.py#L260-L266).

`prepare_batch()` already returns each chunk's text and token count. After persisting the manifests, the worker reselects them and immediately reconstructs the same inputs in the same transaction. Each reconstruction calls `read_slice()`, repeating collection, lock, workspace-state, table metadata, and source queries—roughly seven SQL statements per chunk, or up to 224 unnecessary statements for a 32-input batch while the workspace write lock remains held.

Reuse `PreparedChunk.text` and `metadata.token_count` for newly prepared chunks, and reserve reconstruction for persisted manifests resumed from previous turns. Both paths can feed the same provider-budget selection.

The thermos review found no correctness/security defects or structural blockers. Validation: 54 focused tests passed, Ruff and targeted basedpyright clean.

daryllimyt (changes requested):
lgtm, with some concerns:
1. possible ensure-schedule failure in worker startup path. 
2. possible unexpected cpu bound work running on temporal main event loop

one idea is that we could introduce a new worker class that just handles background/adhoc work so as to avoid interfering with DSL and Agent workloads

daryllimyt (approved):

tracecat/search/indexing_types.py
aaron-tracecat: Fixed in [46f43b31](https://github.com/TracecatHQ/tracecat/commit/46f43b31946a3c358c528c19185196711128799e). Normal outcomes now use an enum, and failure outcomes use the existing error enums. This prevents arbitrary strings or typos while keeping the values sent between workers unchanged. Tests check that every supported value survives serialization and that unknown values are rejected.

tracecat/search/indexing_schedule.py
aaron-tracecat: Fixed in [46f43b31](https://github.com/TracecatHQ/tracecat/commit/46f43b31946a3c358c528c19185196711128799e). Schedule setup now runs in a background task, so ordinary workflow processing can start even if registration fails. Each attempt times out after 30 seconds and retries after 30 seconds. Worker shutdown cancels the task, and reconciliation still preserves an operator-paused schedule. Tests cover create/update failures, timeouts, recovery, and cancellation.

aaron-tracecat:
@daryllimyt Addressed the remaining concerns in [46f43b31](https://github.com/TracecatHQ/tracecat/commit/46f43b31946a3c358c528c19185196711128799e):

- **CPU work on the event loop:** tokenizer initialization, token counting, and chunk computation now run in threads. Database reads stay on their original event loop. Tests verify that the loop can keep doing other work while tokenization is waiting. A separate worker could still provide stronger CPU and memory isolation; this change avoids doing the expensive synchronous computation directly on the shared loop.
- **Rebuilding freshly prepared inputs:** fresh chunks now reuse the text and token counts already produced during preparation. Only saved chunks resumed from an earlier turn are reconstructed and checked against their saved fingerprints. Both paths share the same provider limits, and the full checkpoint is saved before the provider request, so an interruption does not lose the remaining work. A database test checks that a fresh, single-window chunk reads its source only once; resume tests check that the same input and token limits apply after interruption.

The enum and schedule-startup fixes are also pushed, with details in their inline threads. Validation: 142 focused tests passed, followed by five affected cases after final test typing fixes. Ruff, full Python type checking, frontend client generation/checks, and the normal signed-commit hooks passed.

